PR #29, opened 2017-05-22
checkcommits: Fix handling of ignoring branches.

jodh-intel:
Previously, the destination branch would be ignored but that was not the
intent (destination is effectively always "master"). Now, the tool
allows the specification of one or more regular expressions to filter
out source (PR) branches which will be ignored (always considered
successful).


devimc:
lgtm

Issue #24, opened 2017-05-17
Fix versions for clear container kernel and image for CI


dlespiau:
We can't always use the latest versions of kernel/clear image in CI: if we do that, tests runs have too many moving pieces and we can't draw good conclusions.

dlespiau:
This means using something like versions.txt in 2.x. I guess this file could be part of the tests repo.

jodh-intel:
We should add comments to that file too when it's copied across to explain what every entry is for.

chavafg:
Now that we are moving away from semaphore and that we do not have problems with the kernel versions, I think we can close this issue as the kernel and clear-containers-image version will be taking from the OBS repositories. 

When there is a new fix in the image or the kernel, QA usually test the change before enabling the package in the stable obs repository, so we do not end with a version that doesn't work.

What do you think @jodh-intel ?

chavafg:
Now that OBS is down, I agree that we should still maintain the versions that we use for testing. I opened #514 to not rely on OBS and updated test-versions.txt

PR #34, opened 2017-05-23
Standalone clear image & kernel install scripts

dlespiau:
Note: I couldn't test this in CI, and these are fairly intrusive changes. At least the 2 individual scripts do work locally.

I decided I wouldn't install clear kernel and images by hand any longer so I split two scripts out of the ci setup script.

```
# ./.ci/install_clear_kernel.sh 14700 4.9.4-57 /usr/share/clear-containers
# ./.ci/install_clear_image.sh 14700 /usr/share/clear-containers
```

### DIFF
--- a/.ci/install_cc_env_ubuntu.sh
+++ b/.ci/install_cc_env_ubuntu.sh
@@ -29,9 +29,6 @@ chronic sudo apt-get install -y rpm2cpio
 echo "Update apt repositories"
 chronic sudo apt-get update
 
-echo "Install linux-container kernel"
-chronic sudo apt-get install -y --force-yes linux-container
-
 echo "Install qemu-lite binary"
 chronic sudo apt-get install -y --force-yes qemu-lite
 
--- a/.ci/install_cc_env_ubuntu.sh
+++ b/.ci/install_cc_env_ubuntu.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
-
+#
 # Copyright (c) 2017 Intel Corporation
-
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -13,7 +13,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
 
 set -e
 
--- a/.ci/install_cc_env_ubuntu.sh
+++ b/.ci/install_cc_env_ubuntu.sh
@@ -16,6 +16,8 @@
 
 set -e
 
+cidir=$(dirname "$0")
+
 echo "Add clear containers sources to apt list"
 sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/clearlinux:/preview:/clear-containers-2.1/xUbuntu_16.10/ /' >> /etc/apt/sources.list.d/cc-oci-runtime.list"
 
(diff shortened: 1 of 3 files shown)


.ci/install_clear_image.sh
```diff
@@ -0,0 +1,50 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+if [ "$#" -ne 2 ]; then
+    echo "Usage: $0 CLEAR_RELEASE PATH"
```
jodh-intel: How will the user know what value to provide for `CLEAR_RELEASE`?

It would be helpful if the script supported `latest` (either to simply show the newest-available release version and exit, or actually use that version). Same comment applies to `install_clear_kernel.sh`).

jodh-intel: Also, it would be better if these messages went to `stderr` (same comment for script below).

dlespiau: All of the above would be nice, but note I tried to stay very simple and leave the decisions like "latest" or a fixed version to an entity above that (so I don't have to repeat the same logic or source a dependency in those leaf scripts). I think even without your suggestions this more versatile than what we had before.

jodh-intel: Well, we now have 3 scripts that all hard-code the download URL. I'd still argue that `install_cc_env_ubuntu.sh` doesn't need to know that URL since it's querying for `latest` anyway.

Then, to avoid the last bit of duplication `install_clear_image.sh` and `install_clear_kernel.sh`  could both source a script/config file for that URL. I know the URL is unlikely to change, but that isn't the point here.

https://github.com/01org/cc-oci-runtime/tree/master/.ci/ gets all this right and has been in use for a long time. But we now have these scripts plus the bunch in https://github.com/clearcontainers/runtime/tree/master/.ci all doing it slightly differently. It's getting rather confusing. Is the plan to migrate all the scripts from the `runtime` to the `tests` repo?

.ci/install_cc_env_ubuntu.sh
```diff
@@ -29,42 +30,19 @@ chronic sudo apt-get install -y rpm2cpio
...
-sudo mkdir -p ${cc_img_path}
-echo "Install clear containers image"
-sudo install --owner root --group root --mode 0755 clear-${latest_version}-containers.img ${cc_img_path}
 
-echo -e "Create symbolic link ${cc_img_path}/${cc_img_link_name}"
-sudo ln -fs ${cc_img_path}/clear-${latest_version}-containers.img ${cc_img_path}/${cc_img_link_name}
+"./$cidir/install_clear_image.sh" ${clear_release} "${cc_img_path}"
 
 bug_url="https://github.com/clearcontainers/runtime/issues/91"
+kernel_clear_release=12760
```
devimc: should we add a ```versions.txt``` to pull the versions from it ?

dlespiau: This kernel is version is fixed because we newer ones oops on semaphore CI (issue https://github.com/clearcontainers/runtime/issues/91). We should at some point use a versions.txt (we could have different configurations, one versions.txt for semaphore CI, one for devs). That part is captured in https://github.com/clearcontainers/tests/issues/24.

.ci/install_clear_image.sh
chavafg: @jodh-intel , can we merge this one and open an issue to address your comment above?

jodh-intel: Sure. Let's not block on this. But we really must remember to clean up all these very similar scripts floating about.

jodh-intel:
let's get this landed and improve incrementally to keep the momentum for now.

lgtm

sboeuf:
LGTM

chavafg:
lgtm

PR #38, opened 2017-05-31
functional: Runs functional tests from any directory

devimc:
with this patch full paths are used to run functional tests
allowing the execution of them from any directory

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,15 @@ SHIM ?= "${GOPATH}/src/github.com/clearcontainers/shim/cc-shim"
 # The time limit in seconds for each test
 TIMEOUT ?= 5
 
-functional:
-	ginkgo ./functional/ -- -runtime "${RUNTIME}" -proxy "${PROXY}" -shim "${SHIM}" -timeout ${TIMEOUT}
+TESTS_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
+GINKGO_PATH = ${GOPATH}/bin/ginkgo
+
+ginkgo:
+	go get github.com/onsi/ginkgo/ginkgo
+
+functional: ginkgo
+	$(GINKGO_PATH) $(TESTS_DIR)/functional/ -- -runtime "${RUNTIME}" -proxy "${PROXY}" -shim "${SHIM}" -timeout ${TIMEOUT}
 
 check:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -10,8 +10,17 @@ SHIM ?= "${GOPATH}/src/github.com/clearcontainers/shim/cc-shim"
...
 TIMEOUT ?= 5
 
-functional:
-	ginkgo ./functional/ -- -runtime "${RUNTIME}" -proxy "${PROXY}" -shim "${SHIM}" -timeout ${TIMEOUT}
+TESTS_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
+GINKGO_PATH = ${GOPATH}/bin/ginkgo
+
+ginkgo:
+ifeq ($(wildcard $(GINKGO_PATH)),)
```
dlespiau: what is this ifeq supposed to do? it's not clear to me, GINKGO_PATH is never empty as defined just above.

dlespiau:
Actually, not a big fan of using makefiles here. Probably simpler in the long run to have one entry script for each kind of tests (functional, docker, CRI-O), etc...

Maybe we could have a makefile target (or a master script) to run all the different kind of testing in succession.

Makefile
devimc: on success return the path (non-empty string) , else nothing

devimc:
Actually ginkgo is used to run our tests
```
$ ginkgo pathToFunctionalTests
$ ginkgo pathToIntegrationTests
```
I think we do not need scripts

Makefile
dlespiau: I don't find this readable nor obvious, so can't give my lgtm, but I'm sure others can.

jodh-intel: I agree that some of the make functions can be pretty opaque, but I'm used to this idiom personally. However, how about changing it to use `$(realpath )` since that make is _slightly_ clearer what is going on.

Alternatively, you could do something like the following which makes the logic clearer still:

```
# Return non-empty string if file exists
define file_exists
$(realpath $1)
endef

exists = $(call file_exists,$(GINKGO_PATH))

ginkgo:
ifeq ($(exists),)
  go get github.com/onsi/ginkgo/ginkgo
endif

```

GabyCT:
lgtm

PR #44, opened 2017-06-02
Improve CI

### LINKED ISSUE #36: Improve CI scripts
There are now some issues in the CI scripts that we need to address: 

1. Some URLs are hard-coded and are used in more than 1 script. 
2. There are similar/duplicated scripts on other repos that are trying to do the same.

From @jodh-intel in #34: 

_Well, we now have 3 scripts that all hard-code the download URL. I'd still argue that install_cc_env_ubuntu.sh doesn't need to know that URL since it's querying for latest anyway.

Then, to avoid the last bit of duplication install_clear_image.sh and install_clear_kernel.sh could both source a script/config file for that URL. I know the URL is unlikely to change, but that isn't the point here.
https://github.com/01org/cc-oci-runtime/tree/master/.ci/ gets all this right and has been in use for a long time. But we now have these scripts plus the bunch in https://github.com/clearcontainers/runtime/tree/master/.ci all doing it slightly differently. It's getting rather confusing. Is the plan to migrate all the scripts from the runtime to the tests repo?_

devimc:
Improve CI

fixes #36


sboeuf:
@devimc You should not remove the scripts you are removing with this PR. We could need them for different purposes. For example, the script installing clear-containers.img should be called from setup_env_ubuntu.sh or setup_env_fedora.sh. The same for the the kernel script that @dlespiau created with the parameter for the version.
Try to keep small scripts so that we can reuse them across repos. Virtcontainers needs to install the proxy the same way the runtime has to do it, then it does worth to have a script for proxy intall. And that's the same for the shim I think.
Does that make sense ?

devimc:
@sboeuf I only agree in the first point in order to support multiple distros
in the second point, the idea is to share setup.sh, run.sh and teardown.sh between repos

PR #45, opened 2017-06-05
Improve CI scripts

### LINKED ISSUE #36: Improve CI scripts

devimc:
Improve CI scripts

fixes #36


.ci/install_proxy.sh
```diff
@@ -17,22 +17,12 @@
...
+source ${TESTS_CI_DIR}/lib.sh
 
-echo "Install cc-shim"
-sudo make install
+clone_build_and_install "github.com/clearcontainers/proxy"
 
-popd
-popd
+echo "Install proxy service (/etc/init/cc-proxy.conf)"
+sudo cp ${TESTS_CI_DIR}/upstart-services/cc-proxy.conf /etc/init/
```
chavafg: does this path exists: `upstart-services/cc-proxy.conf`?

.ci/install_virtcontainers.sh
```diff
@@ -0,0 +1,31 @@
...
+set -e
+
+source ${TESTS_CI_DIR}/lib.sh
+
+virtcontainers_project="github.com/containers/virtcontainers"
+virtcontainers_dir="${GOPATH}/src/${virtcontainers_project}"
+virtcontainers_pause_bin=pause
+virtcontainers_install_dir="/var/lib/clear-containers/runtime/bundles/pause_bundle/bin"
+
+clone_and_build ${virtcontainers_project} ${virtcontainers_pause}
```
chavafg: I do not see `$virtcontainers_pause` declared

devimc: fixed

devimc: thanks :)

chavafg:
Can you fix small typo in the commit message of `a2cd91d4b4a53ca42b5b4bed9acb13c7ce8a7f43`:
`s/papath/path/`

chavafg:
While reviewing this PR, found that `.ci/install_clear_image.sh`  line 1 is empty. Can you fix it here?

.ci/install_runtime.sh
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+#
+
+set -e
+
+source ${TESTS_CI_DIR}/lib.sh
+
+runtime_dir="${GOPATH}/src/github.com/clearcontainers/runtime"
+runtime_config_path="/etc/clear-containers"
+runtime_config_file="configuration.toml"
```
jodh-intel: Nit: this could be `config/configuration.toml` to avoid having to add the `config/` in the `cp` below (but I guess you did it this way to make the message clearer to users).

.ci/setup.sh
```diff
@@ -0,0 +1,38 @@
...
+# Setup workflow:
+# 1. setup environment (install qemu, kernel and image)
+# 2. Clone repos, fetch branches, build and install
+
+set -e
+
+export TESTS_CI_DIR=${GOPATH}/src/github.com/clearcontainers/tests/.ci
+
+echo "Set up environment"
+bash -f ${TESTS_CI_DIR}/setup_env_ubuntu.sh
```
jodh-intel: This is safe for now, but I wonder if we should add a check at some point to ensure the host system is actually running Ubuntu (`grep -q "^ID=ubuntu$" /etc/os-release` or similar).

.ci/lib.sh
```diff
@@ -0,0 +1,49 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -e
+
+function clone_and_build() {
+	github_project=$1
+	make_target=$2
```
jodh-intel: I'd always quote args "just in case" ;-)

jodh-intel:
lgtm

.ci/setup_env_ubuntu.sh
```diff
@@ -36,7 +34,7 @@ chronic sudo apt-get install -y --force-yes qemu-lite
 clear_release=$(curl -sL https://download.clearlinux.org/latest)
 cc_img_path="/usr/share/clear-containers"
 
-"./$cidir/install_clear_image.sh" ${clear_release} "${cc_img_path}"
+"${TESTS_CI_DIR}/install_clear_image.sh" ${clear_release} "${cc_img_path}"
```
dlespiau: I'd rather we don't start adding CI specific env variables as dependencies of these scripts. My end goal with those is to be able to call them for developers as well in order to setup or update their environment.

We can fix this another way, this keeping `cidir`, but removing the leading `.` I believe.

.ci/install_runtime.sh
devimc: yes, ```${runtime_config_file}``` is used to make the message clearer

.ci/setup.sh
devimc: sounds good, but we need more ```setup_env_*``` scripts

.ci/lib.sh
devimc: fixing, thanks

devimc:
@chavafg  @jodh-intel  @dlespiau fixed , can we merge it?

chavafg:
lgtm

PR #54, opened 2017-06-09
Run command test

devimc:
Implement tests for the run command


bundle.go
```diff
@@ -0,0 +1,151 @@
...
+	err = bundle.Save()
+	if err != nil {
+		return nil, err
+	}
+
+	return bundle, nil
+}
+
+// createRootfs creates a rootfs in the specific bundlePath
+func createRootfs(bundlePath string) error {
```
jodh-intel: Maybe return an error immediately if `bundlePath == ""`?

bundle.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Save to disk the Config
+func (b *Bundle) Save() error {
+	content, err := json.Marshal(b.Config)
+	if err != nil {
+		return fmt.Errorf("failed to marshal config: %s", err)
+	}
+
+	configFile := filepath.Join(b.Path, "config.json")
+	err = ioutil.WriteFile(configFile, content, 0664)
```
jodh-intel: Shouldn't that be `0644`? Related:

- What user do we expect this code to run as?
- Do we need to assert that?

devimc: yes ```0644``` is ok

> What user do we expect this code to run as?

I think only root is able to start containers

> Do we need to assert that?

I do not think so

bundle.go
devimc: fixing
thanks

chavafg:
lgtm

devimc:
depends on https://github.com/clearcontainers/runtime/pull/209

PR #61, opened 2017-06-14
Fetch branches tool

### LINKED ISSUE #77: Add fetch branches tool
The fetch branches tool that will be in charge of checking the branch where a pull request needs to be tested it.

GabyCT:
This is the commit of the fetch branches tool that
will be in charge of checking the branch where a pull request
needs to be tested it.

Fixes #77 

### DIFF
--- a/cmd/fetchbranches/CI.go
+++ b/cmd/fetchbranches/CI.go
@@ -1,0 +1,41 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Semaphore environment variables https://semaphoreci.com/docs/available-environment-variables.html
+
+package main
+
+import "os"
+
+// CI Continous Integration
+type CI interface {
+	// GetPR obtains the pull request number
+	GetPR() (*pr, error)
+}
+
+// Semaphore environment variables
+const (
+	ciEnvar   = "CI"
+	toolEnvar = "SEMAPHORE"
+)
+
+// NewCI verifies that the testing environment is using Semaphore
+func NewCI() CI {
+	ci := os.Getenv(ciEnvar)
+	semaphore := os.Getenv(toolEnvar)
+	if ci == "true" && semaphore == "true" {
+		return &SemaphoreCI{}
+	}
+	return nil
+}
--- a/cmd/fetchbranches/CI_test.go
+++ b/cmd/fetchbranches/CI_test.go
@@ -1,0 +1,69 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+func TestNewCI(t *testing.T) {
+	os.Unsetenv(ciEnvar)
+
+	if ci := NewCI(); ci != nil {
+		t.Fatal("expected an error")
+	}
+
+	os.Setenv(ciEnvar, "false")
+	os.Unsetenv(toolEnvar)
+
+	if ci := NewCI(); ci != nil {
+		t.Fatal("expected an error")
+	}
+
+	os.Setenv(toolEnvar, "false")
+
+	if ci := NewCI(); ci != nil {
+		t.Fatal("expected an error")
+	}
+
+	os.Setenv(ciEnvar, "1")
+	os.Setenv(toolEnvar, "2")
+
+	if ci := NewCI(); ci != nil {
+		t.Fatal("expected an error")
+	}
+
+	os.Setenv(ciEnvar, "true")
+	os.Setenv(toolEnvar, "false")
+
+	if ci := NewCI(); ci != nil {
+		t.Fatal("expected an error")
+	}
+
+	os.Setenv(ciEnvar, "false")
+	os.Setenv(toolEnvar, "true")
+
+	if ci := NewCI(); ci != nil {
+		t.Fatal("expected an error")
+	}
+
+	os.Setenv(ciEnvar, "true")
+
+	if ci := NewCI(); ci == nil {
+		t.Fatal("expected an error")
+	}
+
+}
--- a/cmd/fetchbranches/SemaphoreCI.go
+++ b/cmd/fetchbranches/SemaphoreCI.go
@@ -1,0 +1,78 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Semaphore environment variables https://semaphoreci.com/docs/available-environment-variables.html
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+type SemaphoreCI struct{}
+
+const (
+	prEnvar   = "PULL_REQUEST_NUMBER"
+	repoEnvar = "SEMAPHORE_REPO_SLUG"
+)
+
+// GetPR retrieves the PR number from the Semaphore environment
+func (ci *SemaphoreCI) GetPR() (*pr, error) {
+	val := os.Getenv(prEnvar)
+
+	if val == "" {
+		return nil, fmt.Errorf("%v environment variable is empty", prEnvar)
+	}
+
+	number, err := strconv.Atoi(val)
+	if err != nil {
+		return nil, err
+	}
+
+	semaphoreRepository := os.Getenv(repoEnvar)
+
+	// Checks that repository name is not empty
+	if semaphoreRepository == "" {
+		return nil, fmt.Errorf("%v environment variable is empty", repoEnvar)
+	}
+
+	testRepository := strings.Split(semaphoreRepository, "/")
+
+	// Valid names should have and owner and a repository name
+	if len(testRepository) != 2 {
+		return nil, fmt.Errorf("invalid repository name: %v", semaphoreRepository)
+	}
+
+	ownerName := testRepository[0]
+	repositoryName := testRepository[1]
+
+	// Checks that the owner name is not empty
+	if ownerName == "" {
+		return nil, fmt.Errorf("missing an owner name: %v", semaphoreRepository)
+	}
+
+	// Checks that the repository name is not empty
+	if repositoryName == "" {
+		return nil, fmt.Errorf("missing a repository name: %v", semaphoreRepository)
+	}
+
+	return &pr{
+		number: number,
+		owner:  ownerName,
+		repo:   repositoryName,
+	}, nil
+}
--- a/cmd/fetchbranches/branch.go
+++ b/cmd/fetchbranches/branch.go
@@ -1,0 +1,70 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"regexp"
+	"strings"
+)
+
+// Branch is the structure providing specific parameter for the GetBranch execution
+type Branch struct {
+	Repo string
+	URL  string
+}
+
+// Valid tag names
+var tags = []string{"proxy", "shim", "tests", "hyperstart", "runtime"}
+
+// GetBranch retrieves the branches from the comment
+func GetBranch(comment string) []Branch {
+	var branches []Branch
+
+	for _, t := range tags {
+		// Obtaining full name (branch and url)
+
+		regex := regexp.MustCompile("branch_" + t + ":[[:space:]]*http[s]{0,1}:[[:graph:]]+")
+		index := regex.FindStringIndex(comment)
+		if len(index) == 0 {
+			continue
+		}
+
+		// Valid syntax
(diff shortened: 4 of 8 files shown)


cmd/fetchbranches/CI.go
```diff
@@ -0,0 +1,38 @@
...
+
+package main
+
+import "os"
+
+// CI Continous Integration
+type CI interface {
+	// GetPR will obtain the pull request number
+	GetPR() (int, error)
+	// GetRepository will obtain the owner and the repository name
```
devimc: missing space between previous function and comment =

cmd/fetchbranches/CI.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Semaphore environment variables https://semaphoreci.com/docs/available-environment-variables.html
+
+package main
+
+import "os"
+
+// CI Continous Integration
+type CI interface {
+	// GetPR will obtain the pull request number
```
devimc: please replace all ```will + verb``` with ```verb```
here is an example https://golang.org/src/os/exec.go?s=2947:3026#L118
```
// Signal sends a signal to the Process.
```

cmd/fetchbranches/CVR.go
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Semaphore environment variables https://semaphoreci.com/docs/available-environment-variables.html
+
+package main
+
+// CVR Control Version Repository
+type CVR interface {
+	// GetPRComments returns the comments of a specific pull request
```
devimc: comment should match with the name of the function, in this case ```GetPRAuthorComments```

cmd/fetchbranches/CVR.go
```diff
@@ -0,0 +1,38 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Semaphore environment variables https://semaphoreci.com/docs/available-environment-variables.html
+
+package main
+
+// CVR Control Version Repository
```
dlespiau: I'm not sure this interface adds a lot of value, do we expect to have a second implementation of it? A good rule is to avoid generalisation/interfaces when there's only one object implementing the interface (and no other object is on the horizon)

cmd/fetchbranches/SemaphoreCI.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+type SemaphoreCI struct{}
+
+// GetPR retrieves the PR number from the Semaphore environment
+func (ci *SemaphoreCI) GetPR() (int, error) {
```
dlespiau: How about introducing a PR object instead of returning an int? That object could have an `id` and a `URL` for instance with methods on it instead of global methods on the github oject: github.GetPRAuthorComments() becomes PR.GetAuthorComments()

jodh-intel: +1.

cmd/fetchbranches/CVR.go
jodh-intel: +1. We can always add something like this at a later date when needed.

cmd/fetchbranches/SemaphoreCI.go
```diff
@@ -0,0 +1,69 @@
...
+	val := os.Getenv("PULL_REQUEST_NUMBER")
+
+	if val == "" {
+		return -1, fmt.Errorf("pull_request_number environment variable is empty")
```
jodh-intel: I'd keep the original envvar name rather than lower-casing it as they are distinct variables. In fact, something like:

```
envVar := "PULL_REQUEST_NUMBER"
val := os.Getenv(envVar)
if val == "" {
		return -1, fmt.Errorf("%v environment variable is empty", envVar)
}
```

cmd/fetchbranches/SemaphoreCI.go
```diff
@@ -0,0 +1,69 @@
...
+	return strconv.Atoi(val)
+}
+
+// GetRepository retrieves the owner and the repository name from the Semaphore environment
+func (ci *SemaphoreCI) GetRepository() (string, string, error) {
+	semaphoreRepository := os.Getenv("SEMAPHORE_REPO_SLUG")
+
+	// Checks that repository name is not empty
+	if semaphoreRepository == "" {
+		return "", "", fmt.Errorf("semaphore_repo_slug environment variable is empty")
```
jodh-intel: Same comment as above - `semaphore_repo_slug` is technically a different environment variable to `SEMAPHORE_REPO_SLUG`.

cmd/fetchbranches/SemaphoreCI.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	testRepository := strings.Split(semaphoreRepository, "/")
+
+	// Valid names should have and owner and a repository name
+	if len(testRepository) != 2 {
+		return "", "", fmt.Errorf("this is not a valid repository name")
```
jodh-intel: I'd make the returned error something like:

```
fmt.Errorf("invalid repository name: %v", semaphoreRepository)
```

cmd/fetchbranches/SemaphoreCI.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	ownerName := testRepository[0]
+	repositoryName := testRepository[1]
+
+	// Checks that the owner name is not empty
+	if ownerName == "" {
+		return "", "", fmt.Errorf("missing an owner name")
```
jodh-intel: Again, I'd include the value of `semaphoreRepository` in the error to help with debugging the issue.

cmd/fetchbranches/SemaphoreCI.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	// Checks that the repository name is not empty
+	if repositoryName == "" {
+		return "", "", fmt.Errorf("missing a repository name")
```
jodh-intel: Same comment as above.

cmd/fetchbranches/branch.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, t := range tags {
+		// Obtaining full name (branch and url)
+		regex := regexp.MustCompile("branch_" + t + ":[[:space:]]*http[s]{0,1}:[[:graph:]]+")
+		index := regex.FindStringIndex(comment)
+		if len(index) == 0 {
+			continue
+		}
+
+		// i.e branch_test: https://xxxx.com/
+		tag := comment[index[0]:index[1]]
```
jodh-intel: If `len(index) == 1`, this might cause a crash as `index[1]` won't be valid.

cmd/fetchbranches/branch.go
```diff
@@ -0,0 +1,57 @@
...
+// Valid tag names
+var tags = []string{"proxy", "shim", "tests", "hyperstart", "runtime"}
+
+// GetBranch retrieves the branches from the pull request
+func GetBranch(comment string) []Branch {
+	var branches []Branch
+
+	for _, t := range tags {
+		// Obtaining full name (branch and url)
+		regex := regexp.MustCompile("branch_" + t + ":[[:space:]]*http[s]{0,1}:[[:graph:]]+")
```
jodh-intel: Why not just use https://golang.org/pkg/net/url/ for URL parsing?

cmd/fetchbranches/main.go
```diff
@@ -0,0 +1,113 @@
...
+			os.Exit(1)
+		}
+
+		// Obtains the name of the branch i.e. topic/tests
+		// from the URL
+		wordKey := "/tree/"
+		i := strings.Index(branchURL, wordKey)
+
+		if i == -1 {
+			fmt.Printf("the word tree was not found\n")
```
jodh-intel: This is rather an amusing error message. It's a little too literal though :-)

I'd say something like:

```
fmt.Printf("Failed to establish branch name from URL: %v\n", branchURL)
```

cmd/fetchbranches/main.go
```diff
@@ -0,0 +1,113 @@
...
+		fetchDirectory := filepath.Join(fullDirectory, owner, branchRepo)
+		os.Chdir(fetchDirectory)
+
+		cmd = exec.Command("git", "fetch", "origin")
+		nameFetch := filepath.Join("origin", branchName)
+		cmd = exec.Command("git", "checkout", "-b", branchName, nameFetch)
+		err = cmd.Run()
+
+		if err != nil {
+			fmt.Printf("it was not possible to fetch the branch\n")
```
jodh-intel: Which branch? I'd add that to the message.

cmd/fetchbranches/main.go
```diff
@@ -0,0 +1,113 @@
...
+	if ci == nil {
+		fmt.Fprintf(os.Stderr, "CI not found\n")
+		os.Exit(1)
+	}
+
+	// Shows PR number
+	pr, err := ci.GetPR()
+
+	if err != nil {
+		fmt.Printf("%s\n", err)
```
jodh-intel: Just a general point. It's often useful to use a set of standard prefixes for all logging messages. For example, for all the `Printf()` calls in this function, I'd be tempted to do something like:

```
fmt.Printf("ERROR: %v\n", err)
```

Alternatively, you could use a logging library like https://godoc.org/github.com/sirupsen/logrus which will add in the appropriate tags based on the log level. These prefixes make the output much more consistent and  "consumable".

cmd/fetchbranches/branch.go
```diff
@@ -0,0 +1,57 @@
...
+// Branch is the structure providing specific parameter for the GetBranch execution
+type Branch struct {
+	Repo string
+	URL  string
+}
+
+// Valid tag names
+var tags = []string{"proxy", "shim", "tests", "hyperstart", "runtime"}
+
+// GetBranch retrieves the branches from the pull request
```
jodh-intel: This comment is rather confusing as the function parameter is `comment`, not `pullRequest`.

cmd/fetchbranches/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func main() {
```
jodh-intel: This function is quite long - you might want to refactor it a little.

Another thing to think about is adding some unit tests for functions like `GetBranch()`, `GetPR()`, `GetRepository()` and `NewCI()`.

You might find some useful bits in https://github.com/clearcontainers/tests/blob/master/cmd/checkcommits/checkcommits_test.go as that too has to query Semaphore environment variables so I had to add tests for all sorts of interesting scenarios ;-)

GabyCT:
@dlespiau  and @jodh-intel thanks for the feedback, I will work on the changes :)

GabyCT:
@dlespiau and @jodh-intel changes were applied. Thanks

cmd/fetchbranches/branch.go
```diff
@@ -0,0 +1,59 @@
...
+		if len(index) == 0 {
+			continue
+		}
+
+		// Valid sintaxis
+		// i.e branch_test: https://xxxx.com/
+		tag := comment[index[0]:index[1]]
+
+		// Looking for the beggining of the URL (http)
+		i := strings.Index(tag, "http")
```
jodh-intel: This looks fine, but I'm wondering:

- If there is an easier way. I came across https://github.com/mvdan/xurls which might be worth a look.
- About validation. It wouldn't hurt to call:
  ```
  possibleURL := tag[i:]

  // ensure it really is a valid-looking URL
  _, err := url.Parse(possibleURL)
  if err != nil {
      fmt.Fprintf(os.Stderr, "WARNING: ignoring url: %v\n", possibleURL)
      continue
  }
  ```

@dlespiau - any thoughts on this?

cmd/fetchbranches/branch.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, t := range tags {
+		// Obtaining full name (branch and url)
+
+		regex := regexp.MustCompile("branch_" + t + ":[[:space:]]*http[s]{0,1}:[[:graph:]]+")
+		index := regex.FindStringIndex(comment)
+		if len(index) == 0 {
+			continue
+		}
+
+		// Valid sintaxis
```
jodh-intel: Typo: "Valid syntax".

cmd/fetchbranches/branch.go
```diff
@@ -0,0 +1,59 @@
...
+		// i.e branch_test: https://xxxx.com/
+		tag := comment[index[0]:index[1]]
+
+		// Looking for the beggining of the URL (http)
```
jodh-intel: Typo: "beginning".

cmd/fetchbranches/branch_test.go
```diff
@@ -0,0 +1,70 @@
...
+// limitations under the License.
+
+package main
+
+import "testing"
+
+func TestGetBranch(t *testing.T) {
+	comment := "branch_hello: https://github.com"
+	if branches := GetBranch(comment); len(branches) != 0 {
+		t.Fatal("expected lenght 0")
```
jodh-intel: Typo: "length".

cmd/fetchbranches/main.go
```diff
@@ -0,0 +1,111 @@
...
+	// Checks that GOPATH exists
+	envVar := "GOPATH"
+	gopath := os.Getenv(envVar)
+	if gopath == "" {
+		return fmt.Errorf("%v environment variable is empty", envVar)
+	}
+
+	for branchRepo, branchURL := range fetchMap {
+		// Clones the repository that is in the PR (branch name)
+		urlCVR := "github.com"
```
jodh-intel: This can be `const` (and I'd add it to the top of the file as a global to make it stand out).

cmd/fetchbranches/main.go
```diff
@@ -0,0 +1,111 @@
...
+		cmd := exec.Command("go", "get", "-d", testRepository)
+		err := cmd.Run()
+
+		if err != nil {
+			return fmt.Errorf("Failed to clone the repository: %v", testRepository)
+		}
+
+		// Obtains the name of the branch i.e. topic/tests
+		// from the URL
+		wordKey := "/tree/"
```
jodh-intel: This can also be `const`.

cmd/fetchbranches/main.go
```diff
@@ -0,0 +1,111 @@
...
+		i := strings.Index(branchURL, wordKey)
+
+		if i == -1 {
+			return fmt.Errorf("Failed to establish branch name from URL: %v", branchURL)
+		}
+
+		branchName := (branchURL[i+len(wordKey):])
+
+		// Performs the fetch
+		githubDirectory := "/src/github.com/"
```
jodh-intel: This can also be `const` (and again I'd add this one to the top of the file as a global to make it stand out).

cmd/fetchbranches/pr.go
```diff
@@ -0,0 +1,68 @@
...
+	"context"
+	"fmt"
+
+	"github.com/google/go-github/github"
+)
+
+type pr struct {
+	number int
+	owner  string
+	repo   string
```
jodh-intel: This could do with a comment explaining the format of this field since it's not obvious like the other two.

cmd/fetchbranches/CI_test.go
```diff
@@ -0,0 +1,69 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+func TestNewCI(t *testing.T) {
```
jodh-intel: Nice set of tests! :smile:

GabyCT:
@jodh-intel thanks for the feedback, I did the changes. Regarding of using  https://github.com/mvdan/xurls, I did it and even implemented it however when I am doing the vendoring I see that it will upload around 11M as it is using or has many dependencies from google.golang.org, golang.org, and github.com. So, I decided not to use it :(

jodh-intel:
Hi @GabyCT - how about just using `url.Parse()`?

GabyCT:
@jodh-intel , I tried using url.Parse however when we have the combination like branch_tests:https://github.com, it is complaining that the first path segment in URL cannot contain colon. So at the end, it did not help me to extract the URL.

jodh-intel:
Hi @GabyCT - sorry, what I meant was, split the `branch_tests:https://github.com` string apart as you are already doing, and once you've got what you think it the URL part of it, call:

```
_, err := url.Parse(urlPart)
```

Then check `err`. If it's `nil`, there's a good chance we really do have a URL. It's just a "safety-check" for verifying what we *think* is a URL actually probably is ;-)

GabyCT:
@jodh-intel, ohhh got it :) thanks changes applied

jodh-intel:
lgtm

devimc:
lgtm

PR #68, opened 2017-06-15
Makefile: Fix Semaphore CI failing because of CRI-O tests

sboeuf:
Because we get Semaphore CI hanging sometimes when trying to build everything from a "make localintegration" for the testing of CRI-O, this patch only calls into the test itself, without the need for rebuilding everything.


chavafg:
lgtm

PR #72, opened 2017-06-16
tests: check if a VM is still running

### LINKED ISSUE #69: functional: tests do not check if a VM is still running after delete a container
Implement function to detect if a VM is still running after delete a container

devimc:
implement function to check if the VM of a container is still running

fixes #69

### DIFF
--- a/vm.go
+++ b/vm.go
@@ -1,0 +1,55 @@
+/// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+)
+
+const procPath = "/proc"
+
+var errFound = errors.New("found")
+
+// IsVMRunning looks in /proc for a hypervisor process that contains
+// the containerID in its command line
+func IsVMRunning(containerID string) bool {
+	err := filepath.Walk(procPath, func(path string, info os.FileInfo, err error) error {
+		if !info.IsDir() {
+			return filepath.SkipDir
+		}
+
+		content, err := ioutil.ReadFile(filepath.Join(path, "cmdline"))
+		if err != nil {
+			return filepath.SkipDir
+		}
+
+		hypervisorRegexs := []string{".*/qemu.*-name.*" + containerID + ".*-qmp.*unix:.*/" + containerID + "/.*"}
+
+		for _, regex := range hypervisorRegexs {
+			matcher := regexp.MustCompile(regex)
+			if matcher.MatchString(string(content)) {
+				return errFound
+			}
+		}
+
+		return nil
+	})
+
+	return err == errFound
+}


vm.go
```diff
@@ -0,0 +1,60 @@
...
+// IsVMRunning looks in /proc for a hypervisor process that contains
+// the containerID in its command line
+func IsVMRunning(containerID string) bool {
+	err := filepath.Walk(procPath, func(path string, info os.FileInfo, err error) error {
+		if !info.IsDir() {
+			return filepath.SkipDir
+		}
+
+		cmdlineFile := filepath.Join(path, "cmdline")
+		if _, err := os.Stat(cmdlineFile); os.IsNotExist(err) {
```
jodh-intel: I don't think the `Stat()` is required as `ReadFile()` will handle `ENOENT` itself.

vm.go
```diff
@@ -0,0 +1,60 @@
...
+			return filepath.SkipDir
+		}
+
+		content, err := ioutil.ReadFile(cmdlineFile)
+		if err != nil {
+			return filepath.SkipDir
+		}
+
+		for _, hypervisor := range hypervisors {
+			if bytes.Contains(content, []byte(hypervisor)) &&
```
jodh-intel: I don't think this is going to be reliable since wouldn't this match my `grep` process if I ran `ps -ef|grep qemu`?

The problem is that you've got to find only those hypervisor processes that we care about (what if the system had `qemu-system_x86_64` and `qemu-lite-system_x86_64` processes running simultaneously?) As such, I wonder if this program should be told the full path to the hypervisor that is actually being used (which might be, say, `/usr/local/bin/qemu-lite-system_x86_64`).

vm.go
grahamwhaley: this may sound silly - something from way in my past - what if it is a directory, not a file? :-) A stat can tell you that as well ;-)

vm.go
```diff
@@ -0,0 +1,60 @@
...
+				bytes.Contains(content, []byte(containerID)) {
```
jodh-intel: Oh - hadn't actually noticed the second part of the test :) I guess you could tighten this check up by looking for `"-name " + containerID`.

vm.go
devimc: you're right I can remove this test

vm.go
devimc: I'm looking for a ```qemu``` process that contains ```containerID``` in its command line

vm.go
devimc: runtime append 'pod-' to the name (```-name pod-$CONTAINERID```) but I'm not sure if it does that all the time

vm.go
devimc: I'm not sure if there is another way to know the pid of the VM associated with the container

vm.go
jodh-intel: Maybe you could test for all of the following in `cmdline`:

- Search for a line containing exactly `-name` followed by a line containing `containerID`.
- Search for a line which is exactly `-qmp` followed by a line containing `virtcontainers/pods/containerID/`.

I've just checked and it looks like we'd actually need to change ciao to allow virtcontainers to store the pid and make that available to the runtime.

@dlespiau, @sameo  - any thoughts on how we could make this test more reliable?

dlespiau: - /proc/$pid/comm could be used to make sure that's a qemu process.
- That looks like a typical function that could use some unit testing with different inputs. Looks like some refactoring would be needed though.

devimc: @dlespiau yes, ```/proc/$pid/comm``` returns the name of the process, the problem here is that we have to find a way to check whether the VM is associated with the container
I think that with ```-name pod-$CONTAINERID``` is enough

dlespiau: I guess my point was `comm` may be better than `cmdline` to make sure that's a qemu process. Then we're sure that matching the container ID is done on a qemu process. `pod-$id` sounds good enough to me atm.

devimc:
@chavafg  @jodh-intel can we merge this?

jodh-intel:
I think there is still a small chance of collision, but you'd have to actively interfere on the system to cause it. Hence,

lgtm.

dlespiau:
we should get into the habit of adding uni-tests, even for simple-looking changes like these. We may need to evolve the feature in the future and having unit-tests helps with that. Anyway, ...

lgtm

vm.go
```diff
@@ -0,0 +1,55 @@
...
+		if !info.IsDir() {
+			return filepath.SkipDir
+		}
+
+		content, err := ioutil.ReadFile(filepath.Join(path, "cmdline"))
+		if err != nil {
+			return filepath.SkipDir
+		}
+
+		hypervisorRegexs := []string{".*/qemu.*-name.*" + containerID + ".*-qmp.*unix:.*/" + containerID + "/.*"}
```
jodh-intel: For better performance, you might want to make that global and create a slice of regexp's by calling `regexp.MustCompile()` in `init()` as othewise, you'll be re-compiling these regexps for every file.

devimc: I can't because ```containerID``` is provided at each call to this function

jodh-intel: yup - ignore that comment! ;)

jodh-intel:
lgtm

chavafg:
lgtm

PR #73, opened 2017-06-16
CI: Override runtime build defaults.

jodh-intel:
Install the runtime with modified defaults to ensure it uses the
expected paths. This requires a few variables to be set that will
override the defaults in the runtimes Makefile along with running
"make -e" to tell make to only set variables if they are not
already set.

This change is required to support:

- https://github.com/clearcontainers/runtime/pull/225.


chavafg:
lgtm

PR #76, opened 2017-06-20
CI: Check if system supports Clear Containers.

jodh-intel:
After installing the runtime, run its "cc-check" command to ensure the
system is capable of running Clear Containers.


chavafg:
lgtm

PR #78, opened 2017-06-20
Makefile: Re-enable and fix CRI-O tests on Semaphore CI

sboeuf:
This patch re-enables the CRI-O tests because they are now fixed by the recent introduction of a specific variable into the bats. By setting the environment variable "CRI_STORAGE", we can now specify which storage driver we want to use.

In case of Semaphore, because it is running Ubuntu 14.04 on a 3.13 kernel, we have to set aufs as the storage driver, otherwise we get some issues using devicemapper.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ crio:
 	bash .ci/install_bats.sh
 	cp $(PWD)/integration/cri-o/crio.bats ${CRIO_REPO_PATH}/test/
 	cd ${CRIO_REPO_PATH} && \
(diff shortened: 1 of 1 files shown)


sboeuf:
We need to wait for https://github.com/kubernetes-incubator/cri-o/pull/614 to be merged before we can merge this one.

sboeuf:
Fixes https://github.com/clearcontainers/runtime/issue/215

sboeuf:
@chavafg please take a look.

chavafg:
lgtm, even the time it takes to run crio seems faster :)

PR #86, opened 2017-06-26
RFC: metrics: add framework and first metrics test

grahamwhaley:
Add the basic metrics framework and the first metrics test.

### DIFF
--- a/metrics/lib/send_results.sh
+++ b/metrics/lib/send_results.sh
@@ -1,0 +1,383 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# This script will store the results into csv files in the
+# metrics 'results' directory.
+# The csv file names are derived from the reported 'test name', and all
+# results for a single 'test name' are collated (appended) into a single csv.
+
+# General env
+MYSELF="${0##*/}"
+SCRIPT_PATH=$(dirname $(readlink -f $0))
+source "${SCRIPT_PATH}/common.bash"
+
+# Override the RESULT_DIR from the test-common.bash for the moment.
+# Once all all tests are migrated to use this script, remove the
+# RESULT_DIR definition from test-common.bash, as all results should
+# be then stored via this script
+RESULT_DIR="${SCRIPT_PATH}/../results"
+
+GROUP="PNP"
+COR_DEFAULT_PATH="/usr/bin"
+COR_DEFAULT_NAME="cc-oci-runtime"
+RUNC_DEFAULT_PATH="/usr/bin"
+RUNC_DEFAULT_NAME="docker-runc"
+CC_RUNTIME_DEFAULT_PATH="/usr/local/bin"
+CC_RUNTIME_DEFAULT_NAME="cc-runtime"
+CC_DEFAULT_IMG_PATH="/usr/share/clear-containers"
+CONTAINERS_IMG_SYSTEM="/usr/share/clear-containers/clear-containers.img"
+CONTAINERS_IMG_GUESS="${CC_DEFAULT_IMG_PATH}/clear-containers.img"
+CONTAINER_KERNEL_SYSTEM="/usr/share/clear-containers/vmlinux.container"
+CONTAINER_KERNEL_GUESS="${CC_DEFAULT_IMG_PATH}/vmlinux.container"
+SYSTEM_CLEAR_RELEASE_FILE="/usr/lib/os-release"
+SYSTEM_CENTOS_RELEASE_FILE="/etc/os-release"
+
+declare TEST_NAME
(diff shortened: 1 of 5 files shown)


grahamwhaley:
@dlespiau @chavafg @devimc @MarioCarrilloA 
I've added this at the moment as RFC/DNM so you can have a look over and comment.
We know long term we want to maybe move away or enhance the CSV usage, and we will be adding more tests and the code that checks the results are within expected limits rsn.
If you have any major structural issues (the subdir it lives in, the place it is in the Makefile etc.) then shout.
Otherwise, we'll drop the RFC/DNM, merge it, and then build upon it.

metrics/lib/send_results.sh
```diff
@@ -0,0 +1,337 @@
...
+	echo "$model ($num_cores $com)"
+}
+
+# Locate the cor runtime binary. If unknown or unfound, return nothing
+function locate_runtime()
+{
+	# For now use the build installed path for the runtime
+	# Ultimately, it would be good if we could ask/extract from
+	# docker itself where the runtime it is using lives
+	if [ -f "${bindir}/cc-oci-runtime" ];then
```
jodh-intel: Since this repo is used by CC 3.0, shouldn't this either be looking for `cc-runtime`, or be able to be configured for either runtime?

grahamwhaley: Ah, indeed - good spot. In fact it is slightly worse than that - the RUNTIME option is passed in to the test scripts from the top level launcher, so really that should be passed on/down somehow to show what we actually ran. I'll look for a nice way to do that.

grahamwhaley: and then it is worse than that. We are using the RUNTIME name to find the commit ID the RUNTIME was built from (for precise recording of  what we tested). To be really correct we need to ascertain exactly which runtime binary docker is using, but there is no way to actually ask docker for that info (sadly `docker info` does not (currently) give that info). Mumble.
I'm probably going to have to do some interim fix that looks at the $RUNTIME passed in from the test Makefile, tries to locate the relevant binary in the default place, and use that. Later on I think maybe the CI system will pass in the actual commit ID info when it invokes us, at which point we can use that for when running under a CI.

grahamwhaley: @jodh-intel have updated with functions that try to find the runtimes for cor, runc and cc-runtime and extract the commit from the version info. Sure, it is still imperfect as it cannot guarantee the runtime it finds is the one docker is actually configured to use - but, in the case of our CI that should be the right one afaik, and if/when the CI system can pass in more accurate info or we can extract that from docker, I think this will do to start with, and then we will fix if we find it is locating the wrong runtime under the CI.

metrics/lib/send_results.sh
```diff
@@ -0,0 +1,424 @@
...
+	if [ -f "${CC_DEFAULT_RUNTIME_PATH}" ];then
+		runtime="${CC_DEFAULT_RUNTIME_PATH}"
+	else
+		# And if we cannot find it at the default, can we find it
+		# in the path
+		runtime="$(which cc-oci-runtime)"
+	fi
+
+	if [ -f "$runtime" ];then
+		result="$($runtime -v | grep commit | sed s/"commit: "//g)"
```
jodh-intel: This doesn't give me the commit for all the possibilities, although the following does:

```
$runtime -v | grep commit | sed "s/commit.*://g"|awk '{print $1}'
```

metrics/lib/send_results.sh
```diff
@@ -0,0 +1,424 @@
...
+	# this is where we would grab and use that info.
+
+
+	# See if runtime is in the default place
+	if [ -f "${CC_DEFAULT_RUNTIME_PATH}" ];then
+		runtime="${CC_DEFAULT_RUNTIME_PATH}"
+	else
+		# And if we cannot find it at the default, can we find it
+		# in the path
+		runtime="$(which cc-oci-runtime)"
```
jodh-intel: It would be nice if:

```
CC_DEFAULT_RUNTIME="cc-oci-runtime"
CC_DEFAULT_RUNTIME_PATH="/usr/bin/${CC_DEFAULT_RUNTIME}"
```

Then, you could say `runtime=$CC_DEFAULT_RUNTIME` here.

metrics/lib/send_results.sh
```diff
@@ -0,0 +1,424 @@
...
+
+	if (( $num_cores > 1 ));then
+		com="cores"
+	fi
+
+	echo "$model ($num_cores $com)"
+}
+
+
+function get_cor_version()
```
jodh-intel: s/get_cor_version/get_runtime_version/ ?

metrics/lib/send_results.sh
```diff
@@ -0,0 +1,424 @@
...
+	if [ -f "$runtime" ];then
+		result="$($runtime -v | grep commit | sed s/"commit: "//g)"
+	else
+		result="cor-unknown"
+	fi
+
+	echo "$result"
+}
+
+function get_cc_runtime_version()
```
jodh-intel: Oh - I see what's going on now ;) Couldn't we just have a single `get_runtime_version()` that accepted a `$RUNTIME`?

grahamwhaley: I just knew you were going to ask that (I should have taken a bet on it...). Yeah, I can use your magic sed'isms to reduce the amount of variation I have to pass in. Would have helped if you runtime guys could have all stuck to the same format ;-)
OK, update coming...

grahamwhaley:
OK @jodh-intel one more time for review...

jodh-intel:
lgtm

grahamwhaley:
Dropped DNM. @dlespiau , I know you may not be available for a bit - if you become the final review blocker then I'll probably drop you from the list, whilst leaving @devimc on.

devimc:
lgtm

PR #98, opened 2017-07-03
metrics: checkmetrics: Add the checkmetrics tool

grahamwhaley:
First addition of the `checkmetrics` tool, used for comparing CSV metrics results files against a TOML baseline.

Expect the tool to gain more features as we extend the coverage of the metrics and the CI system.


grahamwhaley:
Note - I'll work on adding some tests to this now - it should be quite testable both from the TOML and CSV pov.

cmd/checkmetrics/README.md
```diff
@@ -0,0 +1,124 @@
...
+files against a TOML file that contains baseline expectations for
+the results.
+
+`checkmetrics` checks for a matching CSV file for each entry in the
+TOML file. Failure to find a matching CSV is classified as a failure.
+
+`checkmetrics` will continue to process all entries in the TOML file
+and report its final results in a summary table.
+
+Each CSV metrics file must container at least one header line, with the
```
chavafg: s/container/contain/

cmd/checkmetrics/README.md
```diff
@@ -0,0 +1,124 @@
+# checkmetrics
+
+## Overview
+
+The `checkmetrics` tool is used to check the metrics CSV results
```
jodh-intel: It's worth defining `CSV` the first time it's used.

cmd/checkmetrics/README.md
```diff
@@ -0,0 +1,124 @@
...
+files against a TOML file that contains baseline expectations for
```
jodh-intel: Same for `TOML` - ref to wikipedia article?

cmd/checkmetrics/README.md
```diff
@@ -0,0 +1,124 @@
...
+`checkmetrics` checks for a matching CSV file for each entry in the
+TOML file. Failure to find a matching CSV is classified as a failure.
+
+`checkmetrics` will continue to process all entries in the TOML file
+and report its final results in a summary table.
+
+Each CSV metrics file must container at least one header line, with the
+`Results` in column 5, and at least one line containing a test result
+line (that is, the minimum CSV file consists of two lines, one header
+and one result value)
```
chavafg: period(.) at the end of the line?

cmd/checkmetrics/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+The `checkmetrics` tool is used to check the metrics CSV results
+files against a TOML file that contains baseline expectations for
+the results.
+
+`checkmetrics` checks for a matching CSV file for each entry in the
+TOML file. Failure to find a matching CSV is classified as a failure.
+
+`checkmetrics` will continue to process all entries in the TOML file
+and report its final results in a summary table.
```
jodh-intel: Is the summary table sent to a file or stdout?

cmd/checkmetrics/basefile.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/BurntSushi/toml"
+	log "github.com/Sirupsen/logrus"
+)
+
+// metric is the slice of Metrics imported from the TOML config file
```
jodh-intel: Conventionally with go, the type is documented immediately above it, so I think this comment needs to be immediately above the `metrics` slice below.

cmd/checkmetrics/compare.go
```diff
@@ -0,0 +1,96 @@
...
+		"Min",
+		"Mean",
+		"Max",
+		"Iters",
+		"SD",
+		"CoV")
+}
+
+// check takes a basefile metric record and a CVS file import record and checks
+// if the cvs file metrics pass the metrics comparison checks
```
jodh-intel: Two typos: should be `CSV` and `csv`. Plus missing period?

cmd/checkmetrics/compare.go
```diff
@@ -0,0 +1,96 @@
...
+// check returns a formatted string containing the results of the check.
+// The err return will be non-nil if the check fails
+func (mc *metricsCheck) check(m metrics, c csvRecord) (err error, summary string) {
+	var pass bool = true
```
jodh-intel: Nit: I'd make the default `false` as it's easy to forget to add `pass = false` when new tests are added. But by defaulting to fail, you'll get immediate feedback that you forgot to add `pass = true` ;-)

cmd/checkmetrics/compare.go
```diff
@@ -0,0 +1,96 @@
...
+// metricsCheck is a placeholder struct for us to attach the methods to and make
+// it clear they belong this grouping. Maybe there is a better way?
+type metricsCheck struct {
+}
+
+// titleFormat contains the format string for the table header
+// We lay out two format strings so we can align both the title and the
+// lines of the report to try and make it 'tabulised'
+//                        P/F: Name: Min   < Mean  < Max:  Iter: SD:   CoV %"
+var titleFormat string = "%4s: %20s: %7s < %7s < %7s: %5s: %7s: %7s %% "
```
jodh-intel: It would be easier to use https://golang.org/pkg/text/tabwriter/ as that handles most of the formatting complexity for you.

cmd/checkmetrics/csv.go
```diff
@@ -0,0 +1,125 @@
...
+	"errors"
+	"math"
+	"os"
+	"strconv"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/montanaflynn/stats"
+)
+
+// csvRecord holds the data for a single 'test'. Some of the data is imported
```
jodh-intel: > Some of the data is imported from the CSV file
You could make this clear by doing something like:

```
type csvRecordFromFile struct {
    Name  string
    Records  [][]string
    // ...
}

type csvRecord struct {
    csvRecordFromFile
    Mean          float64
    MinVal        float64
    // ...
}
```

cmd/checkmetrics/csv.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		log.Warnf("Failed to open csv file [%s]", filepath)
+		return err
+	}
+
+	defer f.Close()
+
+	r = csv.NewReader(f)
+
+	c.Records, _ = r.ReadAll()
```
jodh-intel: Why are you discarding the `error`?

cmd/checkmetrics/csv.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Check we have at least one header line and one value line
+	if len(c.Records) < 2 {
+		log.Errorf("File [%s] only has [%d] records, need at least 2, including header", filepath, len(c.Records))
```
jodh-intel: 2nd use of `len(c.Records)`, so maybe use a variable to avoid re-calculating.

cmd/checkmetrics/csv.go
```diff
@@ -0,0 +1,125 @@
...
+		return errors.New("Not enough records in file")
+	}
+
+	// Sanity check that the CSV file appears to have the Result column where
+	// we expect to find it
+	if c.Records[0][4] != "Result" {
```
jodh-intel: That might crash as you don't know how long `c.Records[0]` is I think?

cmd/checkmetrics/main.go
```diff
@@ -0,0 +1,191 @@
...
+
+		// Now we have both the baseline and the CSV data loaded,
+		// let's go compare them
+		var mc metricsCheck
+
+		err, summary := mc.check(m, thisCsv)
+		if err != nil {
+			log.Warnf("Check for [%s] failed", m.Name)
+			log.Warnf(" with [%s]", summary)
+			finalerror = errors.New("Fail")
```
jodh-intel: Won't you only ever see the last error generated, rather than the first?

cmd/checkmetrics/main.go
```diff
@@ -0,0 +1,191 @@
...
+			Name:  "basefile",
+			Usage: "path to baseline TOML metrics file",
+		},
+		cli.BoolFlag{
+			Name:  "debug",
+			Usage: "enable debug output in the log",
+		},
+		cli.StringFlag{
+			Name: "log",
+			//Value: "/dev/null",
```
jodh-intel: Debug that can be removed?

cmd/checkmetrics/compare.go
jodh-intel: Even better maybe: https://github.com/olekukonko/tablewriter.

cmd/checkmetrics/compare.go
grahamwhaley: I get the idea, but I'm not sure it fits this function (as it stands at present). Currently the function checks for a set of things you must pass, and sets `false` if you fail any. If we invert the checks then we'd be effectively collecting up 'did you pass one of the checks', which is the wrong notion here - we need to know if we've failed any of the checks, not passed any of them. I'll leave as is for now.

jodh-intel: ack.

cmd/checkmetrics/compare.go
grahamwhaley: I did wonder if there was a suitable package as I was writing it (you know how it goes - first was just one little whitespace alignment to make it nice... which then grows into a whole table format).
I'll go have a look at those options. One thing we may end up wanting is for #104 reporting tool to be able to parse the output of checkmetrics - in which case we may also add some arguments to output in a 'standard format'.

jodh-intel: Right. Here's one way you can do that: https://github.com/clearcontainers/runtime/blob/master/list.go#L124..L133

cmd/checkmetrics/csv.go
grahamwhaley: Yeah, that would be nicer. As it happens, it seems `Name` is un-used, so I've dropped that, and as that only leaves `Records` as something imported from the file, in this case I've commented and laid it out to make that clearer. A new struct for just `Records` now felt a little overkill

cmd/checkmetrics/compare.go
jodh-intel: Make that https://github.com/clearcontainers/runtime/blob/master/list.go#L120..L133.

cmd/checkmetrics/csv.go
grahamwhaley: Oversight. Fixed

cmd/checkmetrics/csv.go
grahamwhaley: Good catch - fixed

cmd/checkmetrics/csv.go
grahamwhaley: fixed (borderline I thought ;-)

cmd/checkmetrics/README.md
grahamwhaley: Fixed. Found another similar one in the help as well, and fixed here and in the code

cmd/checkmetrics/main.go
grahamwhaley: Not sure if you mean from check() or for the whole Metric range loop we are within...
For check(), we will print the failing summary in the log, and then summary is accumulated up for the final report. A failure is non-fatal, so we continue to process the rest of the Metrics in the range.
For the range covering Metrics, finalerror is just an indicator that we have seen at least one failure (I'll go rename that as finalErroFlag or similar now to make that clearer). The actual Failures end up being reported in the report generated by the accumulated `summary` results.
Does that clarify? If not, then elaborate again. Maybe I need to add a few comments in there to clarify?

grahamwhaley: Ah, I see, maybe you are referring to us returning finalerror from the routine. Right - well, all fails just end up with "Fail", and the details are actually in the summary. I'll see if I can make this all a little clearer with a bool to note we have seen an error and a nicer exit path.

grahamwhaley:
Issues addressed thanks @chavafg @jodh-intel 
Whilst updating I also found some static checker items to clean and some other bits and pieces. Subsequently, this is a moderate sized update.

cmd/checkmetrics/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+```
+Report Summary:
++------+-----------------+--------------------+----------------------------+---------+-------+-------+------+
+| P/F  |      NAME       |       FLOOR        |            MEAN            | CEILING | ITERS |  SD   | COV  |
++------+-----------------+--------------------+----------------------------+---------+-------+-------+------+
+| Pass | docker-run-time |              0.500 |                      1.214 |   1.500 |   100 | 0.036 | 2.96 |
+| Fail | docker-run-time |             10.000 |                      1.214 | 100.000 |   100 | 0.036 | 2.96 |
+| Fail | docker-run-time |              0.100 |                      1.214 |   0.500 |   100 | 0.036 | 2.96 |
+| Fail | short           | Failed to load CSV | Not enough records in file |
```
jodh-intel: The formatting looks a little odd here as this fail row only has 4 fields, but the table overall assumes 8. Maybe the error messages could be summarized below the table (although that would require and extra `#` column so you could match up the test number with its corresponding name).

jodh-intel: OOI, how would the user determine the failing CSV file from the summary?

cmd/checkmetrics/csv.go
```diff
@@ -0,0 +1,140 @@
...
+func (c *csvRecord) load(filepath string) error {
+	var err error
+	var f *os.File
+	var r *csv.Reader
+
+	log.Debugf("in csv load of [%s]", filepath)
+
+	f, err = os.Open(filepath)
+	if err != nil {
+		log.Warnf("Failed to open csv file [%s]", filepath)
```
jodh-intel: For all these log entries, you could instead do something like:

```
log.Warn(err)
return err
```

In fact, if you *don't* specify `--log=`, won't the output go to `stdout`/`stderr` anyway meaning you'd end up with 2 copies of the same error with either strategy?

cmd/checkmetrics/csv.go
```diff
@@ -0,0 +1,140 @@
...
+	Results       []float64 // Result array converted to floats
+	Iterations    int       // How many results did we gather
+	Mean          float64   // The 'average'
+	MinVal        float64   // Smallest value we saw
+	MaxVal        float64   // Largest value we saw
+	SD            float64   // Standard Deviation
+	CoV           float64   // Co-efficient of Variation
+}
+
+// load reads in a CSV 'Metrics' results file from the file path given
```
jodh-intel: I think it would be useful to see an example of a valid, minimal CSV results file somewhere (either here, or in the `README`).

jodh-intel:
One other thing - it would be good to add some unit tests. If not on this PR, could you raise an issue for adding them later so we don't forget?

jodh-intel:
Looks like your PR branch needs to be updated wrt master too btw.

cmd/checkmetrics/README.md
grahamwhaley: Yeah, I know the formatting is out - to do that right I'd have to add some sort of func to the compare/metricsCheck struct to either determine how many columns we need, or have it fill out/generate the line for us (probably preferable). It's a sort of nasty data dependency across modules that I don't like... as a malformed or missing CSV file should be rare, I was going with 'good enough' for now. I'll take a look and see if I can improve it.
The CSV file names are the same as the test name (with .csv extension), so the failing file can be determined from the name of the test that failed (in the example, 'short').

jodh-intel: Ok. But there are 3 `docker-run-time` results so does the example need updating, or are those 3 files from different directories or something?

grahamwhaley: Fair point - the example is from a contrived setup - I re-used the same CSV file for 3 lines, and changed the min/max boundary checks that were being performed - this is not a setup I can ever see the tool being used in in reality. I'll rework the test data to be more representative (less confusing), and add a CSV example to the docs.

grahamwhaley:
I mentioned I was adding the Unit tests already :-) https://github.com/clearcontainers/tests/pull/98#issuecomment-312684059  I'll open an Issue for that if/when the PR lands.

cmd/checkmetrics/csv.go
grahamwhaley: The `err` (at least sometimes) has less info - such as in this case it says `no such file or directory`, whereas the `Warnf` adds in the name of the actual file. So I think the `Warnf`'s can/do add more context information.
Yes, the log will go to stdout if you don't set the `--log` setting, but the `err` returns don't come out anywhere unless we do it by hand (apart from maybe the one returned out of 'main' I think) - that is, I don't think we have a double-printing issue.

I can see maybe something like `log.Warnf([%s][%s]", filepath, err)` could be more useful actually - I'll go look at the few Warn's there are and contemplate.

grahamwhaley:
Updated. I believe I have covered all the feedback. Main points:
 - reworked the table reporting so we now get 'full lines' in the error lines
 - improved the README to include more extensive examples of the TOML and CSV files

And a few more static checker items etc.

grahamwhaley:
But now I see we are out of sync with the master. I'll see if this needs a rebase tomorrow (along with some other PRs).

grahamwhaley:
Rebase done

cmd/checkmetrics/compare.go
```diff
@@ -0,0 +1,139 @@
...
+	log "github.com/Sirupsen/logrus"
+)
+
+// metricsCheck is a placeholder struct for us to attach the methods to and make
+// it clear they belong this grouping. Maybe there is a better way?
+type metricsCheck struct {
+}
+
+// reportTitleSlice returns the report table title row as a slice of strings
+func (mc metricsCheck) reportTitleSlice() []string {
```
devimc: I rather don't merge references (mc *metricsCheck) with values (mc metricsCheck)

grahamwhaley: Ah, that was a typo - nice spot. Now I have to go read up on what that actually does... and then fix it most likely

grahamwhaley: fixed and pushed

jodh-intel:
lgtm

grahamwhaley:
Hi @rcaballeromx @iphutch could you give docs are review please.
And we need one 'approver' ack as well still.

chavafg:
lgtm

devimc:
Hi @grahamwhaley 
you should vendor the dependencies

grahamwhaley:
@devimc vendoring done (with dep)
@iphutch fixes applied to README

PR #103, opened 2017-07-03
checkcommits static checks

grahamwhaley:
Fix up some items reported by the go static checkers.

### DIFF
--- a/cmd/checkcommits/checkcommits_test.go
+++ b/cmd/checkcommits/checkcommits_test.go
@@ -643,6 +643,9 @@ func TestGetCommitAndBranch(t *testing.T) {
 
 		// XXX: crucially, no arguments (to trigger the auto-detection)
 		commit, dstBranch, err := getCommitAndBranch([]string{}, []string{})
+		if err != nil {
+			t.Fatal(err)
+		}
 
 		if commit != d.expectedCommit {
 			t.Fatalf("Unexpected commit %v (%+v)", commit, d)
--- a/cmd/checkcommits/checkcommits_test.go
+++ b/cmd/checkcommits/checkcommits_test.go
@@ -44,33 +44,33 @@ var fixesPattern *regexp.Regexp
 var restoreSet map[string]TestEnvVal
 
 var travisPREnv = map[string]TestEnvVal{
-	"TRAVIS":                     TestEnvVal{"true", true},
-	"TRAVIS_BRANCH":              TestEnvVal{"master", true},
-	"TRAVIS_COMMIT":              TestEnvVal{"travis-commit", true},
-	"TRAVIS_PULL_REQUEST_BRANCH": TestEnvVal{"travis-pr", true},
+	"TRAVIS":                     {"true", true},
+	"TRAVIS_BRANCH":              {"master", true},
+	"TRAVIS_COMMIT":              {"travis-commit", true},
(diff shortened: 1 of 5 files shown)


grahamwhaley:
Hi @jodh-intel 
I've marked this as DNM, as there is another static check report that could do with fixing that is not in this series....
gocyclo (as invoked like https://github.com/clearcontainers/runtime/blob/master/.ci/go-static-checks.sh#L92) is complaining that the comlexity of getCommitAndBranch is '16', and we have the limit set to '15' :-(
I had a look at the function, and it's not a bad function, just happens to be full of a pretty linear series of checks. Any ideas on if there is a nice way to reduce the complexity of that, without somewhat artificially slicing it up?

And, of course, whilst here your input on the commits would be welcome - particularly some of the simplification ones, which whilst I have no doubt are correct, I'd like your input on if they do improve readability/clarify, or if they actually detract something.

grahamwhaley:
Updated. Thanks to @jodh-intel for pushing #106 - meant I could proceed and find the remaining items.
@GabyCT you might want to check the fetchbranches related ones (minor stuff)

jodh-intel:
Hi @grahamwhaley - looks like you've hit the Travis bug again, so it's another close-and-reopen-fresh-PR I'm afraid ;(

grahamwhaley:
@jodh-intel I think we are actually hitting a different sort of Travis issue. I did a minor commit message edit and re-pushed and now it has passed. Similar to the rebase I did yesterday to fix the same problem on a different PR.

cmd/checkcommits/checkcommits_test.go
```diff
@@ -642,7 +642,7 @@ func TestGetCommitAndBranch(t *testing.T) {
 		}
 
 		// XXX: crucially, no arguments (to trigger the auto-detection)
-		commit, dstBranch, err := getCommitAndBranch([]string{}, []string{})
+		commit, dstBranch, _ := getCommitAndBranch([]string{}, []string{})
```
jodh-intel: Sorry, just noticed this: this is an oversight - we should be retrieving the error, then:

```
if err != nil {
    t.Fatal(err)
}
```

cmd/fetchbranches/main.go
```diff
@@ -46,7 +46,7 @@ func main() {
 	fmt.Printf("PULL_REQUEST_NUMBER : %d, OWNER : %s, REPOSITORY: %s\n", pr.number, pr.owner, pr.repo)
 
 	// Shows author PR comments
-	comments, err := pr.GetAuthorComments()
+	comments, _ := pr.GetAuthorComments()
```
jodh-intel: Same as above - I think the original code was just missing a check on `err`.

cmd/checkcommits/checkcommits_test.go
grahamwhaley: ok, np, I'll rework. Yep, I did just pretty blindly apply the 'is ignored' ruling. Must remember to engage brain a little more next time

grahamwhaley: updated and re-pushed

grahamwhaley:
Rebased and pushed.
Also noted that I'd broken 'go test' in fetchbranches - now also fixed.

devimc:
lgtm

PR #104, opened 2017-07-04
metrics: Add PnP failure Report tool

MarioCarrilloA:
This tool reports by email the obtained PnP results from checktool execution, next it will send the results by email.

### DIFF
--- a/cmd/emailreport/README.md
+++ b/cmd/emailreport/README.md
@@ -1,0 +1,15 @@
+# PnP failure Report tool
+
+This tool obtains metrics results from `checkmetrics` tool, and sends
+the results out via email.
+The configuration of SMTP server and `checkmetrics` tool arguments are
+given by a configuration file in TOML format, an example TOML configuration
+file is provided in the file `example.toml`
+
+WARNING: The password for the SMTP authentication will be in TOML
+configuration text plane file.
+
+# Usage
+```bash
+$ ./emailreport -f <conf.toml>
+```
--- a/cmd/emailreport/mail.go
+++ b/cmd/emailreport/mail.go
@@ -1,0 +1,85 @@
+/// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"log"
+	"net/smtp"
+	"strings"
+)
+
+// HeaderRFC822 provides a header as a
+// string according the RFC822 Standard for ARPA
+// Internet Text Messages.
+func HeaderRFC822(conf Configuration) string {
+	var header string
+
+	if conf.Mail.From == "" {
+		log.Fatal("Sender not specified")
+	}
+
+	if conf.Mail.Subject == "" {
+		log.Fatal("Message subject not specified")
+	}
+
+	if len(conf.Mail.To) == 0 {
+		log.Fatal("Receipts not specified")
+	}
+
+	header += "From: " + conf.Mail.From + "\n"
+	header += "To: " + strings.Join(conf.Mail.To, ",") + "\n"
+
+	if len(conf.Mail.Cc) > 0 {
+		header += "cc: " + strings.Join(conf.Mail.Cc, ",") + "\n"
+	}
+
+	header += "Subject: " + conf.Mail.Subject + "\n\n"
+
+	return header
+}
+
+// SendByEmail allows to send the result output of
+// checkmetrics execution to a whitelist of emails.
+// This list is described by a TOML configuration file.
+func SendByEmail(conf Configuration, body string) {
+	var header string
+	var msg string
+	var auth smtp.Auth
+	var server string
+
+	header = HeaderRFC822(conf)
+	msg = header + body
+
+	auth = smtp.PlainAuth(
+		conf.Mail.Id,       // SMTP identity
+		conf.Mail.User,     // SMTP user
+		conf.Mail.Password, // SMTP Server password
+		conf.Mail.Smtp,     // SMTP server address
+	)
+
+	server = conf.Mail.Smtp + ":" + conf.Mail.Port
+
+	// Connection/Authentication step
+	err := smtp.SendMail(
+		server,         // SMTP server:port
+		auth,           // SMTP authentication
+		conf.Mail.From, // Sender
+		conf.Mail.To,   // Receipts
+		[]byte(msg),    // Message
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
--- a/cmd/emailreport/mail_test.go
+++ b/cmd/emailreport/mail_test.go
@@ -1,0 +1,41 @@
+/// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+
+	"github.com/BurntSushi/toml"
+)
+
+func TestHeaderRFC822(t *testing.T) {
+	var conf Configuration
+	var header string
+	var expectHdr string
+
+	expectHdr += "From: pnp@example.com\n"
+	expectHdr += "To: maintainer@example.com,owner@example.com\n"
+	expectHdr += "cc: contributor@example.com,developer@example.com\n"
+	expectHdr += "Subject: PnP failure Report <no-reply>\n\n"
+
+	if _, err := toml.DecodeFile("example.toml", &conf); err != nil {
+		t.Fail()
+	}
+
+	header = HeaderRFC822(conf)
+	if header != expectHdr {
+		t.Fail()
+	}
+}
--- a/cmd/emailreport/main.go
+++ b/cmd/emailreport/main.go
@@ -1,0 +1,115 @@
+/// Copyight (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"syscall"
+
+	"github.com/BurntSushi/toml"
+)
+
+// Configuration stores all necessary
+// information for checkmetrics execution
+// and sent output as a report by email.
+type Configuration struct {
+	Mail MailConf
+	Ck   Checkmetrics `toml:"checkmetrics"`
+}
+
+// MailConf stores the authentication
+// configuration for sending the report
+// by email. This information is parsed
+// from TOML input file.
+type MailConf struct {
+	Smtp     string   // SMTP server address
+	Port     string   // SMTP port
+	User     string   // SMTP user
+	Password string   // SMTP password
+	Id       string   // SMTP indetity
+	From     string   // Sender
+	Subject  string   // Email msg subject
+	To       []string // Receipts
+	Cc       []string // Msg copy
+}
+
+// Checkmetrics stores the arguments
+// for checkmetrics execution
+type Checkmetrics struct {
+	Cmd        string
+	Basefile   string
+	Metricsdir string
+}
+
+func main() {
+
+	var confFile string
+	var conf Configuration
+	var body string
+	var currentUid int
+	var ownerUid int
+
+	// Checkmetrics conf vars
+	var cmd string
+	var basefile string
+	var metricsdir string
+
+	flag.StringVar(&confFile, "f", "", "Configuration file")
+	flag.Parse()
+
+	// Get info about file
+	fileinfo, err := os.Stat(confFile)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Owner verification
+	currentUid = os.Geteuid()
+	fStat := fileinfo.Sys().(*syscall.Stat_t)
+	ownerUid = int(fStat.Uid)
+
+	if currentUid != ownerUid {
+		log.Fatal(currentUid,
+			" is not the owner of ", confFile,
+			" uid: ", ownerUid)
+	}
+
+	// Parsing TOML configuration file
+	if _, err := toml.DecodeFile(confFile, &conf); err != nil {
+		log.Fatal(err)
+	}
+
+	// Set checkmetrics configuration
+	cmd = conf.Ck.Cmd
+	basefile = "--basefile " + conf.Ck.Basefile
+	metricsdir = "--metricsdir " + conf.Ck.Metricsdir
+
+	// checkmetrics execution
+	out, err := exec.Command(cmd, basefile, metricsdir).Output()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if len(out) == 0 {
+		log.Fatal("no output from: " + cmd)
+	}
+
+	body = fmt.Sprintf("%s", out)
+	SendByEmail(conf, body)
+}
--- a/cmd/emailreport/example.toml
+++ b/cmd/emailreport/example.toml
@@ -1,0 +1,26 @@
+# This section specifies information for
+# a SMTP server authentication as well as the
+# sender and receipts of the report email.
+# This is in order to send the obtained results
+# from checkmetrics to an email whitelist.
+# WARNING: The password for the SMTP authentication
+# will be in this text plane file.
+[mail]
+smtp = "myserver.smtp.com"
+port = "25"
+user = ""
+identity = ""
+password = ""
+from = "pnp@example.com"
+subject = "PnP failure Report <no-reply>"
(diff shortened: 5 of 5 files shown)


jodh-intel:
Hi @MarioCarrilloA - please could you raise and issue, assign to yourself and add a `Fixes #XXX` comment in one of the commits for this PR to avoid `checkcommits` complaining under Travis.

metrics/report/README.md
```diff
@@ -0,0 +1,12 @@
+# PnP failure Report tool
+
+This tool reports by email the obtained PnP results from checktool
```
jodh-intel: What is "checktool" (and if it's a command, it should be rendered like: `checktool`).

metrics/report/example.toml
```diff
@@ -0,0 +1,23 @@
...
+subject = "PnP failure Report <no-reply>"
+to = [ "manteiner@example.com",
+	"owner@example.com" ]
+cc = [ "contributor@example.com",
+	"developer@example.com" ]
+
+# [checktool] section specifies the arguments for
+# Checktool execution.
+[checktool]
+basefile = "/home/checktool/conf.toml"
```
jodh-intel: The name `basefile` is a little odd as it appears to be a configuration file?

metrics/report/example.toml
```diff
@@ -0,0 +1,23 @@
...
+# from checktool to an email whitelist.
+[mail]
+smtp = "myserver.smtp.com"
+port = "25"
+user = ""
+identity = ""
+pass = ""
+from = "pnp@clear-containers.com"
+subject = "PnP failure Report <no-reply>"
+to = [ "manteiner@example.com",
```
jodh-intel: Typo: "maintainer".

metrics/report/example.toml
```diff
@@ -0,0 +1,23 @@
...
+# sender and receipts of the report email.
+# This is in order to send the obtained results
+# from checktool to an email whitelist.
+[mail]
+smtp = "myserver.smtp.com"
+port = "25"
+user = ""
+identity = ""
+pass = ""
+from = "pnp@clear-containers.com"
```
jodh-intel: I'd stick with the `@example.com` domain as that's recognised for what it is, whereas this might imply we own that domain.

metrics/report/example.toml
```diff
@@ -0,0 +1,23 @@
...
+subject = "PnP failure Report <no-reply>"
+to = [ "manteiner@example.com",
+	"owner@example.com" ]
+cc = [ "contributor@example.com",
+	"developer@example.com" ]
+
+# [checktool] section specifies the arguments for
```
jodh-intel: s/[checktool] section/this section/

metrics/report/mail.go
```diff
@@ -0,0 +1,87 @@
...
+// bodyMsg concatenates the output of
+// checktool in one string, this string
+// is the body of the email message.
+func bodyMsg(outRows []string) string {
+	var body string
+
+	body += "<h3>PnP Checktool results</h3>\n"
+
+	for _, row := range outRows {
+		body += row + "<br>"
```
jodh-intel: This doesn't appear to generate valid XHTML as you're only adding an opening tag.

metrics/report/mail.go
```diff
@@ -0,0 +1,87 @@
...
+	return body
+}
+
+// headerRFC822 provides a header as a
+// string according the RFC822 Standard for ARPA
+// Internet Text Messages.
+func headerRFC822(conf Configuration) string {
+	var header string
+
+	header += "From: " + conf.Mail.From + "\n"
```
jodh-intel: There doesn't appear to be any validation on all these `conf.*` values? I wonder if there are any standard golang packages that can perform the validation for us?

jodh-intel: It would be good to add some unit tests for atleast `main.go` (you might need to split `SendByEmail()` to allow it to be tested without actually having it call `smtp.SendMail()`).

metrics/report/README.md
grahamwhaley: that'll be `checkmetrics` on PR #98

jodh-intel: In which case, this doc should probably refer to it as "`checkmetrics`, the check tool" or similar (and this PR should block on #98).

metrics/report/example.toml
```diff
@@ -0,0 +1,23 @@
...
+# a SMTP server authentication as well as the
+# sender and receipts of the report email.
+# This is in order to send the obtained results
+# from checktool to an email whitelist.
+[mail]
+smtp = "myserver.smtp.com"
+port = "25"
+user = ""
+identity = ""
+pass = ""
```
MarioCarrilloA: Hi, which format do you suggest?

metrics/report/mail.go
MarioCarrilloA: Hi, it is a line break
https://www.w3schools.com/tags/tag_br.asp

metrics/report/example.toml
MarioCarrilloA: that sounds better :)

metrics/report/example.toml
MarioCarrilloA: ups...!!! I will fix it

metrics/report/README.md
MarioCarrilloA: perfect, I will fix it, thanks

metrics/report/mail.go
MarioCarrilloA: let's surf in golang packages docs about that validation, and check about unit tests :)

metrics/report/mail.go
jodh-intel: Sure, but it's an opening tag and there is no corresponding closing tag, so it will generate invalid xhtml/XML.

You can fix it by changing it to `<br/>`.

metrics/report/example.toml
jodh-intel: Storing passwords is something to avoid if at all possible due to all the potential security risks.

@grahamwhaley, @dlespiau - thoughts?

grahamwhaley: A little tricky...
We won't be exposing any passwords on the github repo (just to dismiss any concern there), as config files will be site and server specific, so we will just host examples in the actual code base.
Having said that, sure, a non-plain text option would be nice - but, what options do we have?:
 - this is for SMTP - which ultimately wants a plain text string I believe. Even if there are secure alternatives, we probably need to support vanilla SMTP by default.
 - Does anybody have any other ideas? We could have links to an encrypted password and the relevant keys, but that still leaves you with the problem that anybody who can get to the config file can probably also get to the decryption key.

Ultimately I think this may come down to 'the system owner has to ensure they secure that file enough' via permissions. I'm all open to other suggestions, using a keyring API etc. etc.

metrics/report/example.toml
```diff
@@ -0,0 +1,23 @@
...
+from = "pnp@clear-containers.com"
+subject = "PnP failure Report <no-reply>"
+to = [ "manteiner@example.com",
+	"owner@example.com" ]
+cc = [ "contributor@example.com",
+	"developer@example.com" ]
+
+# [checktool] section specifies the arguments for
+# Checktool execution.
+[checktool]
```
grahamwhaley: Given you are encoding the arguments to the tool, you could go one more step in flexibility and encode the path/name of the command as well.
I'll note the tool in question right now is called `checkmetrics` - you may or may not wish to name the TOML section appropriately.
Of course, if you make this a generic 'tool and argument' section, you will need to add the actual argument commands (such as `--metricsdir`) to the TOML as well.

metrics/report/README.md
```diff
@@ -0,0 +1,12 @@
+# PnP failure Report tool
```
grahamwhaley: Comment for all the files.
Given the other tools in this repo live under the `cmd` directory, I would consider moving this to `cmd/reporttool` to match.
Sure, it is metrics specific, but I didn't place `checkmetrics` in the metrics dir - so we should probably follow suit.

metrics/report/example.toml
```diff
@@ -0,0 +1,23 @@
+# [mail] section specifies information for
+# a SMTP server authentication as well as the
+# sender and receipts of the report email.
+# This is in order to send the obtained results
+# from checktool to an email whitelist.
+[mail]
```
grahamwhaley: btw, I like that you made the email section quite complete, and thus adaptable.

metrics/report/main.go
```diff
@@ -0,0 +1,88 @@
...
+	Mail MailConf
+	Ck   Checktool `toml:"checktool"`
+}
+
+// MailConf stores the authentication
+// configuration for sending the report
+// by email. This information is parsed
+// from TOML input file.
+type MailConf struct {
+	Smtp    string   // SMTP server address
```
grahamwhaley: Consider running https://github.com/clearcontainers/runtime/blob/master/.ci/go-static-checks.sh on the code base and fixing any reports. We will be enabling this for this repo soon, so they would have to be fixed at that point anyhow.
Just two `lint` reports I think right now both in this struct.

metrics/report/mail.go
MarioCarrilloA: oh yeah.... that is for XHTML, I will fix that

metrics/report/main.go
```diff
@@ -0,0 +1,88 @@
...
+	basefile = "--basefile " + conf.Ck.Basefile
+	metricsdir = "--metricsdir " + conf.Ck.Metricsdir
+
+	// Checktool execution
+	out, err := exec.Command(cmd, basefile, metricsdir).Output()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	outRows = strings.Split(string(out), "\n")
```
grahamwhaley: Just a heads up that `checkmetrics` just changed its output to a more formatted table using https://github.com/olekukonko/tablewriter (see
https://github.com/clearcontainers/tests/pull/98/files#diff-1b7dc2c4625a82d201f9557ff3035f82R94 for example output) - so we may need to check, or update later, to accommodate that.
As mentioned before, if we need a more parseable or email friendly format (although that table is OK I think), then we can add something like a `--json` option to `checkmetrics`.

metrics/report/example.toml
MarioCarrilloA: oh yeah, I just put that fake information as an example, but its better to follow "@example.com" in order to clarify that. great I will fix that

metrics/report/mail.go
```diff
@@ -0,0 +1,87 @@
...
+	"strings"
+)
+
+// bodyMsg concatenates the output of
+// checktool in one string, this string
+// is the body of the email message.
+func bodyMsg(outRows []string) string {
+	var body string
+
+	body += "<h3>PnP Checktool results</h3>\n"
```
grahamwhaley: I'd like us to consider if we need html emails here - would plain text not serve equally well at this point?
The report summary currently has no rich text embedded (although it would be nice if it could highlight the failure points more clearly) - and, unless we are currently encoding the summary table into a fixed width font (which I can't see in the code, but maybe I missed it or it is a subtlety), might this mess up the fixed width table formatting?

metrics/report/README.md
MarioCarrilloA: agree that make sense :)

metrics/report/main.go
MarioCarrilloA: sounds good, I just will take the output of checkmetrics tool as the body message.

metrics/report/mail.go
MarioCarrilloA: it's better plain text, I will fix that, thanks :)

metrics/report/example.toml
MarioCarrilloA: make sense I will add that.

metrics/report/main.go
MarioCarrilloA: I will check that tool and for lint I used gofmt :)

grahamwhaley: sure thing - the script invokes a number of tools (gocyclo, misspell, golint, ineffassign, vet, fmt). And note, it runs `fmt` with the -s 'simplify' option as well. Heh heh heh...

metrics/report/example.toml
jodh-intel: *If* we decide a plain text password is acceptable, we should document clearly the risk of using it (add a warning in the example config file and in the `README`) and ensure the tool refuses to run unless the config file is only readable by the user running the tool.

A couple of other points:

- I'd also rename the `pass` config option to `password` to avoid any possible confusion.
- Can we give this tool a more meaningful name than `report`?

chavafg:
lgtm

chavafg:
Hi @iphutch, @rcaballeromx, can you please check this PR?

grahamwhaley:
This too looks like it has fallen foul of a CRI-O build test failure. @chavafg @jodh-intel @dlespiau do we have any more insight into that failure type yet - presuming something changed, my best guess is could CRI-O have moved on and broken for us?

jodh-intel:
It looks like the same problem as on https://github.com/clearcontainers/shim/pull/44.

I think we might just need to install `btrfs-progs`. I'll raise a PR on `tests`...

jodh-intel:
Ref: https://github.com/clearcontainers/tests/pull/128.

cmd/emailreport/mail_test.go
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/BurntSushi/toml"
```
grahamwhaley: Convention is for the system packages to come first I believe:
https://github.com/golang/go/wiki/CodeReviewComments#imports

cmd/emailreport/main.go
```diff
@@ -0,0 +1,112 @@
...
+	var body string
+	var currentUid int
+	var ownerUid int
+
+	// Checkmetrics conf vars
+	var cmd string
+	var basefile string
+	var metricsdir string
+
+	flag.StringVar(&confFile, "f", "", "Configuration file")
```
grahamwhaley: Just a passing comment.
I see we add the example toml file and readme in commits after this one - as they sort of relate directly to and/or are required by this tool, normally you'd either submit them before or in the same commit as the tool itself to make things more sort of logical, iyswim.
It's not a problem here, and as they don't actually have a direct runtime dependence in this case we didn't 'break the build' between commits - hence, just a note and not a blocker at all.

cmd/emailreport/mail.go
```diff
@@ -0,0 +1,85 @@
...
+
+package main
+
+import (
+	"log"
+	"net/smtp"
+	"strings"
+)
+
+// headerRFC822 provides a header as a
```
grahamwhaley: nitpick: first word (headerRFC822) has a different capitalisation from the actual function name it is describing.

cmd/emailreport/mail.go
```diff
@@ -0,0 +1,85 @@
...
+		header += "cc: " + strings.Join(conf.Mail.Cc, ",") + "\n"
+	}
+
+	header += "Subject: " + conf.Mail.Subject + "\n\n"
+
+	return header
+}
+
+// SendByEmail allows to send the result output of
+// checkmetrics execution to a whitelist of emails.
```
grahamwhaley: I think we may have missed the intent/meaning of the whitelisting here, and how we are going to implement it. It's doesn't have to be a blocker, and I will open a new Issue to add the 'whitelist checking' as a next step to this tool.
The notion of the whitelist is that we should only send emails to a defined list of 'allowed' addresses. And we have chosen that rather than have an actual list in a file that defines legal destination email adresses, we will use the github 'organisation' list as the list to check (that is, we will only send out emails to PRs submitted by members of the organisation). Let me put that all in a new Issue.

cmd/emailreport/main.go
```diff
@@ -0,0 +1,112 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"github.com/BurntSushi/toml"
```
grahamwhaley: similar to the other file, I think we need to re-order this. I know `gofmt` probably sorted the list for you, but I think that is because `gofmt` will sort on a 'block' basis, and we had not broken this into blocks with a blank line. Hence, I think if you try:
```
import (
    "flag" ....
    <blank line>
    "github.com/BurntSushi/toml"
  .....
)
```

Then it should work...

cmd/emailreport/main.go
```diff
@@ -0,0 +1,112 @@
...
+		log.Fatal(err)
+	}
+
+	// Owner verification
+	currentUid = os.Geteuid()
+	fStat := fileinfo.Sys().(*syscall.Stat_t)
+	ownerUid = int(fStat.Uid)
+
+	if currentUid != ownerUid {
+		log.Fatal("You are not the owner of ", confFile)
```
grahamwhaley: Consider printing the two UUID values in the error as well. I'm surprised @jodh-intel didn't note that already ;-)

cmd/emailreport/main.go
MarioCarrilloA: ok, lets try it

cmd/emailreport/mail_test.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func TestHeaderRFC822(t *testing.T) {
+	var conf Configuration
+
+	if _, err := toml.DecodeFile("example.toml", &conf); err != nil {
+		log.Fatal(err)
+	}
+
+	HeaderRFC822(conf)
```
MarioCarrilloA: something like this: expectedHeader="XYZ";  if exepetedHeader != returnedHeader then error?

grahamwhaley: Yes, something like that I was thinking. I don't know if the `testing` package has anything specific to help you there, but I have realised you should probably be calling t.Fail() and not log.Fatal() in the tests, you think?:
https://golang.org/pkg/testing/#B.Fail

cmd/emailreport/README.md
```diff
@@ -0,0 +1,15 @@
+# PnP failure Report tool
+
+This tool reports by email the obtained PnP results from `checkmetrics`
+execution, next it will send the results by email.
+The configuration of SMTP server and checktool arguments are given
+by a configuration file in TOML format, this section includes an
+example of that.
```
MarioCarrilloA: the intention is to indicate that there is an example of configuration file for email tool in this directory however I am not sure it it worth to mentioned or it could be implicit?

grahamwhaley: OK, np - then rather than saying `this section includes`, we could say something like:

        an example TOML configuration file is provided in the file `example.toml`

MarioCarrilloA:
All comments have been fixed :)

grahamwhaley:
This is looking OK to me. Only comment is that the https://github.com/clearcontainers/runtime/blob/master/.ci/go-static-checks.sh script I am trying to land in #110 throws up 4 warnings in main.go - can you take a look at that and clean them up (otherwise I'll just have to do it before landing #110 ;-)

chavafg:
lgtm

grahamwhaley:
oh well, we missed the go-static-checks.sh cleanups, so I'll go do those before trying to land #110 ;-)

Issue #41, opened 2017-06-01
Make check problem with .ci/go-lint.sh


GabyCT:
Currently as we have it now if I do a $ make check and if in my machine it is already installed the gometalinter it will produce the following errors

.ci/go-lint.sh
vendor/github.com/onsi/ginkgo/reporters/stenographer/support/go-colorable/colorable_windows.go:353::warning: cyclomatic complexity 89 of function (*Writer).Write() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/scannerc.go:2304::warning: cyclomatic complexity 82 of function yaml_parser_scan_flow_scalar() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/scannerc.go:665::warning: cyclomatic complexity 70 of function yaml_parser_fetch_next_token() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/readerc.go:91::warning: cyclomatic complexity 68 of function yaml_parser_update_buffer() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/scannerc.go:2560::warning: cyclomatic complexity 53 of function yaml_parser_scan_plain_scalar() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/decode.go:331::warning: cyclomatic complexity 53 of function (*decoder).scalar() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/scannerc.go:2062::warning: cyclomatic complexity 52 of function yaml_parser_scan_block_scalar() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/emitterc.go:1428::warning: cyclomatic complexity 50 of function yaml_emitter_write_double_quoted_scalar() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/emitterc.go:983::warning: cyclomatic complexity 44 of function yaml_emitter_analyze_scalar() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/emitterc.go:311::warning: cyclomatic complexity 37 of function yaml_emitter_emit_document_start() is high (> 15) (gocyclo)
vendor/gopkg.in/yaml.v2/parserc.go:359::warning: cyclomatic complexity 33 of function yaml_parser_parse_node() is high (> 15) (gocyclo)
[...]

PR #110, opened 2017-07-04
ci: enable go static checks for repo

grahamwhaley:
Enable go static checks for this repo.

### DIFF
--- a/.ci/go-static-checks.sh
+++ b/.ci/go-static-checks.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+set -e
+
+# May not be necessary for CI machine runs, but enables local runs for some
+# users
+PATH=${PATH}:${GOPATH}/bin
+
+# Perform static go tests.
+
+function usage {
+	echo "Usage $0 [OPTIONS] [PACKAGES]"
+	echo "Perform static go checks on PACKAGES (./... by default)."
+	echo
+	echo "List of options:"
+	echo "  -h, --help             print this help"
+	echo "  -n, --no-network       do not access the network"
+}
+
+for i in "$@"; do
+	case $i in
+		-h|--help)
+			usage
+			exit 0
+			;;
+		-n|--no-network)
+			NONETWORK=1
+			shift
+			;;
+		*)
+			args="$args $i"
+			;;
+	esac
+done
+
+go_packages=$args
+
+# process everything in the tree, apart from any vendor bits, as they are
+# not our responsibility to check and fix
+[ -z "$go_packages" ] && {
+	go_packages=$(go list ./... | grep -v vendor)
+}
+
+function install_package {
+	url="$1"
+	name=${url##*/}
+
+	if [ -n "$NONETWORK" ]; then
+		echo "Skipping updating package $name, no network access allowed"
+		return
+	fi
+
+	echo Updating $name...
+	go get -u $url
+}
+
+# Ensure the checkers are installed and upto date
+install_package github.com/fzipp/gocyclo
+install_package github.com/client9/misspell/cmd/misspell
+install_package github.com/golang/lint/golint
+install_package github.com/gordonklaus/ineffassign
+
+echo Doing go static checks on packages: $go_packages
+
+echo "Running misspell..."
+go list -f '{{.Dir}}/*.go' $go_packages |\
+	xargs -I % bash -c "misspell -error %"
+
+echo "Running go vet..."
+go vet $go_packages
+
+cmd="gofmt -s -d -l"
+echo "Running gofmt..."
+diff=$(find . -not -wholename '*/vendor/*' -name '*.go' | \
+	xargs $cmd)
+if [ -n "$diff" -a $(echo "$diff" | wc -l) -ne 0 ]
+then
+	echo 2>&1 "ERROR: '$cmd' found problems:"
+	echo 2>&1 "$diff"
+	exit 1
+fi
+
+echo "Running cyclo..."
+gocyclo -over 15 `go list -f '{{.Dir}}/*.go' $go_packages`
+
+echo "Running golint..."
+for p in $go_packages; do golint -set_exit_status $p; done
+
+echo "Running ineffassign..."
+go list -f '{{.Dir}}' $go_packages | xargs -L 1 ineffassign
+
+echo "All Good!"
--- a/.ci/setup_tests.sh
+++ b/.ci/setup_tests.sh
@@ -42,3 +42,5 @@ checkcommits \
 	--body-length 72 \
 	--subject-length 75 \
 	--verbose
+
+.ci/go-static-checks.sh


grahamwhaley:
Marked as DNM until #103 has landed (as this will, or should, fail without that in place)
@jodh-intel @chavafg Do you think this is the right method to implement - feels 'too simple' somehow :-)

chavafg:
Time ago @dlespiau added a similar script[1] that uses a tool called `gometalinter`[2]  which executes the tools that you install in your script. 

I think that right now that script it is not called into the CI setup and currently seems that `misspell` is disabled.

[1] https://github.com/clearcontainers/tests/blob/master/.ci/go-lint.sh 
[2] https://github.com/alecthomas/gometalinter

.ci/go-static-checks.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+#  This file is part of cc-oci-runtime.
+#
+#  Copyright (C) 2016 Intel Corporation
```
jodh-intel: 2017?

.ci/go-static-checks.sh
```diff
@@ -0,0 +1,102 @@
...
+			args="$args $i"
+			;;
+	esac
+done
+
+go_packages=$args
+
+[ -z "$go_packages" ] && {
+	go_packages=$(go list ./... | grep -v vendor |\
+		    sed -e 's#.*/cc-oci-runtime/#./#')
```
jodh-intel: Shouldn't this (also?) exclude `cc-runtime`?

.ci/go-static-checks.sh
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function install_package {
+	url="$1"
+	name=${url##*/}
+
+	[ -n "$NONETWORK" ] && return
```
jodh-intel: I'd be tempted to add an `echo` so it's clear nothing is being updated here (I guess, you indirectly get that by *not* seeing "Updating ...", but... ;)

.ci/go-static-checks.sh
grahamwhaley: doh! copy-n-paste...

.ci/go-static-checks.sh
grahamwhaley: Ah, I didn't realise the script had repo specifics in it - will go read.

.ci/go-static-checks.sh
grahamwhaley: I'll have a look. If so, we might want to consider propogating this back across the other repos

grahamwhaley:
@chavafg ah, yes, I remember seeing `gometalinter` go past before. So, we have two scripts that do almost identical things - @dlespiau what direction are we heading here - do we go with `go-lint.sh` or `go-static-checks.sh`, or both in some form? We should decide and unify this across all the repos I suspect.
I'll wait for a direction check here before I go fixing other comments up.

.ci/go-static-checks.sh
jodh-intel: Right. This isn't a blocker, just a nice-to-have.

dlespiau:
fwiw, this repo had gometalinter setup (https://github.com/clearcontainers/tests/blob/master/.ci/go-lint.sh) but it was (unfortunately) disabled at some point, see https://github.com/clearcontainers/tests/issues/41.

I switched to gometalinter for the proxy so we can maintaining our own list of tools and the script around them.

Some checking is better than no checking, so any of those 2 is fine by me. Probably a good idea to remove the other one though.

grahamwhaley:
Ah, good pointer @dlespiau I'll see if I can resurrect that, and if we find that it is missing `misspell` then I'll mod the go-static... script as needed so we don't duplicate coverage.

.ci/go-static-checks.sh
grahamwhaley: Actually @jodh-intel I'm a little confused as to what this is trying to achieve and why - it seems to be trimming out the leading 'github.com..../cc-oci-runtime/blah' down to './blah', which seems to me to neither make sense for this repo or say the clearcontainers/runtime one. @chavafg do you have any background on this script and specifically this piece of the script?
In my local runs that `sed` part seems un-necessary, but maybe something is needed there for the running in the CI systems?
(I will probably drop the sed part and re-push to see what happens if I don't hear anything back soon...)

jodh-intel: @grahamwhaley - where did you get the code from? Seriously, it's getting difficult to track the origins of code spread across so many repos now ;)

grahamwhaley: iirc, I took a copy from the clearcontainers/runtime/.ci dir
https://github.com/clearcontainers/runtime/blob/master/.ci/go-static-checks.sh#L53
I think @chavafg submitted that one, hence the query above - @chavafg any history on this or insight into the sed'ism ?

Yep, I have considered if we could have a common .ci dir/repo across all the clearcontainers repos, but I'm guessing that just ain't going to fit in with any go vendor or github/sub-repo workflows

chavafg: Seems like that lines were submitted by @dlespiau

dlespiau: That was to work around oddities in how we setup the 2.x repo for golang in travis (because 2.x was a C/C++ project and Travis didn't have a GOPATH setup for us). We may well be able to remove that `sed`.

grahamwhaley: Right, thanks both @dlespiau @chavafg   I removed it locally already and it worked, so I'll resubmit the PR without the `sed` and we'll see how it flys. If that is all OK then I'll try to go mop up the other repos as well...

chavafg: that would be great, thanks @grahamwhaley

grahamwhaley:
Updated. Need to wait on #131, and then see if we need a rebase again (hopefully not)

chavafg:
seems like this needs rebase, so semaphore does not complain about the btrfs-progs issue.
@grahamwhaley, if we move on with this script, can you also please remove the one that calls gometalinter?

grahamwhaley:
Sure. This now needs #151 to land. I'll do the rebase after that.
I'll open a new issue for removing .ci/go-lint.sh and do that separately.

grahamwhaley:
Hi @devimc . I'm playing a constant game of catchup with this one as things get merged...
I'm getting warnings from some `localCI` items from gocyclo. Could you take a look and see if you can simplyfy for me?
We have our gocyclo limit set to complexity 15 - looks like you hit a 16 and a 19 :-)

```
19 main (*Repo).runTest /work/gwhaley/gopath/src/github.com/clearcontainers/tests/cmd/localCI/repo.go:373:1
16 main (*Repo).setup /work/gwhaley/gopath/src/github.com/clearcontainers/tests/cmd/localCI/repo.go:111:1
```

grahamwhaley:
and the chase continues....
so, we now have `gofmt` issues crept into `options_test.go` by the look of it as well.
I'm going to open (and assign) Issues for things currently showing up - and let's try and do a sweep and get this landed to prevent any further discrepancies creeping in...

```
ERROR: 'gofmt -s -d -l' found problems:
./functional/options_test.go
diff ./functional/options_test.go gofmt/./functional/options_test.go
--- /tmp/gofmt216042621	2017-07-21 12:20:01.830670191 +0100
+++ /tmp/gofmt335864760	2017-07-21 12:20:01.830670191 +0100
@@ -17,9 +17,9 @@
 import (
 	"fmt"
 
-	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/clearcontainers/tests"
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 )
 
@@ -28,7 +28,7 @@
 	return Entry(fmt.Sprintf("with option '%s'", option), command, fail)
 }
 
...
```

.ci/go-static-checks.sh
```diff
@@ -0,0 +1,109 @@
...
+	xargs $cmd)
+if [ -n "$diff" -a $(echo "$diff" | wc -l) -ne 0 ]
+then
+	echo 2>&1 "ERROR: '$cmd' found problems:"
+	echo 2>&1 "$diff"
+	exit 1
+fi
+
+echo "Running cyclo..."
+gocyclo -over 15 `go list -f '{{.Dir}}/*.go' $go_packages`
```
devimc: why ```over 15``` ?

sboeuf: That's what we have always used and it is a pretty good number if we don't want to end up with huge functions...

jodh-intel: We probably want to add a variable for this at the top of the file and put a comment on it explaining the rationale as you've done here.

grahamwhaley: :-)  Tis a very valid question. I didn't choose it - if anybody here ( @dlespiau  or @jodh-intel ) does have the actual background (even if it was 'guess and see'), let's hear it.
But, it got me curious. A quick surf turns up some vague info, but it would seem one of the primary reference papers (http://www.literateprogramming.com/mccabe.pdf) recommends '10' - so, maybe 15 is not so bad after all ;-)
It is a shame there is no #define//comment style 'override' for functions where we know what we are doing (option parsing is a classic). @jodh-intel hit one of these a couple of weeks ago and had to break it up. price we pay for having the check in place I'm afraid.

jodh-intel: It looks like `gocyclo` defaults to `-over 0`. I guess that's logical, but probably not what 99.9% of folk actually find usable ;)

.ci/go-static-checks.sh
```diff
@@ -0,0 +1,109 @@
...
+echo "Running misspell..."
+go list -f '{{.Dir}}/*.go' $go_packages |\
+	xargs -I % bash -c "misspell -error %"
+
+echo "Running go vet..."
+go vet $go_packages
+
+cmd="gofmt -s -d -l"
+echo "Running gofmt..."
+diff=$(find . -not -wholename '*/vendor/*' -name '*.go' | \
```
grahamwhaley: @GabyCT I had a quick look into why you were not seeing the gofmt warnings on #239 - I think it is this `find` line here probably causing the confusion/problems. The rest of the script seems to use the script args as $go_packages and work from that, but the find line just uses `.`, and hence probably does not necessarily find the code when packages are passed on the cmdline.
I'm thinking we should change this to be more like the `go list -f` as used a few lines above for the misspell search. At least making things uniform across the script would be good :-)
@chavafg what do you think?
I'm happy to rework this script before we land this PR, and then propagate it or relevant changes across the other repos.

icarus-sparry: It would be slightly more efficient to not walk the vendor tree, rather than walk it and filter out the names which contain vendor.

    find . \( -name vendor -prune \) -o -name '*.go'

Translation: If the name matches 'vendor' then prune off everything below it and move onto the next thing, or if the name matches '*.go' then print the path

jodh-intel:
Travis is failing due to a typo in the tests ;)

grahamwhaley:
heh heh - is that a mispell catch?
Yeah, have not gotten to clean up the static checking across all the repos yet. And tbh, I don't see me getting to that for at least the next couple of weeks either.

grahamwhaley:
I think we are happy with the 'status quo' for clearcontainers wrt checks, and kata has taken on board the improved idea of a single central script 'to rule them all' -  https://github.com/kata-containers/tests/pull/7 (thanks @jodh-intel ), so I'm closing this one down!

PR #140, opened 2017-07-10
ci: run apt-get using sudo -E

### LINKED ISSUE #139: failed to run apt-get behind a proxy
```apt-get``` commands should run with ```sudo -E``` to inherit the environment variables

devimc:
running the CI behind a proxy ```apt-get``` fails trying to
connect to ubuntu servers since it's unable to resolve
the proxy. To fix this issue the CI should inherit the environment
variables of the calling process hence we should run ```apt-get```
using ```sudo -E```
    
fixes #139


grahamwhaley:
lgtm
Could have mentioned in the commit that this was found specifically with a proxy setup, like you mentioned in the Issue I think.

devimc:
@grahamwhaley  fixed
thanks

PR #142, opened 2017-07-10
.ci: Fix CRI-O build by installing manually libdevmapper and btrfs

### LINKED ISSUE #143: CRI-O build is broken
CRI-O build doesn't work anymore because btrfs and devicemapper packages provided by Semaphore CI environment (Ubuntu 14.04) are too old.
We should build from sources instead.

sboeuf:
This patch adds some commands to build and install libdevmapper and
btrfs-tools from sources, so that we ensure we get the most recent
version. Indeed, CRI-O needs some pretty recent versions of those
libraries/tools.

Fixes #143

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -16,9 +16,6 @@
 
 set -e
 
-echo "Install dependencies: libglib2.0-dev libseccomp-dev libapparmor-dev libgpgme11-dev libdevmapper-dev btrfs-tools"
-sudo -E apt-get install -y libglib2.0-dev libseccomp-dev libapparmor-dev libgpgme11-dev libdevmapper-dev btrfs-tools
-
 echo "Get CRI-O sources"
 go get -d github.com/kubernetes-incubator/cri-o || true
 pushd $GOPATH/src/github.com/kubernetes-incubator/cri-o
--- a/.ci/setup_env_ubuntu.sh
+++ b/.ci/setup_env_ubuntu.sh
@@ -34,6 +34,28 @@ chronic sudo -E apt-get update
 echo "Install qemu-lite binary"
 chronic sudo -E apt-get install -y --force-yes qemu-lite
 
+echo "Install CRI-O dependencies"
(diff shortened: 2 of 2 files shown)


.ci/install_crio.sh
```diff
@@ -16,8 +16,27 @@
 
 set -e
 
-echo "Install dependencies: libglib2.0-dev libseccomp-dev libapparmor-dev libgpgme11-dev libdevmapper-dev btrfs-tools"
-sudo -E apt-get install -y libglib2.0-dev libseccomp-dev libapparmor-dev libgpgme11-dev libdevmapper-dev btrfs-tools
+echo "Install dependencies: libglib2.0-dev libseccomp-dev libapparmor-dev libgpgme11-dev"
+sudo -E apt-get install -y libglib2.0-dev libseccomp-dev libapparmor-dev libgpgme11-dev
```
devimc: centOS does not have ```apt-get```

.ci/install_crio.sh
```diff
@@ -16,8 +16,27 @@
```
devimc: this script must be compatible with all other distros, if not then it should be part of ```setup_env_ubuntu.sh```

sboeuf: Oh yes you're right

chavafg:
lgtm

.ci/setup_env_ubuntu.sh
```diff
@@ -34,6 +34,28 @@ chronic sudo -E apt-get update
 echo "Install qemu-lite binary"
 chronic sudo -E apt-get install -y --force-yes qemu-lite
 
+echo "Install CRI-O dependencies"
+sudo -E apt-get install -y libglib2.0-dev libseccomp-dev libapparmor-dev libgpgme11-dev
+
+echo "Install libdevmapper"
+git clone git://sourceware.org/git/lvm2.git
```
devimc: could you change this to http ?

sboeuf: done

devimc:
lgtm

PR #161, opened 2017-07-13
run_all_metrics: detect script path

bwarden:
Add SCRIPT_PATH to run_all_metrics.sh script so it can execute
tests successfully from any location.

### DIFF
--- a/metrics/run_all_metrics.sh
+++ b/metrics/run_all_metrics.sh
@@ -16,7 +16,9 @@
 
 set -e
 
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+
 echo "Running all metrics tests"
 
 # Run the time tests
-bash metrics/time/docker_workload_time.sh true busybox $RUNTIME 100
+bash ${SCRIPT_PATH}/time/docker_workload_time.sh true busybox $RUNTIME 100


chavafg:
@bwarden, can you please create an issue for the problem you are solving with this PR and add to the commit message a line with `Fixes #XYZ.` refering to the issue number that you created, so that travis-ci does not complain?

bwarden:
Opened #168 for this. Do I need to amend the actual commit, the original comment in this thread, or just add it in this comment like so?

Fixes #168

bwarden:
Alright, Travis doesn't like amended commits. I'll just open a new PR.

grahamwhaley:
Hi @bwarden . You need to modify the commit itself, as that is what `checkcommits` checks on Travis. It cannot read the comments or merge message I'm afraid. I PR'd to update the CONTRIB file to make this clearer yesterday.
force pushing updates to the PR is our normal workflow (rather than adding extra patches to the PR or opening new PRs to replace existing ones). But, we have been seeing a couple of Travis issues recently:
 - sometimes if you force push 'too fast' Travis picks up the old commit and then we fail, and the only way we have found to get out of that is the close/re-open-new-PR path
 - Travis and checkcommits are currently having frequent debates with each other about which commit/branch needs checking. This is probably what you ran into. Force pushes clear that more times than not. We are working on figuring that one out...

Issue #149, opened 2017-07-12
metric: migrate CC2.x metrics tests to this repository


grahamwhaley:
The metrics tests that currently live in the CC2.x runtime repo at:
https://github.com/01org/cc-oci-runtime/tree/master/tests/metrics
need migrating over to this repository so we can utilise and share with CC3.x as well.
For an example of where and how to adapt and integrate them, have a look at:
https://github.com/clearcontainers/tests/pull/86
And then to ensure the metrics outputs are fitting in with the planned scheme, also ensure the pending https://github.com/clearcontainers/tests/pull/98 interprets the results correctly.
I am expecting `checkmetrics` will need some work/updates - for instance, I believe it will only handle float type data in the Results field right now, and we will need to modify it to handle at least int type as well.

PR #162, opened 2017-07-13
Merge metrics from 2.x

bwarden:
Merging existing network metrics tests from the 2.x repository, preserving history. Moving from "tests/metrics/network" to "metrics/network", since this repository is already named "tests". No other porting changes included.


chavafg:
Not sure if this is going to be a 2-step process, but some of the files here (those with `.in`) were processed by `autotools` and we do not have that here, so they will not work. 

If that is going to be fixed in another PR, can you please create an issue to track it?
Thanks.

bwarden:
This will be multiple steps. First, I'm bringing the existing files over, then I'll make each one work.
- Rework generated files: #170

bwarden:
This is the first step toward resolving #149

grahamwhaley:
tbh, I would have preferred that we fixed the files as we imported them, otherwise we end up placing a bunch of 'known broken' files in our tree, and generally that is not our workflow.
I do like you keeping the history of the patches there though @bwarden 
Would you be amenable to adding your patch(es) on top of the history before we do the PR merge @bwarden , so we don't end up with a bunch of broken code in the tree (albeit temporarily).

bwarden:
Alright, so you'd like me to push patches that implement patching the tests in this same PR? Should work. I'll keep each patch to a logical set of changes, of course.

grahamwhaley:
Thanks. I think that is going to allow us to land a better first set of code (rather than something known to not work).
It it works better for you then you could do the tests individually or in smaller batches. See how 'big and complex' it feels. Smaller bits can bring some benefits:
 - easier (and thus more likely to get) review
 - less inter-dependence - if one part blocks it does not block the other PRs.

metrics/network/README.md
```diff
@@ -0,0 +1,87 @@
+# Networking tests for the Intel® Clear Containers runtime
```
iphutch: "the" is not necessary here. 
Should read: Networking tests for Intel® Clear Containers runtime

metrics/network/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Currently, the Clear Containers have a series of network performance tests.
```
iphutch: remove/the

metrics/network/README.md
bwarden: Will remove

metrics/network/README.md
bwarden: Will remove

bwarden:
Travis doesn't like this commit (Commit 5ca8d209d517fc75531cf5a34d82357bef75c933: Failed to find subsystem in subject: "Dockerfile for network image"):
`
commit 5ca8d209d517fc75531cf5a34d82357bef75c933
Author: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>
Date:   Tue May 9 19:13:18 2017 +0000

    Dockerfile for network image
    
    This Dockerfile contains all the tools and dependencies
    requiered to run the network tests.
    
`

bwarden:
Moved qemu-finding function change to separate PR #182

grahamwhaley:
Hi @bwarden Thanks for fixing the things from the feedback.
I'm guessing you have more patches to come, such as adding the `save_results` call to all the tests etc.
If so, we should mark this PR as work in progress (WIP) or do not merge, probably best by adding the relevant github label to this PR. That way we don't run the risk of somebody accidentally merging or giving a full 'ack' until the series is complete.
On that note - you can push patches up to this PR whilst still working on the port, but note, each time you push it notifies us and we will come do a review (at least read the new code), so you may want to batch any changesets up to nice logical bits to review - both for uniformity and to not burn out the reviewers :-)

bwarden:
Once the next force push succeeds, review solicited. Commit 52fe8167497d21d3a3bdfb0882ac5bfcb6560a98 focuses on putting the necessary common functions in place for the network metrics to actually run. I have not verified that the tests are doing the right thing yet.

metrics/lib/common.bash
```diff
@@ -18,12 +18,17 @@ SCRIPT_PATH=$(dirname $(readlink -f $0))
 RESULT_DIR="${SCRIPT_PATH}/../results"
 LIB_DIR="${SCRIPT_PATH}/../lib"
 
+# Set variables to reasonable defaults if unset or empty
+DOCKER_EXE="${DOCKER_EXE:-docker}"
+RUNTIME="${RUNTIME:-cor}"
```
grahamwhaley: maybe set the default to the CC3.x runtime name rather than the CC2.x runtime name.

bwarden: Changed to "cc-runtime" and pushed.

metrics/lib/common.bash
```diff
@@ -84,74 +89,105 @@ function init_env()
...
+	else
+		return 0
+	fi
+}
+
+# This function checks if there are containers or
+# cc-shim and qemu-lite processes up, if found, they are
+# killed to start test with clean environment.
+# The function is intended to be used in the setup() of
+# the docker integration tests for cc-oci-runtime.
```
jodh-intel: Mention of `cc-oci-runtime`. Does this comment need updating?

metrics/network/dockerfile_image/Dockerfile
```diff
@@ -0,0 +1,40 @@
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/lib/network-test-common.bash
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-latency.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-metrics-cpu-consumption.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-metrics-iperf3.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-metrics-memory-pss-1g.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-metrics-memory-pss.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-metrics-memory-rss-1g.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-metrics-nuttcp.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-metrics.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

metrics/network/wip/network-nginx-ab-benchmark.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+#  This file is part of cc-oci-runtime.
```
jodh-intel: Comment is now incorrect.

grahamwhaley:
As an update from an offline discussion - to ease the process of updating and landing all the tests we are going to initially land much of this code in a form that won't function properly (hence, items now placed under the `wip` folder), and then do individual updates to enable and move tests into their working place and state one at a time. Right now that is going to be the most efficient workflow to get all the tests landed and working.
Thus, @bwarden please review @jodh-intel comments for any parts not in the wip folder. For the parts in wip, thanks for the feedback @jodh-intel We may initially land with some of that still in place, and pick up the feedback as we do the subsequent PRs to update.
So, sure, in a perfect world I'd rather not land things in tree that don't actually function, but here we are trying to be prudently efficient.
I guess an alternative flow would (have been) a topic branch. I'm happy to land this PR (post a few more edits) though and push on.

metrics/network/dockerfile_image/Dockerfile
grahamwhaley: @jodh-intel you could have just said 'and in a number of other files, please grep and fix' :-)

bwarden:
I've removed all comments referring to cc-oci-runtime. I've also adjusted the default CC_SHIM_PATH value to /usr/libexec/clear-containers/cc-shim. This is only used in a function that looks for running processes to kill to ensure the test starts with a clean slate (kill_processes_before_start in metrics/lib/common.bash). Do I need to add cc-proxy to that function as well?

grahamwhaley:
Hi @bwarden for the process reset/clean/kill - the proxy is/would have to be treated a little differently as it is something owned/managed by systemd. There is already some code in that file that uses `systemctl` do do something similar, and we should probably examine that code and amend to make sure it also does the correct thing for the `cc-runtime` runtime:
https://github.com/clearcontainers/tests/blob/master/metrics/lib/common.bash#L81

grahamwhaley:
Thanks @bwarden I'm going to merge this. If I were being picky then I'd probably ask for some patch shuffle and merging, like we could do all the file name/moves in a single patch etc. - but, let's get this in and then work on fixing the tests up into the final positions.

PR #179, opened 2017-07-14
run_all_metrics: Check value of RUNTIME

### LINKED ISSUE #172: metrics: script should quit cleanly if no $RUNTIME
The run_all_metrics script relies on $RUNTIME being set in the env to run cleanly - if it is not set you get unhelpful errors from the subsequent badly formed command invocations: 
https://github.com/clearcontainers/tests/blob/master/metrics/run_all_metrics.sh#L22

We should add a check to the script to abort cleanly if any of its pre-reqs are not met.

bwarden:
Environment variable RUNTIME must be defined before running this test.
Verify that it's set to an integer that's greater than zero.

Fixes #172

### DIFF
--- a/metrics/run_all_metrics.sh
+++ b/metrics/run_all_metrics.sh
@@ -17,6 +17,14 @@
 set -e
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/lib/common.bash"
+
+# Check arguments
+if [[ ! -v RUNTIME ]]; then
+    die Variable RUNTIME must be set to the name of your CC runtime
+fi
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 
 echo "Running all metrics tests"
 


metrics/run_all_metrics.sh
```diff
@@ -16,6 +16,20 @@
 
 set -e
 
+# Check arguments
+case ${RUNTIME#[-+]} in
+    ''|*[!0-9]* )
+        echo Variable RUNTIME must be set to a valid integer
```
jodh-intel: For consistency, I'd recommend sourcing `lib/common.bash` then just calling `die "Variable RUNTIME must be set to a valid integer`".

grahamwhaley:
Also, `RUNTIME` here is the name of the runtime as configured in docker (such as `runc`, `cor` or `cc-runtime`), and not an integer - or at least, that is the intent, and is how the top level `Makefile` sets it here:
https://github.com/clearcontainers/tests/blob/master/Makefile#L37
from
https://github.com/clearcontainers/tests/blob/master/Makefile#L16

So, rather than checking for an integer, we just need to check the variable is set to something. We don't need to check what, as that is down to the user and system config - as long as it is not empty (not set), then we should proceed to try and run the tests.

bwarden:
Ah, OK, I parsed this: "runtime=${RUNTIME} units=seconds" as time to run. Will fix, and source common to use die.

metrics/run_all_metrics.sh
```diff
@@ -17,6 +17,14 @@
 set -e
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/lib/common.bash"
+
+# Check arguments
+if [[ ! -v RUNTIME ]]; then
+    die Variable RUNTIME must be set to the path of your CC runtime binary
```
grahamwhaley: Sorry to be a bit of a pedant :-) - it is not 'to the path' it is 'to the name'. Specifically, it is the name of the runtime as configured into Docker. For instance, for CC2.x the name is `cor`, but the actual binary is `cc-oci-runtime`. We use the name here as that is what we pass to the docker run commands with `--runtime=xxx`.

grahamwhaley:
@chavafg @devimc we have a qa-failed here, but semaphore and Travis are happy - can you get us some details on that? thx

devimc:
@grahamwhaley was a problem with the proxy in the systemd unit file

metrics/run_all_metrics.sh
bwarden: Understood; will fix.

grahamwhaley:
hi @bwarden ah, did you post an update - I might have missed that. Best post a message in the comments telling us there is something new to review when you push it. I'll go have a look now...

metrics/run_all_metrics.sh
```diff
@@ -17,6 +17,14 @@
 set -e
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/lib/common.bash"
+
+# Check arguments
+if [[ ! -v RUNTIME ]]; then
+    die Variable RUNTIME must be set to the name of your CC runtime
```
grahamwhaley: nitpick: it doesn't have to be a CC runtime ;-) But, seeing as this is a CC repo, it's not really a problem, and won't stop merge :-)

grahamwhaley:
lgtm

grahamwhaley:
@jodh-intel as you had a review request for changes, OK with you as well?

jodh-intel:
lgtm

PR #180, opened 2017-07-14
functional: Implement standalone testing of the exec command

### LINKED ISSUE #171: Add runtime standalone testing for exec command
We don't have any tests for the runtime `exec` command. Particularly, we want to test the ability to detach/attach.

sboeuf:
We are submitting patches to the runtime repository in order to
fix the behavior of the exec command, but we were missing some
tests to validate those expected behaviors.

This patch adds some testing of the exec command when it is
either started "attached" or "detached" from the shell.

Fixes #171

### DIFF
--- a/functional/exec_test.go
+++ b/functional/exec_test.go
@@ -13,3 +13,104 @@
 // limitations under the License.
 
 package functional
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/clearcontainers/tests"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+var sleepingContainerWorkload = []string{"sh", "-c", "sleep 1000"}
+
+func execDetachTiming(detach bool) TableEntry {
+	comparator := ">="
+	if detach {
+		comparator = "<"
+	}
+
+	timeout := 1
+
+	process := Process{
+		Workload: []string{"sleep", fmt.Sprintf("%d", timeout)},
+		Detach:   detach,
+	}
+
+	return Entry(fmt.Sprintf("check time as detach=%t", detach), process, comparator, timeout)
+}
+
+func execDetachOutput(detach bool) TableEntry {
+	output := "HelloWorld"
+
+	process := Process{
+		Workload: []string{"echo", output},
+		Detach:   detach,
+	}
+
+	expectedOutput := output
+	if detach {
+		expectedOutput = ""
+	}
+
+	return Entry(fmt.Sprintf("check output as detach=%t", detach), process, expectedOutput)
+}
+
+var _ = Describe("exec", func() {
+	var (
+		container *Container
+		exitCode  int
+		err       error
+	)
+
+	BeforeEach(func() {
+		container, err = NewContainer(sleepingContainerWorkload, true)
+		Expect(container).ShouldNot(BeNil())
+		Expect(err).ShouldNot(HaveOccurred())
+		_, _, exitCode := container.Run()
+		Expect(exitCode).Should(Equal(0))
+	})
+
+	AfterEach(func() {
+		Expect(container.Exist()).Should(BeTrue())
+		_, _, exitCode = container.Delete(true)
+		Expect(exitCode).Should(Equal(0))
+		Expect(container.Cleanup()).Should(Succeed())
+	})
+
+	DescribeTable("container",
+		func(process Process, comparator string, timeout int) {
+			process.ContainerID = container.ID
+			tInit := time.Now()
+
+			_, _, exitCode = container.Exec(process)
+			Expect(exitCode).Should(Equal(0))
+
+			duration := time.Since(tInit)
+			Expect(duration.Seconds()).Should(BeNumerically(comparator, timeout))
+		},
+		execDetachTiming(false),
+		execDetachTiming(true),
+	)
+
+	DescribeTable("container",
+		func(process Process, expectedOutput string) {
+			process.ContainerID = container.ID
+
+			stdout, stderr, exitCode := container.Exec(process)
+			Expect(exitCode).Should(Equal(0))
+
+			if expectedOutput != "" {
+				Expect(stdout.String()).Should(ContainSubstring(expectedOutput))
+			} else {
+				Expect(stdout.String()).Should(BeEmpty())
+			}
+
+			Expect(stderr.String()).Should(BeEmpty())
+		},
+		execDetachOutput(false),
+		execDetachOutput(true),
+	)
+})
--- a/container.go
+++ b/container.go
@@ -46,19 +46,32 @@ type Container struct {
 	// if nil then try to run the container without --log option
 	LogFile *string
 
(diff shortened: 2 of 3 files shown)


devimc:
lgtm

functional/exec_test.go
```diff
@@ -13,3 +13,104 @@
...
+var _ = Describe("exec", func() {
+	var (
+		container *Container
+		exitCode  int
+		err       error
+	)
+
+	BeforeEach(func() {
+		container, err = NewContainer(sleepingContainerWorkload, true)
+		Ω(container).ShouldNot(BeNil())
```
grahamwhaley: `Ω` - really? OMG! so, the docs say you can also use the functionally equal `Expect` name here from gomega. That would be much preferable to me personally - I don't think I've got any chance of grep'ing `Ω` etc. etc. Thoughts?

functional/exec_test.go
```diff
@@ -13,3 +13,104 @@
...
+	})
+
+	AfterEach(func() {
+		Ω(container.Exist()).Should(BeTrue())
+		_, _, exitCode = container.Delete(true)
+		Ω(exitCode).Should(Equal(0))
+		Ω(container.Cleanup()).Should(Succeed())
+	})
+
+	DescribeTable("container",
```
grahamwhaley: It would be good to have a comment for each test block/type/grouping - such as here, it takes some effort to read the code here and the related code it is calling to try and work out it is testing that a `sleep` in a detach takes as long as is expected.
Side note - I think you are doing the sleep test for 1s, and then checking the duration below. `sleep` is not really a very accurate tool - do not be surprised if we get spurious failures on this test later with the time being <1s or >1s due to granularity and other issues. Just a heads up.

grahamwhaley:
with feedback,
lgtm

functional/exec_test.go
```diff
@@ -13,3 +13,104 @@
...
+func execDetachTiming(detach bool) TableEntry {
+	comparator := ">="
+	if detach {
+		comparator = "<"
+	}
+
+	timeout := 1
+
+	process := Process{
+		Workload: []string{"/bin/sleep", fmt.Sprintf("%d", timeout)},
```
jodh-intel: For consistency with the function below, I'd just specify `sleep`.

functional/exec_test.go
```diff
@@ -13,3 +13,104 @@
...
+		Detach:   detach,
+	}
+
+	return Entry(fmt.Sprintf("check time as detach=%t", detach), process, comparator, timeout)
+}
+
+func execDetachOutput(detach bool) TableEntry {
+	outputHelloWorld := "HelloWorld"
```
jodh-intel: That's a rather unwieldy (but admittedly very precise) variable name. How about just `output`?

functional/exec_test.go
jodh-intel: +1. It looks cool, but it's a bit fanciful and not super-clear.

/me wonders if there is a IOGCC... :smile:

functional/exec_test.go
sboeuf: I think I need to add some environment variables, otherwise it does not work. I'll check that.

functional/exec_test.go
sboeuf: I'll change that.

sboeuf:
@jodh-intel I agree the ginkgo framework is a bit obscure ! I mean it's kind of a new language but we started implementing standalone functional test with this framework, that's why I have submitted this PR. Maybe we can move to something else if we consider this is not clear enough.

functional/exec_test.go
grahamwhaley: OK, I'm going to request that we don't use `Ω`, but we use `Expect`, and to provide some semi-random other use case, it looks like the kubernetes end to end tests use ginkgo, and use `Expect`:
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/common/init_container.go#L86

@sboeuf I suspect you just used the idiom that was already in the file - that's fine, but can you edit your commit to be Expect. I've opened #192 so we can track conversion of the rest of the `Ω` calls conversion.

grahamwhaley:
Oh, and I meant to point this out yesterday - I think this PR has landed before the actual functionality in the rest of the stack up - thus (and correctly, which is a bit of nice confirmation), the tests have failed on semaphore - so we can't/won't actually merge this PR until the rest of the stackup has caught up and the tests start passing.
Therefore you might want to mark this as WIP or DNM for the minute

functional/exec_test.go
devimc: @grahamwhaley  according with the documentation http://onsi.github.io/gomega/#making-assertions
```Ω```  is ok, because He is using ```ShouldNot```

grahamwhaley: Technically - fine. Semantically horrid. **NACK**

devimc: nahh that's fine, for me is the same (```Expect``` == ```Ω```).

😈 I guess we have to edit @jodh-intel 's comment because He used a non-ascii character  😝

jodh-intel: /me throws :doughnut:'s at @devimc! :smile:

functional/exec_test.go
sboeuf: done

functional/exec_test.go
sboeuf: done

functional/exec_test.go
sboeuf: This works too. I have made the change.

functional/exec_test.go
sboeuf: @grahamwhaley I don't think those tests will fail sporadically. Sleep is not accurate but I am sure it will last at least one second, and because my test is about making it is 1s or greater in that case, I should be fine.
Regarding the other case, the sleep is not applied and I test that it doesn't take a second to detach from the command.

sboeuf: @grahamwhaley each test has a comment as you can see the functions `execDetachTiming()` and `execDetachOutput()` are providing some strings to the `Entry()` function call.

sboeuf:
@jodh-intel @grahamwhaley I have updated this PR, please take a look again. This PR will fail the semaphore functional testing until https://github.com/clearcontainers/runtime/pull/327 gets merged. Please retrigger semaphore testing when https://github.com/clearcontainers/runtime/pull/327 is merged.

jodh-intel:
clearcontainers/runtime#327 is now merged and semaphore re-triggered.

lgtm

sboeuf:
Needs https://github.com/clearcontainers/tests/pull/230 to be merged to fix Semaphore CI.
And it also depends on https://github.com/clearcontainers/runtime/pull/356 to be merged too.

sboeuf:
@grahamwhaley could you please review again and check that I have done the changes you wanted.

grahamwhaley:
Thanks for the changes @sboeuf Given we have #192 to get the rest of the file(s) updated:
lgtm

sboeuf:
@grahamwhaley Please merge it.

grahamwhaley:
oh, ok, just because it is for you @sboeuf ;-)

Issue #170, opened 2017-07-13
Network metrics scripts currently require autotools


bwarden:
Need to rework the scripts to be static files with appropriately passed environment variables:
- metrics/network/network-metrics-cpu-consumption.sh.in
- metrics/network/network-metrics-memory-pss-1g.sh.in
- metrics/network/network-metrics-memory-pss.sh.in
- metrics/network/network-metrics-memory-rss-1g.sh.in

Looks like they only need QEMU_PATH.

dlespiau:
Of course, we don't want autotools here, just a makefile rule to generate the file files from the .in files.

bwarden:
I plan to rework the individual scripts to use an environment variable instead, and if necessary, generate a single script that they include, in which the variable is actually defined. If QEMU_PATH is already defined by the test execution process, then I don't even need to generate files.

grahamwhaley:
Personally I prefer pre-setting an ENV var rather than pre-processing the scripts - but it is a bit swings/roundabouts.

We already have a common file for common things:
https://github.com/clearcontainers/tests/blob/master/metrics/lib/common.bash
which may be a convenient place to set the VARS up, and it is already pulled into the one existing metrics test:
https://github.com/clearcontainers/tests/blob/master/metrics/time/docker_workload_time.sh#L27

Just to note, finding the appropriate QEMU is actually quite difficult to get 100% correct. Technically you'd have to, I believe, scour the docker setup to find the correct runtime files, and then scour the runtime config files to locate the actual QEMU binary configured to be used.
That's all a bit tricky, and for now I think we will probably have to do our 'best guess'. @chavafg I believe is the expert here on what the pitfalls are and what is pragmatic to do at present.

PR #182, opened 2017-07-14
common: implement helper function to find qemu

### LINKED ISSUE #190: Implement helper function to find qemu binary for tests
First step of fixing #170.

bwarden:
Adds a helper function to the metrics common library to extract the qemu
path from "cc-runtime cc-env" or configuration files at runtime. This
allows several of the network metrics tests to be converted to static
scripts instead of generated via autotools.

Fixes #190

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -42,3 +42,75 @@ function save_results(){
 	bash $LIB_DIR/send_results.sh -n "$1" -a "$2" -r "$3" -u "$4"
 }
 
+function get_hypervisor_from_toml(){
+    ## Regular expressions used for TOML parsing
+    # Matches a section header
+    section_re="^\s*\[(\S+)]"
+    # Matches the name of the hypervisor section
+    hypervisor_re="^hypervisor(\..*)?"
+    # Matches the variable containing the qemu path
+    qemu_re="^\s*path\s*=\s*\"([^\"]+)"
+
+    # Case insensitive
+    shopt -s nocasematch
+
+    for line in "$@"; do
+        if [[ $line =~ $section_re ]]; then
+            # New section
+            section=${BASH_REMATCH[1]}
+        elif [[ $section =~ $hypervisor_re ]]; then
+            # Look for qemu path
+            if [[ $line =~ $qemu_re ]]; then
+                # Found it
+                qemu_path="${BASH_REMATCH[1]}"
+                echo "${qemu_path}"
+                break;
+            fi
+        fi
+    done
+}
+
+# Find a reasonable path to the hypervisor on this system
+function get_qemu_path(){
+   # Create a list of potential configuration files
+    declare -a conf_files
+
+    # Use cc-env, if available
+    CC_RUNTIME=$(command -v cc-runtime)
+    if [[ $? -eq 0 ]] && [[ -n ${CC_RUNTIME} ]]; then
+        cc_env_tmp=$(mktemp cc-env.XXXX)
+        ${CC_RUNTIME} cc-env > "${cc_env_tmp}"
+        conf_files+=("${cc_env_tmp}")
+    fi
+
+    # Search for other configuration files
+    conf_files+=("${LIB_DIR}/../../../runtime/config/configuration.toml")
+    conf_files+=($(find /etc -type f -name configuration.toml -exec \
+        grep -l 'hypervisor.qemu' {} + 2>/dev/null))
+
+    # Check the potential files sequentially
+    for conf_file in "${conf_files[@]}"; do
+        [[ -f "${conf_file}" ]] || continue
+
+        # Attempt to parse the found config file (TOML)
+        declare -a config
+        while read line; do
+            config+=("$line")
+        done < "${conf_file}"
+        qemu_path=$(get_hypervisor_from_toml "${config[@]}")
+
+        # Got one?
+        [[ -n "${qemu_path}" ]] || continue;
+        [[ -x "${qemu_path}" ]] && break;
+    done
+
+    # Cleanup
+    [[ -n "${cc_env_tmp}" ]] && rm "${cc_env_tmp}"
+
+    # Check whether we got a good result
+    [[ -n "$qemu_path" ]] || die Failed to find qemu path in $conf_file
+    [[ -f "$qemu_path" ]] || die "$qemu_path does not exist"
+    [[ -x "$qemu_path" ]] || die "$qemu_path is not executable"
+
+    echo "$qemu_path"
+}


bwarden:
This is the first step in addressing #170.

jodh-intel:
I'm rather confused by this. If the code is only for CC 3.0 you can simply parse the output of:

```
$ cc-runtime cc-env | grep -A 2 Hypervisor
[Hypervisor]
  Path = "/usr/bin/qemu-lite-system-x86_64"
  Resolved = "/usr/bin/qemu-lite-system-x86_64"
```

Also note that you need to raise an issue and add a "Fixes #XXX" comment in one of the commits to keep Travis and `checkcommits` happy.

grahamwhaley:
@jodh-intel I'd like this to work for both CC2.x and CC3.x. We can definitely prioritise CC3.x as the primary target here though.
One remaining problem though (and this is part of the reason I currently hard coded some of the items here https://github.com/clearcontainers/tests/blob/master/metrics/lib/send_results.sh#L42 and noted so in my initial PR) is how do you work out what is actually configured into `docker`. If we are sort of assuming everything is set up using the default names and paths then we could just use though verbatim :-)

I think for running the tests in the CI in the tree/default installation then we can just hard wire things for now.
Ideally I'd like to be able to run the tests on any installed system, outside of a source tree.

So, let's be pragmatic. Step one is we get this up and running with a CC3.x source tree installation under a CI test system. And then we will adapt it to run CC2.x and binary only installations etc.

What would be nice I guess is if we make some hard assumptions about where things are, then we should fail gracefully and obviously if they are not there - so at least we don't end up misleading folks with 'fake results' if the code is run outside of a supported setup.

bwarden:
I can try cc-env first and fall back to checking config files if that makes sense, or only use cc-env for now and worry about CC2.x later, whichever makes more sense.

jodh-intel:
The ideal would be to have a single code path so if CC 2.x needs to be handled, it may be best to avoid calling `cc-runtime cc-env` - @grahamwhaley - what's your view?

metrics/lib/common.bash
```diff
@@ -42,3 +42,54 @@ function save_results(){
...
+    qemu_re="^\s*path\s*=\s*\"([^\"]+)"
+
+    # Create a list of potential configuration files
+    declare -a conf_files
+    conf_files+=("${LIB_DIR}/../../../runtime/config/configuration.toml")
+    conf_files+=($(find /etc -type f -name configuration.toml -exec \
+        grep -l 'hypervisor.qemu' {} + 2>/dev/null))
+
+    # Check the potential files sequentially
+    for ((i=0; i<${#conf_files[@]}; i++)); do
```
jodh-intel: I think this would be a lot clearer if you don't use an index variable, jump out of the loops asap (to avoid the deep indentation) and just use single square brackets. Something like:

```
for conf_file in ${conf_files[*]}; do
    [ -f "${conf_file}" ] || continue

    # ...

done
```

metrics/lib/common.bash
```diff
@@ -42,3 +42,54 @@ function save_results(){
...
+                            break 2 # Get out of both loops
+                        fi
+                    fi
+                fi
+            done < $conf_file
+        fi
+    done
+
+    # Check whether we got a good result
+    if [[ "" == "$qemu_path" ]]; then
```
jodh-intel:
```
if [ -z "$qemu_path" ]; then
    # ...
fi
```

Or even better:

```
[ -n "$qemu_path" ] || die "Failed to find qemu path in $conf_file"
[ -x "$qemu_path" ] || die "$qemu_path is not executable"

echo "qemu_path"
```

metrics/lib/common.bash
```diff
@@ -42,3 +42,54 @@ function save_results(){
...
+        die Failed to find qemu path in $conf_file
+    elif [[ ! -x "$qemu_path" ]]; then
```
jodh-intel: You could add another check:

```
machine_type=pc-lite
$qemu_path -machine help | grep -q "^${machine_type}"
[ $? -eq 0 ] || die "$qemu_path" doesn't support machine type $machine_type"
```

grahamwhaley: apart from we may have different machine types between different runtimes - and ideally this code would 'just work' with other modified or experimental VMs. I'm happy to skip this check here, and let that fail in the runtime itself - and/or, if it does not fail in the runtime then the runtime is doing something we don't understand ;-)
It is a little swings-and-roundabouts - I wish we didn't have to rely on such internal knowledge of how the runtime works in the first place, but to do some of the metrics measures (footprint of all the components, cpu consumption of similar) we do.

grahamwhaley:
@jodh-intel @bwarden I don't mind having two (or multiple) code paths if n-1 of them are for specifically identified runtimes and then the n'th one is a generic 'make our best guess' fallthrough. That should not turn out too bad in the code.

metrics/lib/common.bash
bwarden: Single square bracket is a separate program, vs. bash built-in double bracket. I'll look at simplifying the loops.

bwarden:
OK, let's try this... I reduced the nesting, moved TOML parsing to a separate function, added support for "cc-runtime cc-env", tightened up the testing of potential results, made parsing case-insensitive and more flexible to handle cc-env, and added the test for pc-lite (whoops, forgot to amend that one).

It feels a little bit dirty using a tmpfile for the results of cc-env (which I clean up afterward), but it lets me use the same code path as for parsing config files, vs. more complexity. I'm open to suggestions.

metrics/lib/common.bash
```diff
@@ -42,3 +42,80 @@ function save_results(){
...
+
+    # Cleanup
+    [[ -n "${cc_env_tmp}" ]] && rm "${cc_env_tmp}"
+
+    # Check whether we got a good result
+    [[ -n "$qemu_path" ]] || die Failed to find qemu path in $conf_file
+    [[ -f "$qemu_path" ]] || die "$qemu_path does not exist"
+    [[ -x "$qemu_path" ]] || die "$qemu_path is not executable"
+
+    # Check whether this qemu supports the required machine type
```
grahamwhaley: I'd drop the machine type check. As I think I noted, we are not tied to that model, and are about to move to a different one `q35`. I'd like the tests to be able to run no matter how the qemu or runtime are configured.

grahamwhaley:
with the note about the machine check,
lgtm
I think the pragmatic way forwards here is to get this merged so we can get the tests up with `cc-runtime`, and then if we need to adapt to get them up with `cc-oci-runtime` later, then so be it.

metrics/lib/common.bash
bwarden: Fixing.

grahamwhaley:
lgtm

grahamwhaley:
@jodh-intel OK with you now it is rewritten to use the cc-runtime env if available etc.

jodh-intel:
lgtm

PR #205, opened 2017-07-18
.ci: Install qemu Q35 from sources

### LINKED ISSUE #198: Use q35 instead of pc-lite machine type
We want to test with qemu q35 machine type because this will become the new standard for Clear Containers 3.0.

sboeuf:
There is a pull request on the runtime repository which introduces
the support for the machine type q35 for qemu. We need to install
this specific binary that we build from source on our CI environment.

Fixes #198

### DIFF
--- a/.ci/setup_env_ubuntu.sh
+++ b/.ci/setup_env_ubuntu.sh
@@ -37,6 +37,27 @@ chronic sudo -E apt-get update
 echo "Install qemu-lite binary"
 chronic sudo -E apt-get install -y --force-yes qemu-lite
 
+echo "Install qemu Q35 binary"
+chronic sudo -E apt-get install -y libcap-ng-dev libpixman-1-dev libcap-dev libattr1-dev
+git clone --branch qemu-lite-v2.9.0 https://github.com/clearcontainers/qemu.git --depth 1
+qemu_dir="qemu"
+pushd ${qemu_dir}
+git checkout qemu-lite-v2.9.0
+./configure --disable-tools --disable-libssh2 --disable-tcmalloc --disable-glusterfs        \
+	--disable-seccomp --disable-{bzip2,snappy,lzo} --disable-usb-redir --disable-libusb \
+	--disable-libnfs --disable-tcg-interpreter --disable-debug-tcg --disable-libiscsi   \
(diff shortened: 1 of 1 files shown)


.ci/setup_env_ubuntu.sh
```diff
@@ -37,6 +37,19 @@ chronic sudo -E apt-get update
 echo "Install qemu-lite binary"
 chronic sudo -E apt-get install -y --force-yes qemu-lite
 
+echo "Install qemu Q35 binary"
+chronic sudo -E apt-get install -y libcap-ng-dev libpixman-1-dev libcap-dev libattr1-dev
+git clone https://github.com/clearcontainers/qemu.git
```
devimc: since we don't care all the history, could you add ```--depth 1``` ?

.ci/setup_env_ubuntu.sh
```diff
@@ -37,6 +37,19 @@ chronic sudo -E apt-get update
...
+qemu_dir="qemu"
+pushd ${qemu_dir}
+git checkout qemu-lite-v2.9.0
+./configure --disable-tools --disable-libssh2 --disable-tcmalloc --disable-glusterfs --disable-seccomp --disable-{bzip2,snappy,lzo} --disable-usb-redir --disable-libusb --disable-libnfs --disable-tcg-interpreter --disable-debug-tcg --disable-libiscsi --disable-rbd --disable-spice --disable-attr --disable-cap-ng --disable-linux-aio --disable-brlapi --disable-vnc-{jpeg,png,sasl} --disable-rdma --disable-bluez --disable-fdt --disable-curl --disable-curses --disable-sdl --disable-gtk --disable-tpm --disable-vte --disable-vnc --disable-xen --disable-opengl --disable-slirp --enable-trace-backend=nop --enable-virtfs --enable-attr --enable-cap-ng --target-list=x86_64-softmmu
```
devimc: how about split this line in two or three lines?

.ci/setup_env_ubuntu.sh
sboeuf: Good catch! Done.

.ci/setup_env_ubuntu.sh
sboeuf: Sounds like a good idea :)

sboeuf: Done

PR #207, opened 2017-07-18
.ci: Install q35 at /usr/bin instead of /usr/local/bin

### LINKED ISSUE #206: Wrong path for q35, install at /usr/bin instead of /usr/local/bin

sboeuf:
The path where q35 is installed is not compliant with the one expected
by the build, meaning that tests with q35 machine type are failing
because the binary cannot be found.

Fixes #206


chavafg:
lgtm

Issue #209, opened 2017-07-19
CI: exclude all vendor'd .md files in all subtrees from docs review


grahamwhaley:
Now we have more sub-trees in the repo, each with their own vendored subdir, we should exclude any *.md file in a vendor tree from pullapprove docs review.
Modify the .pullapprove.yml file accordingly.

PR #204, opened 2017-07-18
checkmetrics: update vendoring

grahamwhaley:
Update vendoring, which includes:
 - move to pinned version of golang/dep
 - set packages to track 'master' branches
 - update, to pick up BurntSushi/toml license file changes

Fixes: #202

Signed-off-by: Graham Whaley <graham.whaley@intel.com>

PR #210, opened 2017-07-19
CI: exclude all vendor/*.md's from pullaprove

grahamwhaley:
We now have multiple sub-projects that have their own vendor
trees with .md files in them we do not want docs review to happen
for (as they are out of our control). Thus, exclude all vendor dirs
through the whole tree for .md file review.

Fixes: #209

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -47,9 +47,12 @@ groups:
           - "*.rst"
           - "*.md"
         exclude:
+# Don't include the vendor subdirs for docs review, as the docs files
+# in those trees our not under our control.
+# Exclude the main top level vendor dir, and all vendor dirs in all
+# sub projects.
           - "vendor/*"
-          - "cmd/fetchbranches/vendor/*"
-          - "cmd/localCI/vendor/*"
+          - "*/vendor/*"
     required: 1
     users:
       - iphutch


grahamwhaley:
Related to https://github.com/clearcontainers/tests/pull/204#issuecomment-316402388
As far as I can find pullapprove uses '*nix style matching', from this library:
https://docs.python.org/2/library/fnmatch.html
so, I think we need both those vendor exclusion lines, and believe the second double * line should work.
@devimc what do you think??
It's slightly difficult to test this without creating some python around that library or meddling with some .md files to check it out. We might just have to slide it in and see how it goes over time??

devimc:
lgtm

.pullapprove.yml
```diff
@@ -48,8 +48,7 @@ groups:
           - "*.md"
         exclude:
           - "vendor/*"
-          - "cmd/fetchbranches/vendor/*"
-          - "cmd/localCI/vendor/*"
+          - "*/vendor/*"
```
jodh-intel: Nit: would be good to have this next to the other vendor exclude. In fact, as this file is yaml, it should be possible to add a comment in it explaining why we have multiple vendor excludes, but I don't believe anyone has tested comments in `.pullapprove.yml` yet ;)

jodh-intel:
lgtm

.pullapprove.yml
grahamwhaley: @jodh-intel not sure I quite understand - the two vendor excludes are next to each other, and I see no other vendor excludes in this file?
https://github.com/grahamwhaley/tests/blob/15c6854485ccdb35156f131cd8d9b850812112bb/.pullapprove.yml#L50
Am I mis-understanding, or perhaps did you mis-read the patch?

Just went to look at pullapprove yaml comments - yep, the example on their front page has a number in it:
http://docs.pullapprove.com/
and, in fact, our file here has one:
https://github.com/grahamwhaley/tests/blob/15c6854485ccdb35156f131cd8d9b850812112bb/.pullapprove.yml#L7

So, sure, let me go add a couple of comments around the vendor exclusion bits...

grahamwhaley:
re-pushed with comments for the delectation of @jodh-intel

jodh-intel:
lgtm

PR #225, opened 2017-07-20
metrics: Add I/O storage performance test

MarioCarrilloA:
This test measures the bandwidth consumed by a specif I/O
operation (read, write, random-read etc) under certain configuration.
The result is obtained using FIO tool, it runs inside of a container.

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -59,6 +59,30 @@ function check_cmds()
 	done
 }
 
+# Initialization/verification environment. This function makes
+# minimal steps for metrics/tests execution.
+function init_env()
+{
+	cmd=("docker")
+	contr_running=$(docker ps -qa)
+
+	check_cmds "${cmd[@]}"
+
+	# Verify containers running
+	if [ ! -z "$contr_running" ];then
+		docker rm -f $(docker ps -qa)
+	fi
+
+	# Restart services
+	systemctl restart docker
+	runtime=$(docker info | grep -w "Default Runtime:" \
+			| awk '{print $3}')
+
+	if [ "$runtime" == "cor" ];then
+		systemctl restart cc-proxy
+	fi
+}
+
 function get_hypervisor_from_toml(){
     ## Regular expressions used for TOML parsing
     # Matches a section header
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -42,6 +42,23 @@ function save_results(){
 	bash $LIB_DIR/send_results.sh -n "$1" -a "$2" -r "$3" -u "$4"
 }
 
+# This function checks existence of commands.
+# They can be received standalone or as an array, e.g.
+#
+# cmds=(“cmd1” “cmd2”)
+# check_cmds "${cmds[@]}"
+function check_cmds()
+{
+	local cmd req_cmds=( "$@" )
+	for cmd in "${req_cmds[@]}"; do
+		if ! command -v "$cmd" > /dev/null 2>&1; then
+			die "command $cmd not available"
+			exit 1;
+		fi
+		echo "command: $cmd: yes"
+	done
+}
+
 function get_hypervisor_from_toml(){
     ## Regular expressions used for TOML parsing
     # Matches a section header
--- a/metrics/storage/fio_job.sh
+++ b/metrics/storage/fio_job.sh
@@ -1,0 +1,169 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#  Description of the test:
+#  This test measures the bandwidth measured by a specif I/O
+#  operation (read, write, random-read etc) under certain configuration.
+#  The result is obtained using FIO tool, it runs inside of a container.
+
+set -e
+
+# General env
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+TEST_ARGS="$@"
+
+# These values are a basic configuration for a FIO job
+# execution, however they could be overridden in order to
+# run a customized FIO job.
+BLOCK_SIZE="4k"
+ENGINE="sync"
+FILE_SIZE="4G"
+FIO_TEST_NAME="io_test"
+NUM_JOBS="1"
+OPERATION="randread"
+RUNTIME="cor"
+TEST_NAME="storage fio test"
+
+# This docker image includes FIO tool.
+FIO_IMAGE="clusterhq/fio-tool"
+
+# The FIO tool included in clusterhq/fio-tool docker image
+# works in "/tmp" by default. "/tmp" is overridden by /home
+# mount point in order to avoid tmpfs and use the disk.
+FILE_NAME="/home/fio_test"
+
+declare FIO_JOB
+
+function help()
+{
+	usage=$(cat << EOF
+Usage: $0 [-h] [options]
+   Description:
+        This script show I/O bandwidth in KB/s
+        of containers using FIO tool.
+   Options:
+        -h,    Help page.
+        -b,    Block size for I/O units.
+        -e,    FIO engine.
+        -f,    Total  size  of  I/O for this job.
+        -o,    I/O operation e.g. read, write etc.
+        -n,    Max number of jobs.
+        -r,    Runtime for docker.
+        -t,    Test name.
+EOF
+)
+	echo "$usage"
+}
+
+# Create the job configuration for FIO
+# this job will run inside of the container.
+function create_fio_job()
+{
+	FIO_JOB="$(cat << EOF
+	#!/bin/bash
+	# This fio job will run inside of
+	# the container.
+
+	fio --ioengine="$ENGINE" \
+		--name="$FIO_TEST_NAME" \
+		--filename="$FILE_NAME" \
+		--bs="$BLOCK_SIZE" \
+		--size="$FILE_SIZE" \
+		--readwrite="$OPERATION" --max-jobs=$NUM_JOBS
+EOF
+)"
+}
+
+# Parse the output of FIO tool in order to get
+# the bandwidth average cosunmed by the FIO job.
+# Also this function will convert results in KB/s
+# if it is necessary.
+function parse_fio_results()
+{
+	raw_data="$1"
+	bw_data=$(echo "$raw_data" |  grep "bw (")
+
+	if [ -z "$bw_data" ];then
+		die "bandwidth result not found: check FIO job configuration"
+	fi
+
+	units=$(echo "$bw_data" grep "bw (" | grep -o "[A-Z]*")
+	avg=$(echo "$bw_data" | awk -F "," '{print $4}' | cut -d "=" -f2)
+	stdev=$(echo "$bw_data" | awk -F "," '{print $5}' | cut -d "=" -f2)
+
+	echo "Units: $units"
+	case "$units" in
+		MB)
+		    avg=$(echo "$avg * 1024" | bc -l)
+		    ;;
+		GB)
+		    avg=$(echo "($avg * 1024) * 1024" | bc -l)
+		    ;;
+		esac
+
+	save_results "$TEST_NAME" "$TEST_ARGS" "$avg" "KB/s"
+	echo "$avg KB/s"
+}
+
+function main()
+{
+	cmds=("bc" "awk")
+	local OPTIND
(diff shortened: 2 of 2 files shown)


metrics/storage/fio_job.sh
```diff
@@ -0,0 +1,162 @@
...
+
+# General env
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+# Fio job default configuration
+BLOCK_SIZE="4k"
+DEBUG_FILE="debug.log"
+ENGINE="sync"
+FILE_NAME="/home/fio_test"
```
jodh-intel: This looks rather odd. Is this an output file? If so, maybe give it a file extension (`.txt`, `.log` or similar). But also, why is it being written to `/home/`?

metrics/storage/fio_job.sh
```diff
@@ -0,0 +1,162 @@
...
+FILE_SIZE="4G"
+FIO_IMAGE="clusterhq/fio-tool"
```
jodh-intel: Again, this looks slightly odd to me. Maybe you could put a comment about the variables that specified paths giving some details as to why they were chosen.

metrics/storage/fio_job.sh
grahamwhaley: I'll let @MarioCarrilloA answer as well, but iirc, where you place the test file directly influences which type of container mount point you are actually testing the bandwidth of (9p, union, ram etc.).
I believe we are using the `/home/` here to force use of a specific mount point, and will let @MarioCarrilloA explain (and then probably add a comment to the file to make that clearer)

metrics/storage/fio_job.sh
grahamwhaley: That's a fair question.
The alternative (which we should consider for the future) is to construct our own docker image and upload the matching docker file here (somewhat like this https://github.com/01org/cc-oci-runtime/tree/master/tests/metrics/network/dockerfile_image). We have some infra to sort out about exactly where we store those built dockerfiles for pulling down though before we can do that, as we don't want to be rebuilding (the sometimes reasonably large) docker iamges from raw files every time we run the metrics.

metrics/storage/fio_job.sh
```diff
@@ -0,0 +1,162 @@
...
+		esac
+	done
+	shift $((OPTIND-1))
+
+	check_cmds "${cmds[@]}"
+	init_env
+	create_fio_job
+
+	# Launch container (cor runtime by default)
+	output=$(docker run --runtime="$RUNTIME" \
```
grahamwhaley: Just a note - I'm OK with us here using the default path/system docker, as that is generally where we have installed the runtime we are testing. But, just to note, possibly in the future we may have to make this a `$DOCKER` or similar that will be set by the metrics framework.

metrics/storage/fio_job.sh
jodh-intel: That makes sense but in that case, the script needs to assert that the environment is as it expects. We could construct crazy-long `docker run` command-lines, but I don't see why there can't be a "pre-test" `docker run` check that does something like `cd /some/where && mount |grep something || exit 1` to check that the mount point is 9p, ramfs, blah.

metrics/storage/fio_job.sh
jodh-intel: +1. I was thinking the same as I saw all the `docker` calls in `init_env` (but forgot to comment on that :)

metrics/lib/common.bash
```diff
@@ -42,3 +42,21 @@ function save_results(){
 	bash $LIB_DIR/send_results.sh -n "$1" -a "$2" -r "$3" -u "$4"
 }
 
+# This function checks existence of commands.
```
MarioCarrilloA: no problem, I will fix that :)

metrics/storage/fio_job.sh
MarioCarrilloA: Hi @jodh-intel, when you do not specify the path of filename, FIO will work in /tmp by default, we are interested about execution using the disk.

metrics/storage/fio_job.sh
MarioCarrilloA: agree

metrics/lib/common.bash
```diff
@@ -60,3 +60,27 @@ function check_cmds()
...
+	contr_running=$(docker ps -qa)
+
+	check_cmds "${cmd[@]}"
+
+	# Verify containers running
+	if [ ! -z "$contr_running" ];then
+		docker rm -f $(docker ps -qa)
+	fi
+
+	# Restart services
```
MarioCarrilloA: I think it is necessary in order to get more reliable results, however as you said, it is better to describe as a "WARNING" in documentation.

metrics/storage/fio_job.sh
```diff
@@ -0,0 +1,162 @@
+#!/bin/bash
```
MarioCarrilloA: let me change it

metrics/lib/common.bash
```diff
@@ -60,3 +60,27 @@ function check_cmds()
...
+	systemctl restart docker
+	runtime=$(docker info | grep -w "Default Runtime:" \
+			| awk '{print $3}')
+
+	if [ "$runtime" == "cor" ];then
```
grahamwhaley: I opened #234 so we don't completely forget :-)

grahamwhaley:
@MarioCarrilloA did you post some changes for us to review (I will go look...) - best you give us an explicit 'changes pushed, please re-review' to make sure we don't miss them ;-)

metrics/lib/common.bash
```diff
@@ -42,6 +42,49 @@ function save_results(){
...
+# They can be received standalone standalone or as an array, e.g.
+#
+# cmds=(“cmd1” “cmd2”)
+# check_cmds "${cmd[@]}"
+function check_cmds()
+{
+	req_tools="$@"
+	for cmd in ${req_tools[*]}; do
+		command -v "$cmd" &> /dev/null
+		res=$?
```
grahamwhaley: Hi @MarioCarrilloA I tried a test of this on my Ubuntu machine, and I got silent failure I believe as the script invoking this (fio_tests.sh in this case) has `-e` set, so quits at the 'command' failure instead of calling the `die` command.
I added a simple `set +e` inside check_cms() to verify this, and that did seen to be the issue. I don't know if that is the best solution though. /cc @bwarden both for any input on the shell magic here, and also as I think @bwarden should be aware of the functions you are adding here so we can review/share/re-use across all the metrics tests.

grahamwhaley:
Hi @MarioCarrilloA It occurred to me, fio_test.sh should also be added to https://github.com/clearcontainers/tests/blob/master/metrics/run_all_metrics.sh Did you have a plan - to do that either in this PR, or maybe with a follow up PR - or some other thoughts?

MarioCarrilloA:
Hi @grahamwhaley , I would like to add fio_test.sh to run_all_metrics in another PR, it would be great to review the configuration/execution.

metrics/lib/common.bash
MarioCarrilloA: good catch!, I will add that

MarioCarrilloA:
@grahamwhaley @chavafg fixed touched points in comments :)

metrics/lib/common.bash
```diff
@@ -42,6 +42,27 @@ function save_results(){
...
 }
 
+# This function checks existence of commands.
+# They can be received standalone standalone or as an array, e.g.
+#
+# cmds=(“cmd1” “cmd2”)
+# check_cmds "${cmd[@]}"
+function check_cmds()
+{
+	set +e
```
grahamwhaley: OK, I'm going to ask for some input on shell-ness here from @MarioCarrilloA and @bwarden (coz otherwise I just have to go look it up, and you guys probably already know....)
the `set -e`, does that just have effect in this func - are funcs run in sub-shells? (I think they are).
If yes, then this is fine, but we need to remove the `set -e` at the bottom of the func as it is redundant.
If no, then we should not be blatantly doing the `set -e` at the end of the func, as you may have just turned on `-e` for a script that did not have it set when it called this func.
So, give me your input, and let's get this fixed ;-)

bwarden: I'm not totally clear on what will happen, or how to get the desired behavior. This is most interesting: https://delx.net.au/blog/2016/03/safe-bash-scripting-set-e-is-not-enough/
In a nutshell, "set -e" often doesn't do what you want or expect, so we should probably not be counting on it in these tests at any point, and instead writing explicit checks.

bwarden: As I understand, functions are *not* run in sub-shells, unless you call them so (e.g. "(check_cmds)"), and bash just helpfully ignores your set +/-e while you're in a function, or an if-then-else equivalent, or some other not-so-clearly defined special case.

icarus-sparry: The "set -e" is a global switch for the script. @bwarden is correct that the functions are not run in a subshell. The easy way to see if "set -e" is already in effect is to look at the variable $- (dollar minus) which has the currect flags. So you can do

local OLDFLAGS=$-
set +e
....
case "$OLDFLAGS" in
    (*e*) ;; # We were already in a 'set +e' state, do nothing
    (*) set -e ;;
esac

icarus-sparry: Having said all that, and now looked at the code, we don't need the "set [+-]e" at all if we just rewrite the check to be part of the "if" statement.

We can loose the "function" keyword, this is not POSIX syntax.
Lets see if I can do markdown correctly.

    check_cmds()
    {
         for cmd ; do
              if ! command -v "$cmd" >/dev/null 2>&1 ; then
                   die "command $cmd not available"
                   exit 1
              fi
              printf 'command: %s: yes\n' "$cmd"
          done
     }

MarioCarrilloA: well, about "function" keyword, I was following a convention in that file,also I would like to keep "function" keyword in order to avoid "alias" collisions

icarus-sparry: I can't argue against style, not matter how bad!

I don't think there are any alias issues, can you explain?

Note that 'ksh' also has a 'function' keyword, and it has very different effects, causing different scoping of variables.

MarioCarrilloA: https://stackoverflow.com/questions/7917018/what-is-the-function-keyword-used-in-some-bash-scripts

grahamwhaley:
@icarus-sparry thanks for the feedback and suggestions. Looks like `function` is a pro/con choice thing. Sure, if we were writing all the scripts in all the repos again then I think maybe we'd drop `function` - and I'll keep that in mind for the future - but, as we have many scripts all in this style we'll stick with the style, but keep in the back of our minds the type of issues that might come up.
@MarioCarrilloA thanks for the updates. I think we hit #253 on semaphore though, so I think you will need to rebase and push to clear that. Other than that,
lgtm

metrics/lib/common.bash
```diff
@@ -42,6 +42,47 @@ function save_results(){
 	bash $LIB_DIR/send_results.sh -n "$1" -a "$2" -r "$3" -u "$4"
 }
 
+# This function checks existence of commands.
+# They can be received standalone standalone or as an array, e.g.
```
chavafg: duplicate "standalone"?

metrics/lib/common.bash
```diff
@@ -42,6 +42,47 @@ function save_results(){
...
+#
+# cmds=(“cmd1” “cmd2”)
+# check_cmds "${cmd[@]}"
+function check_cmds()
+{
+	req_tools="$@"
```
icarus-sparry: This is incorrectly handling the arguments.

We should not be creating or altering global variables.

The correct thing to do IMNSHO is to use the short form of 'for',  i.e. 'for cmd ; do' and drop the req_tools as I showed in the previous comment.

If you are going to keep the req_tool, then you want to 
1. declare it to be local
1. you want to use an array assignment 
1. use the correct syntax to loop over the elements

As written, this code takes all the arguments, joins them together (usually with a space), and assigns this string to a scalar variable. It then treats this scalar as a one element array, takes all (one) element of this array (if there was more than one it would join them together with spaces), then splits again on spaces. It then splits apart again and loops over the words. All of this joining and splitting allows things to go wrong.

For this function, where you probably are not going to have spaces in the command names, you can get away with it, but in other functions you have the same pattern and do not know if you are going to have spaces in the data.

So this should say (I am leaving in 'function' to avoid the argument, although the links you provided are wrong, there is no alias issue in scripts)

    function check_cmds()
    {
        local cmd req_tools=( "$@" )
        for cmd in "${req_cmds[@]}"; do
            if ! 
    ....

MarioCarrilloA: global var??? I am not sure what do you refer, could you explain that?

icarus-sparry: Before you run check_cmds there is no global variable "req_tool". After you have run it there is. The same is true of "cmd".

In shell scripting all variables are global by default, and are dynamically scoped.  This is in contrast for example with Python, where 'req_tools=1' would create a local variable if there isn't an existing global.

MarioCarrilloA: I have tested the advised changes in a pair of comments below and it did not work, however as I understand I just need to declare req_tool as local right?

icarus-sparry: Declaring it local fixes one of the 3 things that is wrong.

The best test case is to add another directory to you PATH, and put a command in it with a space in the name, e.g. "do it", and then see what happens when you call

     check_cmds "do it"

grahamwhaley:
@icarus-sparry I believe @MarioCarrilloA has updated - OK with you now?

PR #228, opened 2017-07-20
tests: add docker info integration test

### LINKED ISSUE #227: add docker info integration test
add to the integration tests $ docker info

GabyCT:
This will test the docker info command

Fixes #227

### DIFF
--- a/integration/docker/info_test.go
+++ b/integration/docker/info_test.go
@@ -1,0 +1,39 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
+import (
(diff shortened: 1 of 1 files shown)


integration/docker/info_test.go
```diff
@@ -0,0 +1,39 @@
...
+	)
+
+	Context("info with docker", func() {
+		BeforeEach(func() {
+			command = NewCommand(Docker, "info")
+			exitCode = command.Run()
+		})
+		It("should not be empty", func () {
+			Expect(exitCode).To(Equal(0))
+			Expect(command.Stdout.String()).To(ContainSubstring("Runtime"))
```
jodh-intel: Can you check for the expected runtime name in the output too?

GabyCT: @jodh-intel docker info unfortunately does not return a runtime path, it only returns cc3 or runc. So should I look the word of cc3?

jodh-intel: well if you can reliably determine which runtime docker is configured for, I'd say yes, assert that it is `cc3` or whatever.

devimc: I think you need to process the output to get the default runtime (```Default Runtime:```) and then try to get the path

This is an example in bash
```
$ docker info | grep Default
Default Runtime: cc30
$ ps aux | grep dockerd | egrep -o "cc30=[[:graph:]]*"
cc30=/usr/local/bin/cc-runtime
```

grahamwhaley: That's... sneaky :-)  Do we think that might work for all runtime additions to docker(d) - I guess that is the default way we add our runtimes, so maybe. /cc @bwarden just in case we steal (at least some of) that idea later on when we are trying to determine actual docker configured runtime binary paths etc.

PR #245, opened 2017-07-24
tests: Add docker logs test

### LINKED ISSUE #244: add docker logs test
add functional docker logs test

GabyCT:
Add integration docker logs test

Fixes #244

### DIFF
--- a/integration/docker/logs_test.go
+++ b/integration/docker/logs_test.go
@@ -1,0 +1,61 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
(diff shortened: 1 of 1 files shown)


GabyCT:
setup issue #250

integration/docker/logs_test.go
```diff
@@ -0,0 +1,57 @@
...
+	runCommand := func(args []string, expectedExitCode int) {
+		command = NewCommand(Docker, args...)
+		Expect(command).ToNot(BeNil())
+		exitCode = command.Run()
+		Expect(exitCode).To(Equal(expectedExitCode))
+	}
+
+	BeforeEach(func() {
+		id = RandID(30)
+		args = []string{"run", "-td", "--name", id, Image, "sh", "-c", "echo hello"}
```
grahamwhaley: Do we check that we are getting `hello` back in the logs? I can't see that being done, and I think that is what we will get back from this echo, is it?

integration/docker/logs_test.go
```diff
@@ -0,0 +1,57 @@
...
+	})
+
+	AfterEach(func() {
+		Expect(ContainerRemove(id)).To(BeTrue())
+		Expect(ContainerExists(id)).NotTo(BeTrue())
+	})
+
+	Describe("logs with docker", func() {
+		Context("check logs functionality", func() {
+			It("should work", func() {
```
grahamwhaley: My instinct is that we should also have a test for something that does not work. The only one I can really think of is to ask for the logs from a container that does not exist, but then that is probably handled in docker itself without invoking any runtime parts - so, maybe that would not really be much use or addition here. Open for other ideas though.

devimc:
lgtm

PR #253, opened 2017-07-24
.ci: Fix Semaphore CI build by installing libostree from sources

### LINKED ISSUE #250: crio setup scripts are broken in semahpore ci.
https://github.com/kubernetes-incubator/cri-o/pull/687 added a new  dependency `libostree-dev` . We need to add it to ubuntu setup to build cri-o

*Debian, Ubuntu, and related distributions will also need a copy of the development libraries for ostree, either in the form of the libostree-dev package from the flatpak PPA, or built from source (more on that here).*

sboeuf:
CRI-O is now relying on a new library called libostree and there is
no Ubuntu packages available before Ubuntu 16.10. Because Semaphore
runs on Ubuntu 14.04, we need to install this library from sources.

Fixes #250

### DIFF
--- a/.ci/setup_env_ubuntu.sh
+++ b/.ci/setup_env_ubuntu.sh
@@ -112,3 +112,15 @@ make nsenter
 sudo cp nsenter /usr/bin/
 popd
 rm -rf util-linux-${nsenter_version}/ util-linux-${nsenter_version}.tar.xz
+
+echo "Install ostree"
+ostree_dir="ostree"
+chronic sudo -E apt-get install -y liblzma-dev e2p-dev libfuse-dev gtk-doc-tools
+git clone https://github.com/ostreedev/ostree.git
+pushd ${ostree_dir}
+env NOCONFIGURE=1 ./autogen.sh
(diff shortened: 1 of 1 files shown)


sboeuf:
This PR fixes our CI, let's merge it asap!

.ci/setup_env_ubuntu.sh
```diff
@@ -112,3 +112,15 @@ make nsenter
 sudo cp nsenter /usr/bin/
 popd
 rm -rf util-linux-${nsenter_version}/ util-linux-${nsenter_version}.tar.xz
+
+echo "Install ostree"
+ostree_dir="ostree"
+chronic sudo -E apt-get install -y liblzma-dev e2p-dev libfuse-dev gtk-doc-tools
+git clone https://github.com/ostreedev/ostree.git
```
grahamwhaley: just a passing note - you might be able to do a 'shallow clone' here, as we will only be building from the tip (HEAD) of the source, and thus save us some bandwidth and time.

grahamwhaley:
lgtm

PR #255, opened 2017-07-25
tests: Add docker kill test

### LINKED ISSUE #254: add docker kill
add integration docker kill test

GabyCT:
Add integration docker kill test

Fixes #254

### DIFF
--- a/integration/docker/kill_test.go
+++ b/integration/docker/kill_test.go
@@ -1,0 +1,61 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
+import (
+	. "github.com/clearcontainers/tests"
+	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


integration/docker/kill_test.go
```diff
@@ -0,0 +1,58 @@
...
+	Describe("kill with docker", func() {
+		Context("kill a container", func() {
+			It("should assigned exited status", func() {
+				id = RandID(30)
+				args = []string{"run", "-td", "--name", id, Image, "sh"}
+				runCommand(args, 0)
+				args = []string{"kill", id}
+				runCommand(args, 0)
+				Expect(command.Stdout.String()).To(ContainSubstring(id))
+				args = []string{"inspect", id, "--format='{{.State.Running}}'"}
```
grahamwhaley: Hi.
 I think the arguments to inspect are in the wrong order here. When I tried this by hand I have to put the --format before the id, otherwise --format is interpreted as the name of another container. I think we pass the test as it stands at present as inspect first returns all of the info for the id, which on my system includes 15 items that are 'false', and hence match the ContainSubstring.
The fix is obviously just to flip around the id and --format argument order I believe.

integration/docker/kill_test.go
```diff
@@ -0,0 +1,58 @@
...
+		Expect(ContainerRemove(id)).To(BeTrue())
+		Expect(ContainerExists(id)).NotTo(BeTrue())
+	})
+
+	Describe("kill with docker", func() {
+		Context("kill a container", func() {
+			It("should assigned exited status", func() {
+				id = RandID(30)
+				args = []string{"run", "-td", "--name", id, Image, "sh"}
+				runCommand(args, 0)
```
grahamwhaley: Just for a little extra safety check, and to add some extra uniformity to the code, can we add an inspect test here to check that the container did come up 'running' before we then kill it?

GabyCT:
@grahamwhaley changes applied

PR #256, opened 2017-07-25
localCI: remove generated files when make cleaning

grahamwhaley:
make clean rule was not removing the auto generated source
files. Let's do that, to really be clean.

### DIFF
--- a/cmd/localCI/Makefile
+++ b/cmd/localCI/Makefile
@@ -93,4 +93,4 @@ check: $(GENERATED_FILES) Makefile
 	$(QUIET_CHECK)go test .
 
 clean:
-	rm -f $(TARGET) $(UNIT_FILES)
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@devimc let me know if you think this is the right thing to do. Felt right to me ;-)
I guess we might make the generated code dependant on the Makefile as well?

grahamwhaley:
Ah, OK. Travis fail due to no `Fixes` in the commit - I'll await feedback from @devimc before I go 'fix' that and re-push.

devimc:
Hi @grahamwhaley  
remove generated files is needed, did you find an issue?

grahamwhaley:
Not a 'hard' issue. I modified the `PREFIX` in the Makefile, did a `make clean; make`, and the generated file did not get changed - so, thought we should fix either/both the `make clean` rule and the dependency of the generated files on the Makefile. What do you think?

Issue #258, opened 2017-07-25
Use a known working version for crio


jcvenegas:
Because we use master to tests crio sometimes breaks our CI ( #257 #250  ):lets have a kown version to test and update when we know it works well.

PR #259, opened 2017-07-25
ci: add test-versions.txt

jcvenegas:
crio updates has been updated frecuently and
break our ci. Lets add a lock version.

This version could be used for CC 3.0
release notes.

Fixes: #258

### DIFF
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -18,9 +18,13 @@ set -e
 
 cidir=$(dirname "$0")
 
+source "${cidir}/../test-versions.txt"
+
 echo "Get CRI-O sources"
 go get -d github.com/kubernetes-incubator/cri-o || true
 pushd $GOPATH/src/github.com/kubernetes-incubator/cri-o
+git fetch
+git checkout "${crio_version}"
 
 echo "Installing CRI-O"
 make install.tools
--- a/test-versions.txt
+++ b/test-versions.txt
@@ -1,0 +1,2 @@
+#well known working crio tag/commit/branch
+crio_version=065960386fce7088bab3fc22af44ebbbf75d3641


.ci/install_crio.sh
```diff
@@ -18,9 +18,13 @@ set -e
 
 cidir=$(dirname "$0")
 
+source "${cidir}/../test-versions.txt"
+
 echo "Get CRI-O sources"
 go get -d github.com/kubernetes-incubator/cri-o || true
 pushd $GOPATH/src/github.com/kubernetes-incubator/cri-o
+git fetch
```
sboeuf: No need for git fetch

sboeuf:
Need to remove the unneeded `git fetch`, but other than that:
LGTM

test-versions.txt
```diff
@@ -0,0 +1,2 @@
+#well known working crio tag/commit/branch
+crio_version=a95bbe9608d3b09b50651c82867bf7cfd83833e8
```
sboeuf: Actually, this more recent commit `065960386fce7088bab3fc22af44ebbbf75d3641` is working.

grahamwhaley:
taking into account feedback already given,
lgtm

jcvenegas:
sha updated

.ci/install_crio.sh
jcvenegas: @sboeuf  @grahamwhaley  I consider the case someone run `setup` locally but cri-o was already cloned but in an old version. we need to fetch and checkout to the commit we want. In the case of semaphore all is cloned in the moment. but if this setup will run only in semaphore I can remove it.

sboeuf:
LGTM

.ci/install_crio.sh
grahamwhaley: Ah, OK. I understand. I have similar niggles trying to make the static checker scripts runnable by anybody from anywhere that I need to get fixed.....
It is a tricky one. Sure, it would be nice to be able to run the scripts under any circumstance, but they do live in the `.ci` dir, so I don't think it is out of order for them to fail in strange setups.
Ultimately, I don't think we have to cater for hand runs if it costs us. In this case, I don't think the `fetch` will cost us much - so, if @sboeuf agrees, I would ask we have a clear comment added above the fetch explaining why it is there - OK?

sboeuf:
@chavafg or @grahamwhaley please merge it before we get another breakage from CRI-O.

grahamwhaley:
@sboeuf it is blocked on a 'reviewer with write access' - let me see if I can clear that...

grahamwhaley:
ok, let it rip...

PR #272, opened 2017-07-26
CI: Install systemd file for changing default runtime in docker

### LINKED ISSUE #271: cc-runtime is not set as default docker when using systemd
Right now our CI script `install_runtime.sh` just adds a docker option in `/etc/default` to change the default docker runtime, but this does not work for ubuntu 16.04+ as they use systemd.

chavafg:
If environment uses systemd, install the correct configuration
file for changing the default runtime in docker, reload the
systemd units and restart docker service.

Fixes #271.


devimc:
lgtm

gorozco1:
lgtm

devimc:
lgtm

PR #277, opened 2017-07-26
tests: Add docker export test

### LINKED ISSUE #276: add docker export test
add integration docker export test

GabyCT:
Add integration docker export test

Fixes #276

### DIFF
--- a/integration/docker/export_test.go
+++ b/integration/docker/export_test.go
@@ -1,0 +1,58 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
+import (
+	. "github.com/clearcontainers/tests"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"io/ioutil"
+	"os"
+)
+
+var _ = Describe("export", func() {
+	var (
+		args []string
+		id   string
+	)
+
+	BeforeEach(func() {
+		id = randomDockerName()
+		args = []string{"run", "-td", "--name", id, Image}
+		runDockerCommand(0, args...)
+	})
+
+	AfterEach(func() {
+		Expect(ContainerRemove(id)).To(BeTrue())
+		Expect(ContainerExists(id)).NotTo(BeTrue())
+	})
+
+	Describe("export with docker", func() {
+		Context("export a container", func() {
+			It("should export filesystem as a tar archive", func() {
+				file, err := ioutil.TempFile(os.TempDir(), "latest.tar")
+				Expect(err).ToNot(HaveOccurred())
+				defer file.Close()
+				defer os.Remove(file.Name())
+				args = []string{"export", "--output", file.Name(), id}
+				runDockerCommand(0, args...)
+				Expect(file.Name()).To(BeAnExistingFile())
+				fileInfo, err := file.Stat()
+				Expect(err).ToNot(HaveOccurred())
+				Expect(fileInfo.Size).NotTo(Equal(0))
+			})
+		})
+	})
+})


chavafg:
lgtm

integration/docker/export_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	Describe("export with docker", func() {
+		Context("export a container", func() {
+			It("should export filesystem as a tar archive", func() {
+				args = []string{"export", "--output=/tmp/latest.tar", id}
+				runCommand(args, 0)
+				_, err := os.Stat("/tmp/latest.tar")
+				if err != nil {
+					fmt.Printf("ERROR: %v\n", err)
+					os.Exit(1)
```
grahamwhaley: os.Exit feels a bit harsh here. Can we use a Ginkgo `Fail()` instead do you think?

integration/docker/export_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	AfterEach(func() {
+		Expect(ContainerRemove(id)).To(BeTrue())
+		Expect(ContainerExists(id)).NotTo(BeTrue())
+	})
+
+	Describe("export with docker", func() {
+		Context("export a container", func() {
+			It("should export filesystem as a tar archive", func() {
+				args = []string{"export", "--output=/tmp/latest.tar", id}
```
grahamwhaley: Maybe we should use a system generated tmpfile name rather than hard code a /tmp path?

integration/docker/export_test.go
```diff
@@ -0,0 +1,71 @@
...
+				runCommand(args, 0)
+				_, err := os.Stat("/tmp/latest.tar")
+				if err != nil {
+					fmt.Printf("ERROR: %v\n", err)
+					os.Exit(1)
+				}
+				cmd := exec.Command("rm", "-rf", "/tmp/latest.tar")
```
grahamwhaley: of course, it would be great if we could somehow check if the contents of the tar matched the container we ran up. Maybe we can do an exec into the container to generate a specifically named file, and then check that that file ends up in the tar file?

integration/docker/export_test.go
devimc: how bout you use ```Expect(err).ToNot(HaveOccurred())```

integration/docker/export_test.go
devimc: how bout you use ```os.RemoveAll```

integration/docker/export_test.go
```diff
@@ -0,0 +1,71 @@
...
+				err = cmd.Run()
+				if err != nil {
```
devimc: same here

integration/docker/export_test.go
devimc: also I think this should be part of ```AfterEach```

integration/docker/export_test.go
devimc: +1
https://golang.org/pkg/io/ioutil/#TempFile

GabyCT:
@devimc changes applied

integration/docker/export_test.go
```diff
@@ -0,0 +1,64 @@
...
+	})
+
+	Describe("export with docker", func() {
+		Context("export a container", func() {
+			It("should export filesystem as a tar archive", func() {
+				file, err := ioutil.TempFile(os.TempDir(), "latest.tar")
+				Expect(err).ToNot(HaveOccurred())
+				args = []string{"export", "--output", file.Name(), id}
+				runCommand(args, 0)
+				_, err = os.Stat(file.Name())
```
devimc: I think you could use ```BeARegularFile``` to check if the file exist

integration/docker/export_test.go
```diff
@@ -0,0 +1,64 @@
...
+				Expect(err).ToNot(HaveOccurred())
+				defer os.Remove(file.Name())
```
devimc: of use it just before ```runCommand```

integration/docker/export_test.go
devimc: or ```BeAnExistingFile```

GabyCT:
@grahamwhaley and @devimc changes applied

devimc:
lgtm

integration/docker/export_test.go
grahamwhaley: just to check @GabyCT , did you think it might be possible or not to inject a known file into the container which we can then check is in the tar file to ensure we did get the tar file we expected?

chavafg:
@grahamwhaley can you please check this again? Seems like it needs your review.
Thanks.

integration/docker/export_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+var _ = Describe("export", func() {
+	var (
+		args     []string
+		id       string
+		exitCode int
+		command  *Command
+	)
+
+	runCommand := func(args []string, expectedExitCode int) {
```
devimc: definitely this have to be a common function  😄

integration/docker/export_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	AfterEach(func() {
+		Expect(ContainerRemove(id)).To(BeTrue())
+		Expect(ContainerExists(id)).NotTo(BeTrue())
+	})
+
+	Describe("export with docker", func() {
+		Context("export a container", func() {
+			It("should export filesystem as a tar archive", func() {
+				file, err := ioutil.TempFile(os.TempDir(), "latest.tar")
```
devimc: do not forget to close the fd

integration/docker/export_test.go
```diff
@@ -0,0 +1,63 @@
...
+				Expect(err).ToNot(HaveOccurred())
+				defer os.Remove(file.Name())
+				args = []string{"export", "--output", file.Name(), id}
+				runCommand(args, 0)
+				Expect(file.Name()).To(BeAnExistingFile())
```
devimc: you should verify that the file is not empty

devimc: verify that the file is not empty

grahamwhaley:
@GabyCT will you be intending to rebase this onto #297 once that lands? in which case we can hold off this one a bit..., yes?

GabyCT:
@grahamwhaley changes applied, test is now using the common function

PR #301, opened 2017-08-02
localCI: do not fail if we don't know if the pr is mergeable

### LINKED ISSUE #300: localCI: do not fail if we don't know if the pr is mergeable
do not fail if we don't know if the pr is mergeable, just test it

devimc:
in the case that we do not know if a pull request is mergeable
localCI should not fail, we can just test the pull request

fixes #300


grahamwhaley:
lgtm

Issue #314, opened 2017-08-04
metrics: workload time: 'the input device is not a TTY'


MarioCarrilloA:
The docker run using interactive/terminal modes fails during execution by localCI.

PR #315, opened 2017-08-04
metrics: fix docker mode execution

MarioCarrilloA:
The terminal/interactive modes fail during the
execution of docker workload time test by localCI,
it shows: “the input device is not a TTY”.
This commit remove these options mode.

Fixes: #314

### DIFF
--- a/metrics/time/docker_workload_time.sh
+++ b/metrics/time/docker_workload_time.sh
@@ -47,7 +47,7 @@ function run_workload(){
 	# but set it in the subshell, as we'd like the docker command to fail and bail , rather than
 	# have the 'time' pass, but report how long it took for docker to fail.
 	set +e
-	(set -e; time -p docker run --rm -ti --runtime "$RUNTIME" "$IMAGE" "$CMD") &> "$TMP_FILE"
+	(set -e; time -p docker run --rm --runtime "$RUNTIME" "$IMAGE" "$CMD") &> "$TMP_FILE"
 	set -e
 
 	test_data=$(grep ^real "$TMP_FILE" | cut -f2 -d' ')
@@ -60,7 +60,7 @@ function run_workload(){
 		cat $TMP_FILE
 		echo "--------end of log file-------------"
 		rm -f $TMP_FILE
-		die "Failed to execute test 'docker run --rm -ti --runtime "$RUNTIME" "$IMAGE" "$CMD"'"
+		die "Failed to execute test 'docker run --rm --runtime "$RUNTIME" "$IMAGE" "$CMD"'"
 
 	fi
 


chavafg:
lgtm

metrics/time/docker_workload_time.sh
```diff
@@ -60,7 +60,7 @@ function run_workload(){
 		cat $TMP_FILE
 		echo "--------end of log file-------------"
 		rm -f $TMP_FILE
-		die "Failed to execute test 'docker run --rm -ti --runtime "$RUNTIME" "$IMAGE" "$CMD"'"
+		die "Failed to execute test 'docker run --rm --runtime "$RUNTIME" "$IMAGE" "$CMD"'"
```
grahamwhaley: nit, not directly related: seeing us (possibly actually me....) duplicate the `docker run...` text between the invocation and the error line makes me think we should lift this out to a constant.
We don't have to do that in this PR though, as I don't think the PR needs any other edits.

grahamwhaley:
I've had a sneaking feeling for some time that not having the `-ti` terminal options enabled would allow docker to disconnect from the container before it had finished its work, but, I am *wrong* :-)  Testing this change, it's fine, docker still waits for the result (I substituted my normal `true` tests for a `sleep 5` test to confirm).
I will note, adding a `-d` to the command does allow docker to disconnect (as you might expect) :-)
I will also note (and probably open an Issue), that I think we can, and probably should, drop the `-rm` from the test command, as afaik it does not involve the runtime itself, so just adds a constant time overhead to the results, which will only help mask any changes in actual runtime times.

lgtm

grahamwhaley:
and, I also meant to ask .... @devimc @chavafg do we think this might be an issue for some other tests (QA or metrics) that might need to run under localCI? It would be nice to confirm if this is a 'by design' or acceptable feature of localCI.

devimc:
@grahamwhaley , I can allocate a ```tty``` for running process, but I'd like to know if you really need this feature

grahamwhaley:
Hi @devimc 
I was thinking maybe QA would need it to do functional testing of tty functionality for instance...
And then I remember, that some of our metrics tests (such as the network iperf based tests) need to collect data from within the container to analyse the results (such as the output of the iperf command) - in which case I think we'll need a tty to get that stdout info ? (unless we re-write the tests to redirect the results into a file on the container workspace or a mounted volume).
My feeling is there will be a point when we will need tty functionality. We could hold off adding it until we hit that point, presuming you don't think it is difficult to add?

devimc:
@grahamwhaley I think it's easy to implement
https://github.com/clearcontainers/tests/issues/317

grahamwhaley:
As this is removing args we don't actually need anyway, I'm going to merge this, even though #317 should come along sometime.

PR #323, opened 2017-08-07
tests: Add docker user test

### LINKED ISSUE #386: docker swarm test - check interfaces
add integration docker swarm test

GabyCT:
Add integration docker user test to run a container as a root and
non-root user.

Fixes #386

### DIFF
--- a/integration/docker/user_test.go
+++ b/integration/docker/user_test.go
@@ -1,0 +1,63 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
+import (
+	. "github.com/clearcontainers/tests"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("user", func() {
+	var (
+		args []string
+		id   string
+	)
+
+	BeforeEach(func() {
+		id = randomDockerName()
+	})
+
+	AfterEach(func() {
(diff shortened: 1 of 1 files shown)


integration/docker/user_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	AfterEach(func() {
+		Expect(ContainerRemove(id)).To(BeTrue())
+		Expect(ContainerExists(id)).NotTo(BeTrue())
+	})
+
+	Describe("set user with docker", func() {
+		Context("run as non-root user", func() {
+			It("should display the non-root user", func() {
+				Skip("Issue https://github.com/clearcontainers/runtime/issues/386")
```
jodh-intel: Nice.

integration/docker/user_test.go
```diff
@@ -0,0 +1,56 @@
...
+				args = []string{"run", "--name", id, "-u", "postgres", "postgres", "whoami"}
```
jodh-intel: Although this is the image that discovered the problem, I think it may make sense to use a smaller image simply to switch user. I haven't searched exhaustively, but the `alpine` image is a lot smaller and has 2 users defined aside from root: `postgres` and `operator`.

grahamwhaley: +1
`alpine` looks to have many users. `busybox` also has a few. I tested both with user `mail` for instance.

integration/docker/user_test.go
```diff
@@ -0,0 +1,56 @@
...
+				stdout := runDockerCommand(0, args...)
+				Expect(stdout).To(ContainSubstring("postgres"))
```
grahamwhaley: Just to check - if this fails, could the potential error output not container the failed user name as a substring? For instance, I just ran:
```
docker run -u fred busybox sh
docker: Error response from daemon: linux spec user: unable to find user fred: no matching entries in passwd file.
```

Maybe we want to be more precise on the matching?

GabyCT:
@jodh-intel and @grahamwhaley changes applied

jodh-intel:
I just saw another comment from @grahamwhaley on another PR: could you add a test to ensure that searching for an unknown user fails? It's tricky to guarantee the user won't exist, but you could always `userdel` the user first, commit to create a new image and use that to guarantee it. That said, it should be pretty safe to assume that something crazy like user `aaaaaaaaaaaa` doesn't exist (famous last words maybe... ;-)

GabyCT:
@jodh-intel changes applied

chavafg:
lgtm

PR #324, opened 2017-08-07
Add support for getting configuration from default path

MarioCarrilloA:
If the the configuration file path is not specified by command line, emailreport tool will look for that file in a system path by default. This file will be shipped during installation step.


jodh-intel:
lgtm.

-----
Aside: seeing this change, you might want to raise a follow-on PR to add a `--version` option which can be very useful for debugging. If you do that, note the bug in `go build` which means you can only specify `-ldflags` once (see https://github.com/clearcontainers/proxy/pull/91).

cmd/emailreport/main.go
```diff
@@ -73,6 +77,12 @@ func main() {
 	flag.StringVar(&confFile, "f", "", "Configuration file")
 	flag.Parse()
 
+	// If there is any input file specified by command line
```
grahamwhaley: s/is any/is not any/ ?

grahamwhaley:
Single probably typo in comment to fix, then should be fine to ack.

I sort of have a general distaste for those Makefiles invoking go build with options etc., but I guess there is no other better way. My gut just wants to stick with `go build` and `go install`... Same with the version comment, I wish there were a nicer way to generate and track version numbers. But, I suspect there is no better way we know of (apart from hand editing/tracking directly in the source).
To note, https://github.com/clearcontainers/tests/tree/master/cmd/checkmetrics is also missing versioning and a default config file path (but may gain them one day soon - in fact I suspect @MarioCarrilloA will run into that next ;-) )

jodh-intel:
@grahamwhaley - I tend to agree, but the golang tooling of today just doesn't seem to be sophisticated enough for our (not unreasonable) needs.

cmd/emailreport/main.go
MarioCarrilloA: ups! you are right

chavafg:
lgtm

PR #325, opened 2017-08-07
tests: Add attach docker test

### LINKED ISSUE #243: add docker attach test
add docker attach integration test

GabyCT:
Add integration attach docker test.

Fixes #243


jodh-intel:
lgtm

grahamwhaley:
As per usual, I wonder if we should be checking for the expected failure cases as well - check that trying to attach to a non-existent container fails, and check that attaching to an already attached container and quitting then quits both attaches for instance.
But, that does not have to block this:
lgtm

PR #330, opened 2017-08-08
Add support for getting baseline file from default path 

MarioCarrilloA:
If the the baseline file path is not specified by command line, checkmetrics tool will look for the baseline file in a system path by default. This baseline file will be shipped during installation step.


jodh-intel:
lgtm

grahamwhaley:
From the diff it looks like git didn't pick up that the toml file was moved/renamed - but, that's not the end of the world. For reference, I suspect `git mv` might be the easiest way to ensure git has full knowledge.

lgtm

PR #340, opened 2017-08-09
scripts: Add script to generate hypervisor configure options.

### LINKED ISSUE #339: Define qemu build options *once*
Currently, we have the qemu build (well `configure`) options scattered around the place:

- https://github.com/clearlinux-pkgs/qemu-lite/blob/master/configure
- https://github.com/clearcontainers/tests/blob/master/.ci/setup_env_ubuntu.sh
- https://github.com/clearcontainers/packaging/blob/master/qemu-lite/qemu-lite.spec
- https://github.com/clearcontainers/packaging/blob/master/qemu-lite/debian.rules

... and we're about to add:

- https://github.com/clearcontainers/packaging/blob/master/qemu-cc/qemu-cc.spec
- https://github.com/clearcontainers/packaging/blob/master/qemu-cc/debian.rules

This is not only confusing, it's potentially dangerous given some of the options relate to security.
There isn't much we can do about the clear linux build as that is a separate project, but for the remainder we need to define a *single* location these configure options live in.

## Proposal

1. Create https://github.com/clearcontainers/tests/tree/master/scripts/configure-hypervisor.sh
1. Add each option on a separate line.
1. Each option *must* have a comment explaining why it is being used.
1. Make all repos and scripts that need to build `qemu`/`qemu-lite`/`qemu-cc/`qemu-blah` run the script to determine the correct build options and thus guarantee the various builds are consistent.

Notes:

- this will be a change to the CI build which should mirror exactly how the packaged versions of `qemu*` are built.
- some consumers of the script will probably have to do a bit of work to import the output of this script (maybe by renaming config files to `file.in` and adding a `@HYPERVISOR_BUILD_OPTIONS@` or similar into `file.in`).

jodh-intel:
To avoid a proliferation of scripts and config files from specifying
different sets of hypervisor configuration options, this script can be
run to generate the correct set of options.

Each option group has been documented explaining why it has been specified.

The script currently only works for "qemu*" hypervisors but could be
extended in the future.

Fixes #339.

### DIFF
--- a/scripts/configure-hypervisor.sh
+++ b/scripts/configure-hypervisor.sh
@@ -203,12 +203,17 @@ main()
     # FIXME: why is this disabled?
     qemu_options+=(--disable-linux-aio)
 
+    # In "passthrough" security mode
+    # (-fsdev "...,security_model=passthrough,..."), qemu uses a helper
+    # application called virtfs-proxy-helper(1) to make certain 9p
+    # operations safer. We don't need that, so disable it (and it's
+    # dependencies).
+    qemu_options+=(--disable-virtfs)
+    qemu_options+=(--disable-attr)
+    qemu_options+=(--disable-cap-ng)
+
     #---------------------------------------------------------------------
     # Enabled options
-
-    # Required for 9p support (used to mount the workload into the VM)
-    qemu_options+=(--enable-attr)
-    qemu_options+=(--enable-cap-ng)
 
     # Enable kernel Virtual Machine support.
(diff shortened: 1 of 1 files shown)


jodh-intel:
There are a couple of `FIXME`'s in the script - @anthonyzxu - could you give any advice?

devimc:
once ```FIXMEs``` are resolved

lgtm

jodh-intel:
Replaced by https://github.com/clearcontainers/packaging/pull/45.

Issue #347, opened 2017-08-09
CI: missing automation for execution/report metrics


MarioCarrilloA:
Missing a script that executes metrics tests and sends a report with the results by email.

PR #349, opened 2017-08-09
CI: Add script to execute/report metrics

MarioCarrilloA:
This script execute metrics and uses emailreport
tool in order to send results by email.

Fixes: #347

### DIFF
--- a/.ci/run_metrics_ci.sh
+++ b/.ci/run_metrics_ci.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+CURRENTDIR=$(dirname "$(readlink -f "$0")")
+REPORT_CMDS=("checkmetrics" "emailreport")
+
+# Verify/install report tools. These tools will
+# parse/send the results from metrics scripts execution.
+for cmd in "${REPORT_CMDS[@]}"; do
+	if ! command -v "$cmd" > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


.ci/run_metrics_ci.sh
```diff
@@ -0,0 +1,41 @@
...
+		make
+		make install
+		popd
+	fi
+done
+
+# Execute metrics scripts and report the results
+# by email.
+pushd "$CURRENTDIR/../metrics"
+	bash "run_all_metrics.sh"
```
grahamwhaley: question: I think if one of the metrics fails then the run_all_metrics.sh will fail, and this script will quit here maybe - can you confirm/nack that for me, and maybe add a comment here to make that clear.
If this script is invoked from localCI [run] action then I think we need to return the correct error code so localCI can then post the ack/nack text back to github - is that right?

.ci/run_metrics_ci.sh
```diff
@@ -0,0 +1,41 @@
...
+	emailreport
+	rm -rf "results"
```
grahamwhaley: Ah, yes, we do need to remove the results so we don't just collect up results on top of each other... mid-term we should also consider if we would want the results archived (moved) off somewhere rather than just deleting them. For now, I think this is fine.

.ci/run_metrics_ci.sh
MarioCarrilloA: Hi @grahamwhaley yeah you are right, if a test described in run_all_metrics.sh fails then the script will quit, and it will shows an error similar to this: (I put a fake command in order to test)

`run_all_metrics.sh: line 21: fake-command: command not found`

grahamwhaley: right. So we need to fix that, agreed? Maybe a simple `|| true` type fix will be easier than messing with `-e`. Over to you...

MarioCarrilloA: well, I think the information that `-e` gives is a good way in order to know which line fails (more direct) :)

grahamwhaley: but, with `-e`, will we not then quit the script, and not do the `emailreport`? If a test fails, any test, we still want to run all the other tests and do the email report, and note the test(s) that failed as 'FAIL' for instance.

MarioCarrilloA: perfect, I will fix it :)

MarioCarrilloA:
New version sent :)

grahamwhaley:
Thanks for the changes. We know we will have modifications coming soon to this script, but still:
lgtm

PR #353, opened 2017-08-09
.ci: Install specific vmlinux and vmlinuz versions for Semaphore CI

### LINKED ISSUE #350: Install vmlinux and vmlinuz v4.5-50 from the tests repository
We want to be able to install vmlinuz (our specific 4.5-50 version) from the tests repository since it is not provided on any server. Because we don't want to have to different ways to proceed, and because we are trying to simplify the CI scripts, let's add vmlinux too.

sboeuf:
We know that we have a limitation with Semaphore, we have to stick
to one specific kernel version. Because we are trying to move to
the "pc" machine type, we need a compressed kernel, but it is not
provided by the Clear Linux server for this specific version.
Indeed, the vmlinuz has been added recently to our packaging, and
for that reason, it does not exist for 4.5-50 version.

Both vmlinux and vmlinuz have been re-generated from sources and
since we know that we won't change those binaries, there is no
point in trying to download and extract them from Clear Linux
server. This saves some time for our CI.

This patch first adds both kernel binaries to the repository, and
then it uses them from the CI scripts so that we copy and link
them as the expected vmlinux.container and vmlinuz.container.

Fixes #350


devimc:
Binary files!, ehmm really? mmm how about you include the ```.config``` files? just in case we want to see how the kernel was compiled

amshinde:
@sboeuf Can we have the binary files hosted somewhere?

sboeuf:
@devimc we already have this information from koji here: http://kojiclear.jf.intel.com/cgit/packages/linux-container/tag/?h=linux-container-4.5-50

@amshinde Sure that would be better, but where ?

sboeuf:
After discussions, the rpm including vmlinux and vmlinuz will be available here: https://download.clearlinux.org/demos/clear-containers/linux-container-4.5-50.x86_64.rpm

PR #356, opened 2017-08-10
.ci: Fix Semaphore CI hangs

### LINKED ISSUE #326: crio.bats test "ctr device add" hangs under SemaphoreCI
We saw this on PR https://github.com/clearcontainers/runtime/pull/369 where it took 3-4 rebuilds to get that test to run without hanging and timing out the build.

I ssh'd to Semaphore and ran that test multiple times with no problems seem.

sboeuf:
Recently, we have seen a lot of hangs during our CI. This is due to
the low memory available inside the Semaphore box, after we run our
setup. That's the reason why this patch drop all the caches at the
end of the setup, freeing a large amount of memory. This allows our
tests to run properly with the right amount of memory.

For information, I have found that after the setup, the free memory
was around 800MiB out of the 4GiB available on the box. Notice that
when the box is started, there is 1.5GiB already used, meaning that
we really have 2.5GiB available.

Fixes #326

### DIFF
--- a/.ci/setup.sh
+++ b/.ci/setup.sh
@@ -47,3 +47,7 @@ bash -f ${cidir}/install_cni_plugins.sh
 
 echo "Install CRI-O"
 bash -f ${cidir}/install_crio.sh
+
+echo "Drop caches"
+sync
+sudo -E PATH=$PATH bash -c "echo 3 > /proc/sys/vm/drop_caches"


sboeuf:
@grahamwhaley @jodh-intel please review and merge this asap since it will unblock all our waiting PRs. I have re-run semaphore 5 times in a row with this patch and it didn't fail.

grahamwhaley:
If this fixes things, good find. It's a bit of an oddity as the kernel should reclaim/dump caches as and when needed when memory gets low, but if this makes things work then:
lgtm

PR #363, opened 2017-08-10
 integration/docker: pull alpine image before use it

### LINKED ISSUE #359: integration/docker: download alpine image
download alpine image in main_test.go to avoid timeouts in user_test.go

devimc:
pull alpine image before use it to avoid timeouts when
that image is used

fixes #359

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -24,6 +24,9 @@ const (
 
 	// Image used to run containers
 	Image = "busybox"
+
+	// AlpineImage is the alpine image
+	AlpineImage = "alpine"
 )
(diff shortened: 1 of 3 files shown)


docker.go
```diff
@@ -24,6 +24,9 @@ const (
 
 	// Image used to run containers
 	Image = "busybox"
+
+	// AlpineImage is the apline image
```
chavafg: typo s/apline/alpine/

devimc: fixed, thanks

PR #365, opened 2017-08-10
CI: update CI scripts to use new qemu-cc with Q35 support

chavafg:
This change installs the qemu-cc-system-x86_64 binary instead
of the qemu-lite-system-x86_64 binary and configures the runtime
to use the new binary.

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -41,6 +41,9 @@ export PKGLIBEXECDIR=/usr/libexec/clear-containers
 # For the pause bundle
 export LOCALSTATEDIR=/var
 
+# Qemu Binary with Q35 support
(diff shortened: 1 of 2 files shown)


.ci/install_runtime.sh
```diff
@@ -41,6 +41,9 @@ export PKGLIBEXECDIR=/usr/libexec/clear-containers
 # For the pause bundle
 export LOCALSTATEDIR=/var
 
+# Qemu Binary with Q35 support
+export QEMUCMD=qemu-cc-system-x86_64
```
gorozco1: where is this used?

chavafg: this will be used when building the `cc-runtime`
https://github.com/clearcontainers/runtime/blob/master/Makefile#L83

PR #379, opened 2017-08-14
tests: Add docker build test

### LINKED ISSUE #378: docker build test
add docker integration build test

GabyCT:
Add integration docker build test

Fixes #378

### DIFF
--- a/integration/docker/build_test.go
+++ b/integration/docker/build_test.go
@@ -1,0 +1,61 @@
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package docker
+
+import (
+	. "github.com/clearcontainers/tests"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"os"
+	"path/filepath"
+)
+
+const dockerFile = "src/github.com/clearcontainers/tests/Dockerfiles/BuildTest/."
+
+var _ = Describe("build", func() {
+	var (
+		args      []string
+		id        string
+		imageName string = "test"
+		stdout    string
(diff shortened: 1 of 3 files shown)


Dockerfiles/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2017 Intel Corporation
```
chavafg: should this file be renamed to `Dockerfile.buildtest` or something similar? 
I suspect we will need to add more Dockerfiles in this directory.

integration/docker/build_test.go
```diff
@@ -0,0 +1,57 @@
...
+		runDockerCommand(0, args...)
+		Expect(ContainerExists(id)).NotTo(BeTrue())
+	})
+
+	Describe("build with docker", func() {
+		Context("docker build env vars", func() {
+			It("should display env vars", func() {
+				gopath := os.Getenv("GOPATH")
+				entirePath := filepath.Join(gopath, dockerFile)
+				args = []string{"build", "-t", name, entirePath}
```
chavafg: if the Dockerfile name is changed,  we should add `-f ` flag to specify the path of the dockerfile: E.g. `-f Dockerfile.buildtest`

integration/docker/build_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"path/filepath"
+)
+
+const dockerFile = "src/github.com/clearcontainers/tests/Dockerfiles/."
+
+var _ = Describe("build", func() {
+	var (
+		args []string
+		id   string
+		name string = "test"
```
chavafg: maybe name this variable `image_name`

Dockerfiles/Dockerfile
grahamwhaley: I think +1 for that - there are dockerfiles we should add for the metrics tests for instance, and having unique names would be good (sure, that means we would then need a `docker build -f xxx` than defaulting to the dockerfile default name - but it is either that or we have a separate directory for each dockerfile.

GabyCT:
@chavafg changes applied

devimc:
lgtm

chavafg:
lgtm

grahamwhaley:
Hi,
 sorry, I'm a bit late to the party here, but seeing #405 caught my eye and made me think....
 afaik, a `docker build` does not invoke or involve the runtime at all, does it? In which case, maybe we should consider if we need this (and any other tests that do not actually involve or invoke or can be affected by) the runtime.
 I can see that maybe it gives us a little more confidence in our docker install and CI infra, but on the downside it will cost us some space, time and resource and will not actually be testing our code base.

 Thoughts @chavafg @GabyCT ?

PR #381, opened 2017-08-14
Port network metrics

bwarden:
I've moved the remaining scripts back out of the wip subdir, added results reporting to each one, and fixed a few bugs I've spotted, particularly with respect to results gathering. Finally, I've added the remaining tests to the run_all_metrics.sh script.

Some tests, e.g. network-metrics-iperf3.sh, report multiple types of data. At this point, I've reported them all into the same results file, using the Args field to discriminate between them. If it's preferred, I could direct each data type to a different file. Similarly, I've combined the results from the network-metrics-memory* tests (PSS, RSS) into a single results file, but I could break that up if preferred.


metrics/network/network-metrics-iperf3.sh
```diff
@@ -75,10 +94,29 @@ function iperf3_bidirectional_bandwidth_client_server {
...
+	local total_bidirectional_server_bandwidth=${server_results[6]}
+	local total_bidirectional_server_bandwidth_units=${server_results[7]}
+	echo "Network bidirectional bandwidth (client to server) is :" \
+		"$total_bidirectional_client_bandwidth" \
+		"$total_bidirectional_client_bandwidth_units"
+	echo "Network bidirectional bandwidth (server to client) is :" \
+		"$total_bidirectional_server_bandwidth" \
+		"$total_bidirectional_server_bandwidth_units"
+
+	save_results "${test_name}" "network bidir bw client to server" \
```
grahamwhaley: Ah, and here I see what you did with the multi-save of results into a single file... I think that is OK, but you are 'pushing the boundaries' here of what we have in place, so, some notes:
- use of the `Args` field is OK. That was not its original intention (it was to store the args the test had used...), but, it is not really being used or useful for that right now - so, I'd call that acceptable...
- placing multiple mixed results in a single results file will almost definitely break https://github.com/clearcontainers/tests/tree/master/cmd/checkmetrics parsing of those files... I think we can do an interim fix (filter the CSV files with `grep` for instance right now to get the results we need to check), and then add `Args` filtering to `checkmetrics`.

So, not the end of the world, and a pragmatic step forwards in holding mixed and varied results.

grahamwhaley:
going to mark DNM until Mario has akced he's had a look

grahamwhaley:
Oh, and I'll note, I'm just checking that those 'warning, suffers from Issue nnn' warnings are still valid, since they stem from CC2.x, and this merge is CC3.x...

metrics/network/network-nginx-ab-benchmark.sh
```diff
@@ -69,10 +69,16 @@ function nginx_ab_networking {
 	sleep "$sleep_secs"
 	ab -n ${requests} -c ${concurrency} http://${url}/ > "$total_requests"
 
-	result=$(cat "$total_requests" | grep "Requests per second" | cut -d ':' -f2 | sed 's/ //g' | cut -d '[' -f1)
+	cp -p $total_requests result.test
+
+	local result=$(grep "^Requests per second" $total_requests)
+	[[ $result =~ :[[:blank:]]*([[:digit:]]+(\.[[:digit:]]*)?) ]] && rps=${BASH_REMATCH[1]}
```
bwarden: I avoided leaning toothpicks :) I considered pasting sample data in a comment above, but realized I'd probably have to sanitize that since it's actual benchmarking results.

metrics/network/network-metrics-iperf3.sh
bwarden: If it makes more sense to put each result in an independent file, that's fine, and pretty easy to accomplish. I'd just take the unique info out of Args and append it to Test Name. Let me know if you'd prefer that, or adding another field, etc...

grahamwhaley: I've mulled over this a bit. I'm going to make the call to keep the data together for the minute in a single file. Now that you extract more of the results from some of the tests, I think if we put those into separate files then there is some risk in the future that we'd end up with a set of files and no proof that the data came from the same test run (well, OK, maybe timestamps and commit IDs could give a heavy clue... still...). So, let's leave as is, and then we will fix up `checkmetrics` and related other scripts to pull out the data they are looking for.

grahamwhaley:
@MarioCarrilloA - I'm going to merge this - take note of the 'multiple results in a single file'.

PR #385, opened 2017-08-15
functional: fix container.Exist method

### LINKED ISSUE #384: functional: fix container.Exist method
container.Exist method is using container.isVMRunning method but it is not implemented https://github.com/clearcontainers/tests/blob/master/container.go#L306, therefore we have to use the function implemented in vm.go https://github.com/clearcontainers/tests/blob/master/vm.go#L31 to check if the VM is still running

devimc:
this patch uses the function IsVMRunning that is defined in vm.go
to detect if a VM is running and remove the method container.isVmRunning
hece it is no more needed

fixes #384

### DIFF
--- a/container.go
+++ b/container.go
@@ -269,7 +269,7 @@ func (c *Container) Cleanup() error {
 // - the VM is running (qemu)
 // else false is returned
 func (c *Container) Exist() bool {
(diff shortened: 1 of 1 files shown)


GabyCT:
lgtm

PR #393, opened 2017-08-15
localCI: use the proper way to start go routines in a loop

### LINKED ISSUE #389: localCI: using goroutines on loop iterator variables
fix ```main.loop``` function because we are not running the routines correctly 
https://github.com/golang/go/wiki/CommonMistakes

devimc:
according with https://github.com/golang/go/wiki/CommonMistakes
the proper way to start go routines in a loop is:

```
for val := range values {
	go func(val interface{}) {
		fmt.Println(val)
	}(val)
}
```

this patch fixes the main.loop method to avoid issues starting
go routines

fixes #389

### DIFF
--- a/cmd/localCI/main.go
+++ b/cmd/localCI/main.go
@@ -41,7 +41,10 @@ func loop(repos []Repo) error {
 		}
 
 		wg.Add(1)
-		go r.loop()
+		go func(repo Repo) {
+			repo.loop()
+			wg.Done()
+		}(r)
 	}
 
 	wg.Wait()


cmd/localCI/main.go
```diff
@@ -41,7 +41,10 @@ func loop(repos []Repo) error {
 		}
 
 		wg.Add(1)
-		go r.loop()
+		go func(repo Repo) {
+			repo.loop()
+			wg.Done()
+		}(r)
```
amshinde: Nice!

amshinde:
lgtm

cmd/localCI/main.go
```diff
@@ -41,7 +41,10 @@ func loop(repos []Repo) error {
 		}
 
 		wg.Add(1)
-		go r.loop()
+		go func(repo Repo) {
+			repo.loop()
+			wg.Done()
```
grahamwhaley: was the `wg.done()` missing before as well? Just asking :-)

grahamwhaley:
lgtm

grahamwhaley:
Hit the travis ref issue - rebase/push to see if we can clear it pls

cmd/localCI/main.go
devimc: yees  :(

Issue #394, opened 2017-08-15
localCI: error: “invalid repository name: …../….”


MarioCarrilloA:
fetchtool expects 2 argument after split LOCALCI_REPO_SLUG with “/” as delimiter
https://github.com/clearcontainers/tests/blob/master/cmd/fetchbranches/localCI.go#L53

however localCI exports LOCALCI_REPO_SLUG with extra information (github domain)
https://github.com/clearcontainers/tests/blob/master/cmd/localCI/github.go#L88

PR #395, opened 2017-08-15
localCI: Fix LOCALCI_REPO_SLUG var format

MarioCarrilloA:
Fetchtool uses LOCALCI_REPO_SLUG var, this var is
set by localCI, then fetchtool expects 2 arguments
after split LOCALCI_REPO_SLUG with “/” as delimiter,
however localCI exports this var with github domain
as extra information.
This commit fixes the correct format var for fetchtool.

Fixes: #394

### DIFF
--- a/cmd/localCI/cvr.go
+++ b/cmd/localCI/cvr.go
@@ -103,6 +103,9 @@ type CVR interface {
 	// getProjectSlug returns the domain, owner and repo name separated by '/'
 	getProjectSlug() string
 
+	// getRepoSlug returns the owner and the repo name separated by '/'
+	getRepoSlug() string
+
 	// getLatestPullRequestComment returns the latest comment of a specific
 	// user in the specific pr. If user is an empty string then any user
 	// could be the author of the latest comment. If body is an empty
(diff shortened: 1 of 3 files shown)


grahamwhaley:
I think you'll also need to update https://github.com/clearcontainers/tests/blob/master/cmd/localCI/README.rst with the change to `LOCALCI_REPO_SLUG` format.
Adding @devimc and @GabyCT as reviewers, as I'd like to check we are all in sync here on what the format should be, why they are currently disparate between localCI and fetchbranches, and if this is the right fix.

devimc:
could you create a new method ```getRepoSlug``` to return ```owner/repo```  and use it in ```repo.go``` https://github.com/clearcontainers/tests/blob/master/cmd/localCI/repo.go#L213 ?
because ```getProjectSlug``` is used to generate a language environment

devimc:
@grahamwhaley yes this is a bug in localCI, update the README is not needed

MarioCarrilloA:
No problem, I will work on that function

devimc:
lgtm

PR #410, opened 2017-08-17
Add density metrics

grahamwhaley:
Add the PSS memory footprint test to the metrics density subdir.
Derived from https://github.com/01org/cc-oci-runtime/blob/master/tests/metrics/density/docker_memory_usage.sh.in

### DIFF
--- a/metrics/density/docker_memory_usage.sh
+++ b/metrics/density/docker_memory_usage.sh
@@ -1,0 +1,159 @@
+#!/bin/bash
+
+#  Copyright (C) 2017 Intel Corporation
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+#  Description of the test:
+#  This test launches a number of containers in idle mode,
+#  It will then sleep for a configurable period of time to allow
+#  any memory optimisations to 'settle, and then checks the
+#  amount of memory used by all the containers to come up with
+#  an average (using the PSS measurements)
+#  This test uses smem tool to get the memory used.
+
+set -e
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+# Busybox image: Choose a small workload image, this is
+# in order to measure the runtime footprint, not the workload
+# footprint.
+IMAGE='busybox'
+
+CMD='sh'
+NUM_CONTAINERS="$1"
+WAIT_TIME="$2"
+TEST_NAME="memory footprint"
+TEST_ARGS="workload=${IMAGE} units=kb"
+# Set QEMU_PATH unless it's already set
+QEMU_PATH=${QEMU_PATH:-$(get_qemu_path)}
+SMEM_BIN="smem"
+KSM_ENABLE_FILE="/sys/kernel/mm/ksm/run"
+
+# Show help about this script
+function help(){
+cat << EOF
+Usage: $0 <count> <wait_time>
+   Description:
+        <count>      : Number of Clear Containers to run.
+        <wait_time>  : Time in seconds to wait before taking
+                       metrics.
+EOF
+}
+
+# See if KSM is enabled.
+# If so, ammend the test name to reflect that
+function check_for_ksm(){
+	if [ ! -f ${KSM_ENABLE_FILE} ]; then
+		return
+	fi
+
+	ksm_on=$(< ${KSM_ENABLE_FILE})
+
+	if [ $ksm_on == "1" ]; then
+		TEST_NAME="${TEST_NAME} ksm"
+	fi
+}
+
+# This function measures the PSS average
+# memory of a process.
+function get_pss_memory(){
+	ps="$1"
+	mem_amount=0
+	count=0
+	avg=0
+
+	data=$("$SMEM_BIN" --no-header -P "^$ps" -c "pss")
+	for i in $data;do
+		if (( i > 0 ));then
+			mem_amount=$(( i + mem_amount ))
+			(( count++ ))
+		fi
+	done
+
+	if (( $count > 0 ));then
+		avg=$(bc -l <<< "scale=2; $mem_amount / $count")
+	fi
+
+	echo "$avg"
+}
+
+# It calculates the memory footprint
+# of a CC.
+function get_docker_memory_usage(){
+	qemu_mem=0
+	cc_shim_mem=0
+	proxy_mem=0
+	cc_proxy_mem=0
+
+	containers=()
+
+	for ((i=1; i<= NUM_CONTAINERS; i++)); do
+		containers+=($(random_name))
+		${DOCKER_EXE} run --runtime "$RUNTIME" --name ${containers[-1]} -tid $IMAGE $CMD
+	done
+
+	# If KSM is enabled, then you normally want to sleep long enough to
+	# let it do its work and for the numbers to 'settle'.
+	sleep "$WAIT_TIME"
+
+	# Get PSS memory of CC components.
+	# And check that the smem search has found the process - we get a "0"
+	#  back if that procedure fails (such as if a process has changed its name
+	#  or is not running when expected to be so)
+	# As an added bonus - this script must be run as root (or at least as
+	#  a user with enough rights to allow smem to read the smap stats for
+	#  the docker owned processes). Now if you do not have enough rights
+	#  the smem failure to read the stats will also be trapped.
+	qemu_mem="$(get_pss_memory "$QEMU_PATH")"
+	if [ "$qemu_mem" == "0" ]; then
+		die "Failed to find PSS for $QEMU_PATH"
+	fi
+
+	cc_shim_mem="$(get_pss_memory "$CC_SHIM_PATH")"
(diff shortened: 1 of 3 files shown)


metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,161 @@
...
+
+set -e
+
+SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
+source "${SCRIPT_PATH}/../lib/common.bash"
+
+# Busybox image: Choose a small workload image, this is
+# in order to measure the runtime footprint, not the workload
+# footprint.
+IMAGE='busybox'
```
jodh-intel: How about adding this to https://github.com/clearcontainers/tests/blob/master/test-versions.txt ?

Follow-on PRs could then update `bundle.go` and `docker.go` to also read the default test image from that file.

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,161 @@
...
+
+	save_results "$TEST_NAME" "$TEST_ARGS" "$cc_mem_usage" "KB"
+	docker rm -f ${containers[@]}
+}
+
+# Verify enough arguments
+if [ $# -ne 2 ];then
+	echo >&2 "error: No enough arguments"
+	help
+	exit 1;
```
jodh-intel: Semi-colon not needed.

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,161 @@
...
+TEST_NAME="memory footprint"
+TEST_ARGS="workload=${IMAGE} units=kb"
+# Set QEMU_PATH unless it's already set
+QEMU_PATH=${QEMU_PATH:-$(get_qemu_path)}
+SMEM_BIN="smem"
+KSM_ENABLE_FILE="/sys/kernel/mm/ksm/run"
+
+# Show help about this script
+function help(){
+usage=$(cat << EOF
```
jodh-intel: I don't think the `usage` variable is needed - you could just do:

```
function help() {
cat <<EOT
    ... blah ...
EOF
}

File: metrics/density/docker_memory_usage.sh (hunk 1 above)
Comment (grahamwhaley) [inline metrics/density/docker_memory_usage.sh]: If we had a stable base image we used for all tests and metrics, then I think that would make sense - but, we have a number of different images we use for different purposes (like alpine here: https://github.com/clearcontainers/tests/blob/master/docker.go#L30 as well as busybox and some network specific ones here: https://github.com/clearcontainers/tests/blob/master/metrics/network/network-metrics-iperf3.sh#L28 and IO one here: https://github.com/clearcontainers/tests/blob/master/metrics/storage/fio_job.sh#L42)

So, I don't think it's quite as simple as defining one image name in the versions.txt and be done with it. Thus, I'm going to propose we leave alone, if you concur?

File: metrics/density/docker_memory_usage.sh (hunk 3 above)
Comment (grahamwhaley) [inline metrics/density/docker_memory_usage.sh]: Indeed. done.

File: metrics/density/docker_memory_usage.sh (hunk 2 above)
Comment (grahamwhaley) [inline metrics/density/docker_memory_usage.sh]: yep. done

File: metrics/density/docker_memory_usage.sh (hunk 1 above)
Comment (jodh-intel) [inline metrics/density/docker_memory_usage.sh]: It's just an idea, so I don't think it's a blocker. But I personally dislike the fact that the "normal" / default image (`busybox`) is defined in multiple places and this PR adds another place ;)

Comment (grahamwhaley) [inline metrics/density/docker_memory_usage.sh]: OK. Let me see what it looks like with adding a 'default_docker_image' in there, and I'll probably add it to the PR

Comment (grahamwhaley) [inline metrics/density/docker_memory_usage.sh]: Right, I had a stab at it, and it looks alright, but it dropped me into a little bit of bash source path hell. I'm going to push a WIP into another branch and ask @bwarden if he has any ideas on a nice way to fix that... and will them come back here.

Comment (grahamwhaley) [inline metrics/density/docker_memory_usage.sh]: OK, that will be a bit more of a re-write than I want to do right now, and it should not really be in the middle of this PR. I've opened an Issue https://github.com/clearcontainers/tests/issues/411. Let me push those other updates now...

Comment (jodh-intel) [inline metrics/density/docker_memory_usage.sh]: Great - wfm.

**Comment by grahamwhaley** [comment]
re-pushed @jodh-intel . Changes applied

**Comment by chavafg** [comment]
can you rebase? seems like semaphore got an issue that @devimc fixed yesterday.

```
 Test Panicked
    runtime error: invalid memory address or nil pointer dereference
    /usr/local/golang/1.8.3/go/src/runtime/panic.go:489

    Full Stack Trace
    	/usr/local/golang/1.8.3/go/src/runtime/panic.go:489 +0x2cf
    github.com/clearcontainers/tests.IsVMRunning.func1(0xc4203cd690, 0xa, 0x0, 0x0, 0x95c8c0, 0xc4203c0390, 0xc4200184f0, 0x0)
    	/home/runner/workspace/src/github.com/clearcontainers/tests/vm.go:33 +0x4c
    path/filepath.walk(0x7ce010, 0x5, 0x962640, 0xc4201eeea0, 0xc420410900, 0x0, 0x20)
    	/usr/local/golang/1.8.3/go/src/path/filepath/path.go:372 +0x2fe
```

File: metrics/density/docker_memory_usage.sh [hunk 4]
```diff
@@ -0,0 +1,159 @@
...
+}
+
+# See if KSM is enabled.
+# If so, ammend the test name to reflect that
+function check_for_ksm(){
+	if [ ! -f ${KSM_ENABLE_FILE} ]; then
+		return
+	fi
+
+	ksm_on=$(cat ${KSM_ENABLE_FILE})
```
bwarden: $(< ${KSM_ENABLE_FILE}) to handle completely within bash

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+# memory of a process.
+function get_pss_memory(){
+	ps="$1"
+	mem_amount=0
+	count=0
+	avg=0
+
+	data=$("$SMEM_BIN" --no-header -P "^$ps" -c "pss")
+	for i in $data;do
+		if (( $i > 0 ));then
```
bwarden: "$" sigil is redundant within double-parens

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+			mem_amount=$(( $i + $mem_amount ))
+			count=$(( $count + 1 ))
```
bwarden: Entire line could be replaced with `((count++))`

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+function get_pss_memory(){
+	ps="$1"
+	mem_amount=0
+	count=0
+	avg=0
+
+	data=$("$SMEM_BIN" --no-header -P "^$ps" -c "pss")
+	for i in $data;do
+		if (( $i > 0 ));then
+			mem_amount=$(( $i + $mem_amount ))
```
bwarden: Same about sigils

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+			count=$(( $count + 1 ))
+		fi
+	done
+
+	if (( $count > 0 ));then
+		avg=$(echo "scale=2; $mem_amount / $count" | bc -l)
```
bwarden: Could be written as:
`avg=$(bc -l <<< "scale=2; $mem_amount / $count")`

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+# of a CC.
+function get_docker_memory_usage(){
+	qemu_mem=0
+	cc_shim_mem=0
+	proxy_mem=0
+	cc_proxy_mem=0
+
+	containers=()
+
+	for i in $(seq 1 "$NUM_CONTAINERS"); do
```
bwarden: Could do a classic for-loop:
`for ((i=1; i<=NUM_CONTAINERS; i++)); do`

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+	if [ "$cc_shim_mem" == "0" ]; then
+		die "Failed to find PSS for $CC_SHIM_PATH"
+	fi
+
+	proxy_mem="$(get_pss_memory "$CC_PROXY_PATH")"
+	if [ "$proxy_mem" == "0" ]; then
+		die "Failed to find PSS for $CC_PROXY_PATH"
+	fi
+
+	cc_proxy_mem="$(echo "scale=2; $proxy_mem / $NUM_CONTAINERS" | bc -l)"
```
bwarden: <<<

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+	cc_mem_usage=$(echo "scale=2; $qemu_mem + $cc_shim_mem + $cc_proxy_mem" | bc -l)
```
bwarden: <<<

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+
+	save_results "$TEST_NAME" "$TEST_ARGS" "$cc_mem_usage" "KB"
+	docker rm -f ${containers[@]}
+}
+
+# Verify enough arguments
+if [ $# -ne 2 ];then
```
bwarden: Could use bash's built-in `[[` instead.

metrics/density/docker_memory_usage.sh
```diff
@@ -0,0 +1,159 @@
...
+	echo >&2 "error: No enough arguments"
```
bwarden: Should we print what the arguments are?

metrics/density/docker_memory_usage.sh
grahamwhaley: I'm not always convinced these 'fancy new bashisms' are easier to read than plain old *nix style ;-)
(yes, note, I'm old school ...)

grahamwhaley:
@chavafg rebased
@bwarden changes applied - thanks for the 'modernisation' feedback ;-)

bwarden:
lgtm

chavafg:
lgtm

PR #413, opened 2017-08-18
ci: Increase test timeout from 5 to 15 seconds

### LINKED ISSUE #412: Increase tests timeout from 5 to 15 seconds
The 5 seconds timeout is appropriate if tests are run on baremetal, but according to recent changes on QEMU parameters, if running in a nested environment, the tests have been slowed down by disabling fast MMIO access previously performed by virtio 1.0.

And given the fact that our CI runs on a VMM (Semaphore CI), we cannot afford this timeout of 5 seconds. Moving this timeout to 15 seconds makes the tests passing even when run inside a VM.

sboeuf:
The 5 seconds timeout is appropriate if the tests run on baremetal,
but according to recent changes on QEMU parameters if running in a
nested environment, the tests have been slowed down by disabling fast
MMIO access previously performed by virtio 1.0.

And given the fact that our CI runs on a VMM, we cannot afford this
timeout of 5 seconds. Moving this timeout to 15 seconds makes the
tests passing even when run inside a VM.

Fixes #412

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@
 CC_RUNTIME ?= cc-runtime
 
 # The time limit in seconds for each test
-TIMEOUT ?= 5
+TIMEOUT ?= 15
 
 CRIO_REPO_PATH="${GOPATH}/src/github.com/kubernetes-incubator/cri-o"
 crio:


grahamwhaley:
This makes sense to me.
afaik, the timeout is only for the (hopefully very rare) error case, so we don't really care how big it is, as 99% of the time the tests should work and not hit the timeout. As long as the timeout catches the hung up cases, that should be fine.
I've added @chavafg to the review list for his thoughts.

lgtm

sboeuf:
Thanks @grahamwhaley.
@chavafg if you think the change is fine, go ahead and merge it.

chavafg:
lgtm, anyway PnP will notice if our times got too much higher, right @grahamwhaley?

grahamwhaley:
heh, right @chavafg  Yes, I don't see this timeout being used to capture 'you made it all very slow' really :-)

sboeuf:
@chavafg can you merge this please ?

Issue #414, opened 2017-08-18
metrics: CI: add metrics to default metrics ci run list


grahamwhaley:
Now we have the metrics CI script:
https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh
we need to expand the list of metrics it runs to cover the core data from the core components, including:
- speed
- size
- storage
- networking

PR #415, opened 2017-08-18
[RFC] metrics: ci: Add more tests to the metrics ci script

grahamwhaley:
We don't run all the metrics tests for the metrics CI, but we do
want to run a 'good subset' that covers all key data from all key
areas of interest.
Thus, add memory footprint, network and storage tests to the default
set of metrics we run in the ci.

Fixes: #414

### DIFF
--- a/.ci/run_metrics_ci.sh
+++ b/.ci/run_metrics_ci.sh
@@ -17,6 +17,8 @@
 CURRENTDIR=$(dirname "$(readlink -f "$0")")
 REPORT_CMDS=("checkmetrics" "emailreport")
 
+KSM_ENABLE_FILE="/sys/kernel/mm/ksm/run"
+
 # Verify/install report tools. These tools will
 # parse/send the results from metrics scripts execution.
 for cmd in "${REPORT_CMDS[@]}"; do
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@MarioCarrilloA can you check if this matches your thoughts, and also if it works under you CI setup.
Note, the density tests won't work until #410 lands - you could hand-copy over or comment out to test etc.
/cc @bwarden so you can see where this is heading, and cast your script eyes over it

/cc @mcastelino @amshinde @egernst - I cherry picked out some of the network tests to use for our baseline CI - pretty much looking at latency, bandwidth, jitter for TCP and some UDP. We can tweak this as we go along of course, but I thought I'd give you a pre heads-up.

Note to all - we do not want to run all the metrics for the CI - we want to focus on things that ideally don't take too long (but storage and some network tests will defeat that goal just because that is how they have go be) but do cover our basic requirements.

MarioCarrilloA:
Hi @grahamwhaley, yes let's take a look :)

grahamwhaley:
Oh, I should note @MarioCarrilloA - some of these tests (particularly the network ones iirc) will generate multiple results (such as read and write bandwidth or bandwidth and jitter) mixed into a single csv file. Today that will not work with `checkmetrics`, but we have two ways to fix that:
- short term, I think we put some `grep xxx.csv > yyy.csv` into this script as a temporary workaround - so we 'cherry pick' out the results into separate file so the `checkmetrics` toml file can be pointed at them
- longer term, I'd like to add the ability to `checkmetrics` so that it can filter the results out of the .csv files by itself. In my mind I have a [filter=[regexp]] sort of toml entry in the `checkmetrics` config files in mind.

.ci/run_metrics_ci.sh
```diff
@@ -33,9 +35,59 @@ done
 pushd "$CURRENTDIR/../metrics"
 	source "lib/common.bash"
 
+	# If KSM is available on this platform, let's run the KSM tests first
+	# and then turn it off for the rest of the tests, as KSM may introduce
+	# some extra noise in the results by stealing CPU time for instance
+	if [ -f ${KSM_ENABLE_FILE} ]; then
```
bwarden: You can use `[[` here instead of calling the `[` program

grahamwhaley: heh - fixed

.ci/run_metrics_ci.sh
```diff
@@ -33,9 +35,59 @@ done
...
+	bash network/network-metrics-iperf3.sh
+
+	# UDP bandwidths and packet loss
+	bash network/network-metrics-nuttcp.sh
+
+
+	#
+	# Run some IO tests
+	#
+	storage/fio_job.sh -b 16k -o randread -t "storage IO random read bs 16k"
```
grahamwhaley: ah, indeed! fixed

grahamwhaley:
new version pushed. DNM label dropped (although, Mario, if you can confirm the test runs in your infra that would be nice before a merge, but not critical :-) )

MarioCarrilloA:
lgtm

chavafg:
lgtm

PR #423, opened 2017-08-18
functional: do not use omega character

### LINKED ISSUE #192: functional: ginkgo/gomega: use Expect not Ω
When using ginkgo (technically, the gomega pattern matcher with ginkgo) in our tests, can we stick to using the `Expect` API call, and not the `Ω` call.
`Ω` is just going to cause us issues later on searching, parsing, or some similar general confusion, and `Expect` is functionally identical and tbh 'says what it does on the tin' more for me.
http://onsi.github.io/gomega/#making-assertions

devimc:
this patch replace omega character with Expect to make
tests easier to read and understand

fixes #192

### DIFF
--- a/functional/commands_test.go
+++ b/functional/commands_test.go
@@ -41,8 +41,8 @@ var _ = Describe("commands", func() {
 		func(cmd *Command) {
 			_, stderr, exitCode = cmd.Run()
 
-			Ω(exitCode).ShouldNot(Equal(0))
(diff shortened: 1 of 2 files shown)


grahamwhaley:
@devimc many thanks for this. Sorry I never got around to it...
lgtm

devimc:
hi @grahamwhaley  no problem 😄

Issue #438, opened 2017-08-22
metrics: No possible to get version about runtime


MarioCarrilloA:
This is due to `send_results.sh` script looks for a wrong runtime binary (path/name).

PR #439, opened 2017-08-22
metrics: Fix the path/name of the runtime used

MarioCarrilloA:
Due to the paths/names of the runtime binaries are
mixed, it is not possible to get the version of the
runtime used, this commit fixes that.

Fixes: #438

### DIFF
--- a/metrics/lib/send_results.sh
+++ b/metrics/lib/send_results.sh
@@ -32,11 +32,12 @@ source "${SCRIPT_PATH}/common.bash"
 RESULT_DIR="${SCRIPT_PATH}/../results"
 
 GROUP="PNP"
+BINDIR_LOCAL_PATH="/usr/local/bin"
 COR_DEFAULT_PATH="/usr/bin"
 COR_DEFAULT_NAME="cc-oci-runtime"
 RUNC_DEFAULT_PATH="/usr/bin"
 RUNC_DEFAULT_NAME="docker-runc"
-CC_RUNTIME_DEFAULT_PATH="/usr/local/bin"
+CC_RUNTIME_DEFAULT_PATH="/usr/bin"
 CC_RUNTIME_DEFAULT_NAME="cc-runtime"
 CC_DEFAULT_IMG_PATH="/usr/share/clear-containers"
 CONTAINERS_IMG_SYSTEM="/usr/share/clear-containers/clear-containers.img"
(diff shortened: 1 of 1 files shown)


metrics/lib/send_results.sh
```diff
@@ -122,15 +122,16 @@ function get_runtime_info()
 
 	case "$RUNTIME" in
 		cc-runtime)
-			result="$(get_runtime_version ${COR_DEFAULT_PATH} ${COR_DEFAULT_NAME})"
+			result="$(get_runtime_version ${CC_RUNTIME_DEFAULT_PATH} ${CC_RUNTIME_DEFAULT_NAME})"
 			;;
 
 		cor)
-			result="$(get_runtime_version ${RUNC_DEFAULT_PATH} ${RUNC_DEFAULT_NAME})"
+			result="$(get_runtime_version ${COR_DEFAULT_PATH} ${COR_DEFAULT_NAME})"
```
grahamwhaley: Doh! good spot :-)

metrics/lib/send_results.sh
```diff
@@ -36,7 +36,7 @@ COR_DEFAULT_PATH="/usr/bin"
 COR_DEFAULT_NAME="cc-oci-runtime"
 RUNC_DEFAULT_PATH="/usr/bin"
 RUNC_DEFAULT_NAME="docker-runc"
-CC_RUNTIME_DEFAULT_PATH="/usr/local/bin"
+CC_RUNTIME_DEFAULT_PATH="/usr/bin"
```
grahamwhaley: So, iirc, the difference between having `local` or not in the path is that between a, shall we say, development install and a production(release) install. As long as this path is working for the CI testing infra, then I'm happy with the change :-)

grahamwhaley:
@bwarden - would you like to cast your eyes over this, as a second pair of interested eyes. thx

bwarden:
lgtm

metrics/lib/send_results.sh
chavafg: What about checking first if the the `cc-runtime` binary is installed in `/usr/local/bin`, using `which` or `command`. And add an `echo: Using cc-runtime from: $INSTALLED_PATH` so that the user is aware of that. If the user has not installed the runtime by running the `make install`, then he will only have the runtime installed in `/usr/bin`

grahamwhaley: I agree we could be smarter in our search. There is a limit to how smart I think we should be though.
Given that we have two defaults in our build files afaik, then maybe we should do a dual-search here.
An, sure, printing out some diagnostics could be useful (but, otoh, we call this code a lot when running the metrics to store all the results).

(/me notes - I think longer term we should also cache the results found as I suspect doing all this searching for every line saved in the csv is probably costing us quite a bit of runtime...)
@MarioCarrilloA , do you have bandwidth to add the search for /usr/bin and /usr/local/bin?
If not, I'm happy if we open an Issue to note we can be smarter, and merge this so we can get the code up and running

MarioCarrilloA: @chavafg thanks for the observation, @grahamwhaley I can add that in this commit :)

MarioCarrilloA:
Hi @chavafg and @grahamwhaley as you can see, I added a pair of conditionals for runtime paths, I think this script should looks for runtime first in a local path, this is due to continues integration system involves to build locally continually, however the script should work using the OBS packages. I did not use `which` command because it will search from left to right in PATH env var and it should have `/usr/local/bin` first in order to verify in a local path, but PATH can/cannot have `/usr/local/bin` first, it depends of user/system, however if you have any suggestion please let me know :)

grahamwhaley:
I think searching in the local path first is the right thing to do as well.
lgtm

PR #464, opened 2017-08-24
localCI: honour whitelist

### LINKED ISSUE #463: localCI: whitelist is not honoured
localCI does not honour the whitelist

devimc:
with this patch localCI honours the whitelist of
users whose pull requests can be tested

fixes #463


grahamwhaley:
Funny, I'd read the function code before, but not realised it was not called from anywhere!
I note we also have this:
https://github.com/clearcontainers/tests/blob/master/cmd/localCI/branch.go#L56
which I believe means the whitelist does not apply to branch testing, as all branches will only contain code that has already come through a PR/whitelist/review cycle. Just noting...  so,

lgtm

PR #470, opened 2017-08-25
integration/popularImages: change aerospike test

devimc:
use asd (aerospike daemon) to verify if aerospike image
is compatible with CC 3.0


MarioCarrilloA:
lgtm

Issue #471, opened 2017-08-25
metrics: check dependencies before execute all metrics test


MarioCarrilloA:
When  `run_all_metrics.sh` is running is common that execution is interrupted for errors related
to commands/tools dependencies, could be better to ask them before execute metrics tests.

PR #472, opened 2017-08-25
metrics: Verify dependencies before execute all

MarioCarrilloA:
The "all" metrics tests execution is interrupted for
errors related to commands/tools dependencies, it is
better to ask them before execution. This might seem
redundant due to these dependencies could be requested
by each metrics test, however this allows to avoid
this kind of interruption/errors running metrics test
as a group or running an independent test.

Fixes: #471

### DIFF
--- a/metrics/run_all_metrics.sh
+++ b/metrics/run_all_metrics.sh
@@ -26,6 +26,11 @@ fi
 
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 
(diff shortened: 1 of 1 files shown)


chavafg:
lgtm

grahamwhaley:
lgtm
I'm happy to have the checks here as well as in the individual scripts themselves. Sure, we are likely to forget to keep them in sync as we add and update tests, but I'm happy to have issues known issues caught earlier.

Issue #473, opened 2017-08-25
metrics: the network-nginx metrics test does not allow to specify runtime


MarioCarrilloA:
The `network-nginx-ab-benchmark.sh` should use RUNTIME var in order to allow an execution with a different runtime.

PR #474, opened 2017-08-25
metrics: network: allow to specify another runtime

MarioCarrilloA:
This metrics test should use RUNTIME var in order to be executed
using a different runtime. This commit adds that.

Fixes: #473


chavafg:
lgtm

PR #479, opened 2017-08-28
localCI: don't re-test if pull request has not changed

### LINKED ISSUE #468: localCI: don't re-test a PR if its hash doesn't change
PRs are re-tested if its hash doesn't change

devimc:
with this patch if localCI is not able to get the information
of a specific pull request it uses the old one to avoid deleting
it from already tested revisions causing a re-test of a pull
request that has not changed

fixes #468

### DIFF
--- a/cmd/localCI/repo.go
+++ b/cmd/localCI/repo.go
@@ -260,10 +260,23 @@ func (r *Repo) loop() {
 
 	appendPullRequests := func(revisions *[]revision, prs []int) error {
 		for _, prNumber := range prs {
+			var pr revision
+			var err error
+
 			r.logger.Debugf("requesting pull request %d", prNumber)
-			pr, err := newPullRequest(prNumber, r.prConfig)
+			pr, err = newPullRequest(prNumber, r.prConfig)
 			if err != nil {
-				return fmt.Errorf("failed to get pull request '%d' %s", prNumber, err)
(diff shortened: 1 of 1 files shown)


cmd/localCI/repo.go
```diff
@@ -260,10 +260,23 @@ func (r *Repo) loop() {
 
 	appendPullRequests := func(revisions *[]revision, prs []int) error {
 		for _, prNumber := range prs {
+			var pr revision
+			var err error
+
 			r.logger.Debugf("requesting pull request %d", prNumber)
-			pr, err := newPullRequest(prNumber, r.prConfig)
```
MarioCarrilloA: just out of curiosity, why did you change short variable declaration?

devimc: ```newPullRequest``` returns an object of type ```pullRequest``` that can be assigned to an object of type ```revision``` but not vice-versa ("polymorphism"). In the line 274 I look for an object of type ```revision``` with the same ```prNumber``` and assign it to ```rev``` then in the line 279 I copy ```rev``` (type ```revision```) to ```pr``` (type ```revision```) because ```pr``` is appended to a map of ```revision```s. 
In conclusion if the type of ```pr``` was ```pullRequest``` I couldn't assign it to ```rev``` hence I couldn't copy it into the map (```revisionsTested```)

MarioCarrilloA:
lgtm

Issue #480, opened 2017-08-28
metrics: density: It is not possible to measure the density of a runc container


MarioCarrilloA:
The metrics tests just allow to measure the memory density about Clear Containers.

PR #481, opened 2017-08-28
[RFC]: metrics: density: measure runc containers

MarioCarrilloA:
This metrics test just allow measure the memory density
about Clear Containers, it fails when tries to measure
PSS memory of components that only exists during a Clear
container execution (e.g. qemu instance). This commit
allows to measure the PSS of the workload launched by a
runc container.

Fixes: #480

### DIFF
--- a/metrics/density/docker_memory_usage.sh
+++ b/metrics/density/docker_memory_usage.sh
@@ -39,8 +39,6 @@ NUM_CONTAINERS="$1"
 WAIT_TIME="$2"
 TEST_NAME="memory footprint"
 TEST_ARGS="workload=${IMAGE} units=kb"
-# Set QEMU_PATH unless it's already set
-QEMU_PATH=${QEMU_PATH:-$(get_qemu_path)}
 SMEM_BIN="smem"
 KSM_ENABLE_FILE="/sys/kernel/mm/ksm/run"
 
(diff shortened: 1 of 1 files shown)


metrics/density/docker_memory_usage.sh
```diff
@@ -111,33 +110,48 @@ function get_docker_memory_usage(){
...
-	#  the docker owned processes). Now if you do not have enough rights
-	#  the smem failure to read the stats will also be trapped.
-	qemu_mem="$(get_pss_memory "$QEMU_PATH")"
-	if [ "$qemu_mem" == "0" ]; then
-		die "Failed to find PSS for $QEMU_PATH"
-	fi
+	# Check the runtime in order in order to determine which process will
+	# be measured about PSS
+	if [ "$RUNTIME" == "runc" ]; then
+		runc_workload_mem="$(get_pss_memory "$CMD")"
```
grahamwhaley: Hi.
 Right, I think the notion is potentially right, but a few comments:
- we should add the '0 returned' check like we do below - to check that smem actually found the workload
- I'm slightly nervous that we are checking for such as common command as `sh` (the default CMD we run) - it feels like it would be very easy to have a random `sh` laying around running on the system from something else that would then get included in our results ... but, having said that, we look for `^sh`, and that I think vastly reduces the chances of us picking up a stray `/bin/sh` for instance...
- another idea is maybe we could enter the namespaces of the containers to do the smem to ensure we only pick up the CMDs we are looking for - but, I think we'd have to do that for each container separately, and that feels like a lot of effort for a small gain ;-)
- and finally, we could potentially use `docker stat --no-stream` to get the memory usage numbers directly from docker itself. Having said that, on my local test from smem I get PSS of 392K, and from docker I get a mem usage of 100 KiB - so, who to trust??

Given this is better than what we have, and I suspect in 99% of cases will give us the correct data (not include other random `sh`s) and get that data with the same method we use for CC's (`smem`), I'm probably going to be fine with this - but, @MarioCarrilloA - any thoughts on the above?

MarioCarrilloA: Hi @grahamwhaley good points, In my opinion we should take the results provided by `smem` (PSS), it seems that `docker stat` shows the information related to the used cgroup, I could not see info related to PSS in documentation
https://docs.docker.com/engine/admin/runmetrics/#memory-metrics-memorystat

And also I agree regarding to looks for `sh`, it will be very easy to "pollute" the results with more random `sh` instances surfing in the system, I will check that :)

MarioCarrilloA:
Hi @grahamwhaley , this commit includes a function that will get all instances of docker-containerd-shim launched by docker-containerd, then it will calculate the PSS average about the child process launched by each docker-containerd-shim, that in this case is a `sh` instance, these results can be affected just if there are more containers running, however we made a clean up before that

`docker-containerd
                ├── docker-containerd-shim
                                          └── workload <--1
                ├── docker-containerd-shim
                                          └── workload <--2
                ├── docker-containerd-shim
                                          └── workload <--N`

grahamwhaley:
lgtm.
Thansk for the update @MarioCarrilloA I think tracking the payload via the shim is the right thing to do :-)

Hi @sboeuf This PR is being blocked by the Jenkins CI :-) - can you let me know if that is live and needs a rebuild (looks like you aborted the build for this PR, so it is effectively 'failed'), and if not, I can always over-ride the blocker for now whilst we settle Jenkins down.

Issue #488, opened 2017-08-28
metrics: network: It is not possible to measure the CPU % of a runc container


MarioCarrilloA:
The metrics tests just allow to measure the CPU % about Clear Containers.

PR #489, opened 2017-08-28
[RFC]: metrics: network: measure CPU % of runc containers

MarioCarrilloA:
This metrics test just allow measure the cpu % about Clear Containers,
it fails when tries to measure CPU % of components that only exists in
during a Clear container execution (e.g. qemu instance).
This commit allows to measure the CPU % of a  workload launched by a
runc container.

Fixes: #488


grahamwhaley:
This:
lgtm
I will note, I've always been dubious about using the ps to measure the process % - we had a big discussion on that on the original ticket/PR iirc, and we should probably re-visit it sometime maybe.

grahamwhaley:
and, oops, that was an [RFC] - should have marked that with a DNM label, and then github should have blocked it (via pullapprove). @MarioCarrilloA if you think we need to revert that, then please **shout** :-)

Issue #490, opened 2017-08-29
stability: soak test multi container force removes


grahamwhaley:
Add a stability soak test that checks multiple parallel force removes (`docker rm -f`) work completely, and do not leave any runtime artifacts around.

PR #491, opened 2017-08-29
qa: stability: check multi launch and rm stability

grahamwhaley:
launch a number of containers, and then 'rm -f' them all in one go.
Check we have the right number of 'components' running at each
stage.

Fixes: #490

### DIFF
--- a/integration/stability/soak_parallel_rm.sh
+++ b/integration/stability/soak_parallel_rm.sh
@@ -1,0 +1,227 @@
+#!/bin/bash
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This test will run a number of parallel containers, and then try to
+# 'rm -f' them all at the same time. It will check after each run and
+# rm that we have the expected number of containers, shims, proxys,
+# qemus and runtimes active
+# The goals are two fold:
+# - spot any stuck or non-started components
+# - catch any hang ups
+
+# How many times will we run the test loop...
+ITERATIONS=5
+
+# the system 'free available' level where we stop running the tests, as otherwise
+#  the system can crawl to a halt, and/or start refusing to launch new VMs anyway
+# We choose 2G, as that is one of the default VM sizes for CC
+MEM_CUTOFF=(2*1024*1024*1024)
+
+# The default container/workload we run for testing
+# nginx has show up issues in the past, for whatever reason, so
+# let's default to that
+PAYLOAD="nginx"
+
+# do we need a command argument for this payload?
+COMMAND=""
+
+# Set the runtime if not set already
+RUNTIME="${RUNTIME:-cc-runtime}"
+
+# And set the names of the processes we look for
+QEMU_NAME="${QEMU_NAME:-qemu-lite-system-x86_64}"
+RUNTIME_NAME="${RUNTIME_NAME:-cc-runtime}"
+SHIM_NAME="${SHIM_NAME:-cc-shim}"
+PROXY_NAME="${PROXY_NAME:-cc-proxy}"
+
+# The place where virtcontainers keeps its active pod info
+# This is, I believe, ultimately what 'cc-runtime list' uses to get its info, but
+# we can also check it for sanity directly
+VC_POD_DIR="${VC_POD_DIR:-/var/lib/virtcontainers/pods}"
+
+# let's cap the test. If you want to run until you hit the memory limit
+# then just set this to a very large number
+MAX_CONTAINERS=110
+
+function count_containers() {
+	docker ps -qa | wc -l
+}
+
+function check_all_running() {
+	local goterror=0
+
+	echo "Checking ${how_many} containers have all relevant components"
+
+	# check what docker thinks
+	how_many_running=$(count_containers)
+
+	if (( ${how_many_running} != ${how_many} )); then
+		echo "Wrong number of containers running (${how_many_running} != ${how_many}) - stopping"
+		((goterror++))
+	fi
+
+	# Only check for CC components if we are using a CC runtime
+	if (( $check_cc_components )); then
+		# check we have the right number of proxy's
+		how_many_proxys=$(ps --no-header -C ${PROXY_NAME} | wc -l)
+		if (( ${how_many_running} >= 1 )); then
+			# If we have any containers, then we expect to have a single proxy
+			if (( ${how_many_proxys} != 1 )); then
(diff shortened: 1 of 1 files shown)


integration/stability/soak_parallel_rm.sh
```diff
@@ -0,0 +1,168 @@
...
+
+function kill_all_containers() {
+	present=$(docker ps -qa | wc -l)
+	if ((${present})); then
+		docker rm -f $(docker ps -qa)
+	fi
+}
+
+function init() {
+	kill_all_containers
```
MarioCarrilloA: could be good to restart docker/cc-proxy services at this point

integration/stability/soak_parallel_rm.sh
```diff
@@ -0,0 +1,168 @@
...
+# let's cap the test. If you want to run until you hit the memory limit
+# then just set this to a very large number
+MAX_CONTAINERS=110
+
+function count_containers() {
+	docker ps -qa | wc -l
+}
+
+function check_all_running() {
+	echo Checking ${how_many} containers have all relevant components
```
MarioCarrilloA: Missing quotes in this `echo`

integration/stability/soak_parallel_rm.sh
```diff
@@ -0,0 +1,168 @@
...
+		if (( ${how_much} < ${MEM_CUTOFF} )); then
+			echo "And we are out of memory on container ${how_many} (${how_much} < ${MEM_CUTOFF})"
+			return
+		fi
+	}
+	done
+}
+
+function kill_all_containers() {
+	present=$(docker ps -qa | wc -l)
```
MarioCarrilloA: Maybe `count_containers()` could be used here

integration/stability/soak_parallel_rm.sh
```diff
@@ -0,0 +1,168 @@
...
+	if (( ${how_many_running} != ${how_many} )); then
+		echo "Wrong number of containers running (${how_many_running} != ${how_many}) - stopping"
+		exit -1
+	fi
+
+	# check we have the right number of proxy's
+	how_many_proxys=$(ps --no-header -C ${PROXY_NAME} | wc -l)
+	if (( ${how_many_running} >= 1 )); then
+		# If we have any containers, then we expect to have a single proxy
+		if (( ${how_many_proxys} != 1 )); then
```
MarioCarrilloA: if this tests is designed for `runc` as a RUNTIME, could be `${how_many_proxys}` > 1, this is due to `runc` does not use cc-proxy or something like that, then `how_many_proxys` will be 0 and this an expected result

integration/stability/soak_parallel_rm.sh
grahamwhaley: strictly not needed, well not unless ${how_many} might expand into a command line argument or something ;-), but sure, good practice to have - I'll go add

integration/stability/soak_parallel_rm.sh
grahamwhaley: Indeed. Right now the primary goal is to run with CC, but if we were to support 'other' runtimes, then yes, we need to make the script more sentient.
I'm probably about to code something similar up in another script - I'll see if I get to apply that here as well.

integration/stability/soak_parallel_rm.sh
grahamwhaley: oh, good point!

integration/stability/soak_parallel_rm.sh
grahamwhaley: Hmm, yea, I guess. It almost seem backwards - make the system 'more stable' to try and 'find instabilities', but for repeatability it makes a lot of sense. Let me see if I can do something by re-using the stuff we have in the metrics lib files (would be good if we could share some of that across metrics and QA for instance).

chavafg: maybe it is time to add a `lib` directory in the root of the repository, and start populating it and using it. But this can be another PR.

grahamwhaley: An idea from @chavafg - we can also check that there are no pending docker mounts left after the test - something else we have spotted as a potential issue (https://github.com/clearcontainers/packaging/pull/110)

grahamwhaley:
I've rebase and pushed - let's see if we can get the CI to pass and ack/merge this?
We can then open an issue to add the mount-checks.
Oh, and Travis CI will fail as I will have rebased on the PR that just landed that removed the Travis config...

grahamwhaley:
@MarioCarrilloA updated with changes
@devimc added a check for 'dangling mounts'. It is a bit of a simple check - just check we have the same number of mounts in the system after we've deleted all the containers that we had before we launched them. I did that as checking specifically for mounts related to docker and to check after each container launch and delete I believe would need fine knowledge of which graph driver was being used etc. This simple check should be good enough to catch any dangling mounts - it will just catch them at the end of each test cycle rather than at the exact point of failure.
@mcastelino if you are looking at some of the 'parts of the runtime' or 'mount points' left dangling issues then this test might help you reproduce them in a stable manner.

grahamwhaley:
@MarioCarrilloA I should have said - I also made it so you should be able to run with other RUNTIMEs as well (which helped me sanity check the script ;-) )

chavafg:
lgtm

integration/stability/soak_parallel_rm.sh
```diff
@@ -0,0 +1,198 @@
...
+# limitations under the License.
+
+# This test will run a number of parallel containers, and then try to
+# 'rm -f' them all at the same time. It will check after each run and
+# rm that we have the expected number of containers, shims, proxys,
+# qemus and runtimes active
+# The goals are two fold:
+# - spot any stuck or non-started components
+# - catch any hang ups
+
```
jodh-intel: Might be good to have a `set -e` in here. Note that I checked with that locally, but the script then exited prematurely so I had to change a few of the functions to add `|| true` and also this...

```
((how_many++))
```

... had to change to:

```
how_many=$((how_many + 1))
```

grahamwhaley: hmmm. I'm a bit wary of setting `-e` here - if then docker or something fails because we have a bug then the script would just quit, without a chance of dumping any diagnostics. I think I'd rather have the script find that something 'had not happened', and then give it the chance to dump out the diagnostics.
Let me push an update to the script btw, which checks some other things (that `cc-runtime list` and /var/lib/virtcontainers/pods both have the correct number of items - and also I might have the 'check number of mounts' in as well - but I need to stare at that a bit)... hold on, I'll do a push - but, be warned, with the runtime as it is at the minute the test will fail on the vc/pods check I believe)

grahamwhaley:
@jodh-intel I've pushed the new version - it does also check the mount count.

integration/stability/soak_parallel_rm.sh
jodh-intel: Given what this script is doing, yeah that's a fair point.

integration/stability/soak_parallel_rm.sh
```diff
@@ -0,0 +1,227 @@
...
+		echo "Not a CC runtime, not checking for CC components"
+		check_cc_components=0
+	fi
+}
+
+function spin() {
+	for ((i=1; i<= ITERATIONS; i++)); do {
+		echo "Start loop $i"
+		#spin them up
+		go
```
devimc: I wonder if this can cause clashes with the go command

grahamwhaley: heh - fair comment. I suspect there is some bash parse/ordering/precedent rule, but it's going to be easier for me to just rename the func ...

Issue #497, opened 2017-08-29
metrics: network: docker: Error response from daemon: Conflict.....


MarioCarrilloA:
This error is got when `network-metrics-cpu-consumption.sh` using "runc" as RUNTIME

docker: Error response from daemon: Conflict. The container name "/network-server" is already in use by container "6c8e1fc.......". You have to remove (or rename) that container to be able to reuse that name.

PR #498, opened 2017-08-29
metrics: network: Fix conflict of container already in use

MarioCarrilloA:
This script fails due to a container is already using
the same name. This happens because it does not clean the environment
before to run the tool/test. This commit adds a common function
in order to clean up the current containers running.

Fixes: #497

### DIFF
--- a/metrics/network/network-metrics-cpu-consumption.sh
+++ b/metrics/network/network-metrics-cpu-consumption.sh
@@ -53,6 +53,7 @@ function cpu_consumption {
 	# Arguments to run the client
 	local extra_args="-d"
 
+	init_env
 	setup
 	local server_command="iperf -p ${port} -s"
 	local server_address=$(start_server "$server_name" "$image" "$server_command")


metrics/network/network-metrics-cpu-consumption.sh
```diff
@@ -53,6 +53,7 @@ function cpu_consumption {
 	# Arguments to run the client
 	local extra_args="-d"
 
+	init_env
```
grahamwhaley: Hmm, I see what this is doing, but we might want to do a better 'fix all the tests' fix.... for instance, this would be the only network test calling `init_env`, whereas I suspect the underlying issue and fix(es) exist for pretty much all tests...

The reason (as I'm sure you probably know) you see the 'container already in use' error is because a test probably failed and didn't get to clean up after itself - fine, that is going to happen every now and then. But, then it looks like we have two bits of code for 'cleaning up before we start'...:

We have `init_env`: https://github.com/clearcontainers/tests/blob/master/metrics/lib/common.bash#L70
which uses a `docker ps -qa` to get its list - which will give all containers, running or not (annoyingly it uses a var called `contr_running`, which is, well, wrongly named ;-) ).
And then we have `setup`: https://github.com/clearcontainers/tests/blob/master/metrics/network/lib/network-test-common.bash#L28
which calls `kill_processes_before_start`: https://github.com/clearcontainers/tests/blob/master/metrics/lib/common.bash#L190
which uses a `docker ps -q` to get the list of all running containers, but not all stopped ones (as well as killing off any lingering qemu or shim processes...).

I think the better fix is for us to unify the 'initialise and clean up' code to be a single function (which should probably use `docker ps -qa` and probably do a `stop` and then an `rm` rather than the harsher `rm -f`) - and then make sure we are using that in all the tests.

OK, the caveat, is that when running the tests we will remove any containers the user had 'lying around' on their system, inactive or not - well, I think that is just unavoidable really - do not run the tests on a live system or a system you really care about ;-).

bwarden: I didn't look into test cleanup/setup in depth. It seems like something we should define architecturally though: expected starting state, what we clean up afterward... If we find existing containers, do we warn, abort, and/or offer to stop/remove/purge them? Can we name our containers in a consistent fashion so that we know we can clean them up without confirmation?

I was also reminded that we might want to include thermal condition, system load, etc. as preconditions for starting a test run.

bwarden: And of course all of this would be defined in common functions.

MarioCarrilloA: perfect, thanks @grahamwhaley  and @bwarden , yeah, would be great to work in common function, I can work on that, however I would like to do a kind of re-factor/categorization and I wonder if do @bwarden , @grahamwhaley  agree with that, and also maybe this fix could be merged for the moment what do you think ?

grahamwhaley: I'll strike you a deal @MarioCarrilloA  - you open an Issue with some details (copy from above??) noting we need some refactor, and then we can merge this PR ;-)

MarioCarrilloA: @grahamwhaley Done :)

https://github.com/clearcontainers/tests/issues/503

grahamwhaley:
Jenkins was still in stabilisation when this PR got pushed, and the build got cancelled, and is hence a bit 'stuck' right now. Let me see if I can go and nudge a Jenkins rebuild....

grahamwhaley:
Hi @MarioCarrilloA Looks like that jenkins CI fragment is from a test cycle, and the server has since been moved/re-initialised, so I cannot re-kick a build.
Can you try a rebase/repush which should kick jenkins off again...
and if that fails, I'll force merge it.
/cc @chavafg @jcvenegas

PR #512, opened 2017-09-05
kubernetes: pull images before launching workload

### LINKED ISSUE #511: kubernetes: pull images before launching service
Depending on the network speed, the test might get a timeout when trying to launch a container since the image gets downloaded at that time. We need to pull the image before executing the workload, so we do not get these timeouts.

chavafg:
The nginx test has a timeout for launching the workload, as
the required images are downloaded at that time, sometimes the
test fails because of the timeout. To prevent the failure, this fix pulls the
image before launching the workload.

Also when initializing the kubernetes services, including the
dns pod, sometimes it is not enough 60 seconds to pull all images
and start the services, then, this also increases the timeout of
the initialization to 180 secs.

Fixes #511.

### DIFF
--- a/integration/kubernetes/nginx.bats
+++ b/integration/kubernetes/nginx.bats
@@ -18,25 +18,29 @@
 load "${BATS_TEST_DIRNAME}/lib.sh"
 
 setup() {
+	nginx_image="nginx"
+	busybox_image="busybox"
+	service_name="${nginx_image}-service"
 	export KUBECONFIG=/etc/kubernetes/admin.conf
 	master=$(hostname)
 	sudo -E kubectl taint nodes "$master" node-role.kubernetes.io/master:NoSchedule-
+	sudo -E crioctl image pull "$busybox_image"
+	sudo -E crioctl image pull "$nginx_image"
 }
 
 @test "Verify nginx connectivity between pods" {
-	service_name="nginx"
 	wait_time=20
 	sleep_time=3
 	output_file=$(mktemp)
 	echo $output_file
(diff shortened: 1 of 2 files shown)


integration/kubernetes/nginx.bats
```diff
@@ -18,19 +18,22 @@
 load "${BATS_TEST_DIRNAME}/lib.sh"
 
 setup() {
+	service_name="nginx"
+	image_name="$service_name"
```
MarioCarrilloA: Why are you using 2 vars for the same value?

integration/kubernetes/nginx.bats
```diff
@@ -18,19 +18,22 @@
...
 	export KUBECONFIG=/etc/kubernetes/admin.conf
 	master=$(hostname)
 	sudo -E kubectl taint nodes "$master" node-role.kubernetes.io/master:NoSchedule-
+	sudo -E crioctl image pull busybox
```
grahamwhaley: If we are defining `nginx` in a var, we should probably now also do the same for `busybox` to keep it uniform.

grahamwhaley:
with the minor feedbacks given, fundamentally:
lgtm

integration/kubernetes/nginx.bats
chavafg: changed the value of the service to avoid confusion.

integration/kubernetes/nginx.bats
chavafg: make sense, fixed.

MarioCarrilloA:
lgtm

devimc:
lgtm

Issue #517, opened 2017-09-08
.ci: metrics: the results are not saved in a backup path


MarioCarrilloA:
The ci should save the results of metrics tests execution as reference

PR #518, opened 2017-09-08
ci: metrics: save results in backup path

MarioCarrilloA:
This commit allows to save the results of each metrics
tests execution inside a backup path.

Fixes: #517

### DIFF
--- a/.ci/run_metrics_ci.sh
+++ b/.ci/run_metrics_ci.sh
@@ -19,6 +19,8 @@ REPORT_CMDS=("checkmetrics" "emailreport")
 
 KSM_ENABLE_FILE="/sys/kernel/mm/ksm/run"
 GITHUB_URL="https://github.com"
+RESULTS_BACKUP_PATH="/var/local/localCI/backup"
+RESULTS_DIR="results"
 
 # Verify/install report tools. These tools will
 # parse/send the results from metrics scripts execution.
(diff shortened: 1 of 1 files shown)


.ci/run_metrics_ci.sh
```diff
@@ -97,8 +99,21 @@ pushd "$CURRENTDIR/../metrics"
...
 
-	# Clean env
-	rm -rf "results"
+	# Save the results directory in a backup path. The metrics tests will be
+	# executed each new pull request or when a Pull Request has been modified,
+	# then the results from the same Pull Request number will be identified
+	# by epoch point time as name of the direcory.
+	REPO="$(cut -d"/" -f2 <<<"$LOCALCI_REPO_SLUG")"
+	PR_BK_RESULTS="$RESULTS_BACKUP_PATH/$REPO/$LOCALCI_PR_NUMBER"
+	DEST="$PR_BK_RESULTS/$(date +%s)"
```
grahamwhaley: I like the conciseness and precision of `+%s`, but it loses a lot of contextual info about when the log was generated.
I'd probably favor we generated some sort of ISO compliant date name instead.
reading `date -h`, it is not clear there is an obvious candidate (given we probably want precision in `seconds` to ensure we do not name clash any fast updating PRs...). Best I could find with a quick look was `--iso-8601=seconds`, which gives something like:

`2017-09-11T11:09:56+01:00`

which whilst not pretty, is precise, and does not container any path characters or spaces in it :-)

MarioCarrilloA: @grahamwhaley sounds good, thanks

PR #522, opened 2017-09-08
[RFC]: ci: Rename systemd var for runtime

MarioCarrilloA:
The setup of Clear Containers uses a systemd var in order to
include the path of the runtime, the metrics tests do not work
due to they expect “cc-runtime” as name. This commits rename
cc30 var by cc-runtime.


gorozco1:
lgtm

grahamwhaley:
(marking as DNM to re-enforce the RFC status).
This:
lgtm
as long as there was not a deliberate reason we were using `cc30` - I believe it is just historical before we settled on the name `cc-runtime`, but just want to check with @chavafg that we did not choose `cc30` so that we could, say, have multiple CC3 runtimes installed in parallel without clashing or similar.
Fundamentally, I think we should use the default installation name (`cc-runtime`), so that tests could be hand-run on a pre-installed production system without major modifications if need be for instance.

chavafg:
lgtm
I reviewed the tests of this repository and none of those rely on the name of that docker variable, so this works for us.

Issue #503, opened 2017-09-01
metrics: Re-factorization/Categorization tests


MarioCarrilloA:
Some metrics tests have minimal but relevant details such as:

- Some tests need a pair of things for supporting `runc` as runtime
- Mix different semantic in one test (then mix different tools).
- Save different kind of results in the same file
- Maybe, a pair of functions that could be joined in one
- Missing documentation.

Improving these points we can have a better categorization and at the same time more semantic in the current stuff.

/cc: @bwarden  @grahamwhaley @mcastelino @egernst

grahamwhaley:
Hi @MarioCarrilloA @bwarden Whilst here, I was just utilising the `kill_processes_before_start`:
https://github.com/clearcontainers/tests/blob/master/metrics/lib/common.bash#L192
and although it stops any running containers with a `kill`, I think it would also make sense to then `rm` the ones it stopped. This will help get rid of any 'container name clashes' that @MarioCarrilloA has been seeing like on #524 - and, I could also then use this function as a 'cleanup' function at the end of a test as well (which we seem to be missing - maybe I should just do that in my test and remember which containers I've launched and clean up after myself...)
Thoughts? (although I know @MarioCarrilloA probably has a closely related PR about to land maybe??)

PR #524, opened 2017-09-08
metrics: network: Fix dirty environment

MarioCarrilloA:
This script fails due to a container is already using
the same name. This happens because it does not clean
the environment before to run the tool/test. This commit
uses a common function in order to clean up the current
containers running.


MarioCarrilloA:
Hi @grahamwhaley and @chavafg I am aware that we had a talk regarding to this before, however I would like to have everything working until https://github.com/clearcontainers/tests/issues/503 is solved :)

grahamwhaley:
understood:
lgtm
but, CI failed on many counts (not all your fault ;-))
- some odd CRI-O timeout failure on semaphore
- hit the revision range problem on Travis
- you have no Fixes I think in your commit - so you would fail the checkcommits anyway ;-)

So, I think adding Fixes and re-pushing might clear a number of those.

Issue #523, opened 2017-09-08
metrics: network: iperf test does not work using “runc”


MarioCarrilloA:
This test works with `cc-runtime` however if the runtime is changed by `runc` the test does not work

mcastelino:
This should fix the issue. Or course you need to remove some of the debug stuff I added

```
diff --git a/metrics/network/lib/network-test-common.bash b/metrics/network/lib/network-test-common.bash
index 07bdb9a..91bcd40 100755
--- a/metrics/network/lib/network-test-common.bash
+++ b/metrics/network/lib/network-test-common.bash
@@ -36,7 +36,7 @@ function start_server {
        local server_name="$1"
        local image="$2"
        local server_command="$3"
-       $DOCKER_EXE run -d --runtime="${RUNTIME}" --name=${server_name} ${image} \
+       $DOCKER_EXE run -d --runtime="${RUNTIME}" "--privileged" --name=${server_name} ${image} \
                sh -c "${server_command}" > /dev/null
        local server_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" ${server_name})
        echo "$server_address"
diff --git a/metrics/network/network-metrics-iperf3.sh b/metrics/network/network-metrics-iperf3.sh
index 864a62a..128361e 100755
--- a/metrics/network/network-metrics-iperf3.sh
+++ b/metrics/network/network-metrics-iperf3.sh
@@ -27,12 +27,12 @@ port=5201:5201
 # Image name
 image=gabyct/network
 # Measurement time (seconds)
-time=5
+time=120
 # Name of the containers
 server_name="network-server"
 client_name="network-client"
 # Arguments to run the client
-extra_args="-ti --rm"
+extra_args="-ti --rm --privileged"

 set -e

@@ -45,10 +45,14 @@ test_name="network metrics iperf3"
 function iperf3_bandwidth {
        setup
        local server_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -p ${port} -s"
+        echo $server_command
        local server_address=$(start_server "$server_name" "$image" "$server_command")
+        echo $server_address

-       local client_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -c ${server_address} -t ${time}"
+       local client_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -c ${server_address} -t ${time} -P 2"
+       echo $client_command
[...]

PR #527, opened 2017-09-11
metrics: network: Fix docker access

MarioCarrilloA:
This commit allows access to iperf test
execution using runc.

Fixes: #523

### DIFF
--- a/metrics/network/lib/network-test-common.bash
+++ b/metrics/network/lib/network-test-common.bash
@@ -36,7 +36,8 @@ function start_server {
 	local server_name="$1"
 	local image="$2"
 	local server_command="$3"
-	$DOCKER_EXE run -d --runtime="${RUNTIME}" --name=${server_name} ${image} \
+	local extra_args="$4"
+	$DOCKER_EXE run ${extra_args} -d --runtime="${RUNTIME}" --name=${server_name} ${image} \
 		sh -c "${server_command}" > /dev/null
 	local server_address=$($DOCKER_EXE inspect --format "{{.NetworkSettings.IPAddress}}" ${server_name})
 	echo "$server_address"
--- a/metrics/network/network-metrics-iperf3.sh
+++ b/metrics/network/network-metrics-iperf3.sh
@@ -18,6 +18,8 @@
 #
 # Test inter-container network bandwidth and jitter using iperf3
 
+set -e
+
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 
 source "${SCRIPT_PATH}/lib/network-test-common.bash"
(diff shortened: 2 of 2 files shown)


metrics/network/lib/network-test-common.bash
```diff
@@ -36,7 +36,7 @@ function start_server {
 	local server_name="$1"
 	local image="$2"
 	local server_command="$3"
-	$DOCKER_EXE run -d --runtime="${RUNTIME}" --name=${server_name} ${image} \
+	$DOCKER_EXE run -d --runtime="${RUNTIME}" --privileged --name=${server_name} ${image} \
```
grahamwhaley: As I believe we only need this for the one test, and you fix the 'client' call by adding the priv to the `extra_args` - maybe it would be better if we add an `extra_args` argument to the start_server func like we have in the client, and then add the priv to the iperf3 test code only? I think that make the code more uniform, and keeps the 'special case' code in one place (in the relevant test).

grahamwhaley: A question for @mcastelino (and anybody else who knows) - so, is `--privileged` benign in the `cc-runtime` case?
We should correlate that with the limitations doc:
https://github.com/clearcontainers/runtime/blob/master/docs/limitations.md#docker-run---privileged
that currently says 'unsupported'. We can change that to 'benign' or maybe 'impotent' if necessary.

MarioCarrilloA: make sense, thanks

MarioCarrilloA:
Hi @grahamwhaley , I am aware maybe this new modification is not uniform too, however these functions (client/server) are in my scope about this issue https://github.com/clearcontainers/tests/issues/503

MarioCarrilloA:
@grahamwhaley oh yeah, @mcastelino  wrote the solution here https://github.com/clearcontainers/tests/issues/523 for that reason I include him as author :)

grahamwhaley:
@MarioCarrilloA sure, I understand why we have @mcastelino as the author, and you signed it off, but pullapprove (and maybe checkcommits) won't be happy with that - they check that the author matches the signed off by.
Normally I'd say having two signed-off lines would not be the solution, but that might work - especially if we get them in the 'correct order' - that is, @mcastelino sign-off first, and yours after. As a 'process' that works for me - what we need is @mcastelino to ack that you can sign it on his behalf (just for completeness).
Either that or you could modify the original author with @mcastelino consent, and maybe add some relevant 'Originally-by: mano..@...com' line to give some acknowledgement.

I know it's not a big deal, and it is 'only' a small patch - but let's try to get this right, as it will also help us to understand how to 'tickle' the CI system in the future if need be.

MarioCarrilloA:
Hi @grahamwhaley  thanks for you feedback, and I agree, let's try to get this right, and I will try the suggestion about the 'correct order' :)

Issue #532, opened 2017-09-14
metrics: Allows to run only certain group of metrics tests


MarioCarrilloA:
The `run_all_metrics.sh` just allow to execute all tests, however sometimes just could be relevant to run some of them

PR #533, opened 2017-09-14
metrics: run only a group of tests

MarioCarrilloA:
This commits allows to run just a certain group of tests, however
if no option is specified, it will run all metrics tests by default.

Fixes: #532

### DIFF
--- a/metrics/run_all_metrics.sh
+++ b/metrics/run_all_metrics.sh
@@ -31,32 +31,104 @@ echo "Check dependencies"
 cmds=("docker" "bc" "awk" "smem" "ab")
 check_cmds "${cmds[@]}"
 
-echo "Running all metrics tests"
+function help() {
+echo "$(cat << EOF
+Usage: $0 [-h] [--help] [-v] [--version]
+   Description:
+        This script allows to execute the metrics tests
+        in different groups which are focused in different
+        aspects of the containers. If no option is specified
+        all tests will be executed.
+   Options:
+        -h, --help         Shows help
+        -l, --latency      Run latency/time metrics tests
+        -m, --memory       Run memory metrics tests
+        -n, --network      Run network metrics tests
+        -s, --storage      Run I/O storage metrics tests
+EOF
+)"
+}
 
-# Run the time tests
-bash ${SCRIPT_PATH}/time/docker_workload_time.sh true busybox $RUNTIME 100
+# Only run latency metrics tests
+function run_latency_tests() {
+	# Run the time tests
+	bash ${SCRIPT_PATH}/time/docker_workload_time.sh true busybox $RUNTIME 100
+}
 
-# Run the network metrics
-bash ${SCRIPT_PATH}/network/network-latency.sh
-bash ${SCRIPT_PATH}/network/network-metrics-cpu-consumption.sh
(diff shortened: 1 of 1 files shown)


metrics/run_all_metrics.sh
```diff
@@ -31,32 +31,104 @@ echo "Check dependencies"
...
+}
+
+# Only run latency metrics tests
+function run_latency_tests() {
+	# Run the time tests
+	bash ${SCRIPT_PATH}/time/docker_workload_time.sh true busybox $RUNTIME 100
+}
+
+# Only run network metrics tests
+function run_network_tests() {
```
mcastelino: maybe good to split the network tests into network minimum and full. They take a long time to run. Most of the time we care only about iperf3

MarioCarrilloA: then iperf3 tests could be "minimum" tests group?

chavafg:
lgtm

metrics/run_all_metrics.sh
grahamwhaley: good point @mcastelino  In fact, we could split all the groups (density, time, io, network) into two groups each - 'core' and 'full'. We may then be able to replace https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh with a call to this script for 'core', and remove some duplication.

metrics/run_all_metrics.sh
```diff
@@ -31,32 +31,104 @@ echo "Check dependencies"
...
-# Run I/O storage tests
-bash ${SCRIPT_PATH}/storage/fio_job.sh -b 16k -o randread -t "storage IO random read bs 16k"
-bash ${SCRIPT_PATH}/storage/fio_job.sh -b 16k -o randwrite -t "storage IO random write bs 16k"
-bash ${SCRIPT_PATH}/storage/fio_job.sh -b 16k -o read -t "storage IO linear read bs 16k"
-bash ${SCRIPT_PATH}/storage/fio_job.sh -b 16k -o write -t "storage IO linear write bs 16k"
+function help() {
+echo "$(cat << EOF
+Usage: $0 [-h] [--help] [-v] [--version]
+   Description:
+        This script allows to execute the metrics tests
```
grahamwhaley: s/to execute/execution of/

metrics/run_all_metrics.sh
grahamwhaley: For reference, our current 'core' network tests look like (from the ci script)...
```
# ops/second
bash network/network-nginx-ab-benchmark.sh
# ping latency
bash network/network-latency.sh
# Bandwidth and jitter
bash network/network-metrics-iperf3.sh
# UDP bandwidths and packet loss
bash network/network-metrics-nuttcp.sh
```

@mcastelino - if you think we should add/remove from that core network group then feel free to comment.

grahamwhaley: Note: if we do move to splitting the functions and then invoking this script for the metrics CI, we will have to work out how to take care of KSM on/off testing of memory footprint - the current ci script does do that, and invokes the memory test with different parameters ('settle time') depending on the KSM mode for instance.

metrics/run_all_metrics.sh
MarioCarrilloA: sounds better :)

PR #535, opened 2017-09-15
kill: Sleep before sending a signal

### LINKED ISSUE #534: tests: docker: kill_test needs a lock around signal delivery
The docker kill_test suffers from the race condition described at:
https://github.com/clearcontainers/runtime/issues/430

and needs to have some sort of lock-step mechanism inserted between the 'run' and the 'kill' here:
https://github.com/clearcontainers/tests/blob/master/integration/docker/kill_test.go#L78
in order to guarantee the container has run the 'trap' handler and is ready to receive the signal from the kill command.

sameo:
We have to let the container start and process the trap before
sending it the expected signal.

Fixes #534

Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>

### DIFF
--- a/integration/docker/kill_test.go
+++ b/integration/docker/kill_test.go
@@ -76,6 +76,10 @@ var _ = Describe("docker kill", func() {
 
 			DockerRun(args...)
 
+			// we have to wait for the container workload
+			// to process the trap.
+			time.Sleep(5 * time.Second)
+			
 			if signal > 0 {
 				DockerKill("-s", fmt.Sprintf("%d", signal), id)
 			} else {


sameo:
cc @grahamwhaley

grahamwhaley:
Got a Jenkins fail, but I cannot obviously see where in the log:
http://cc-jenkins-ci.westus2.cloudapp.azure.com/job/clear-containers-tests-k8s/28/console

grahamwhaley:
OK, looking in the 'console full log' I see maybe some more 'fail' and 'error' lines... but, needs somebody who understands what those tests were trying to do to diagnose

PR #543, opened 2017-09-18
CI: Use new runtime config if available

### LINKED ISSUE #542: Read runtime config from defaults dir
https://github.com/clearcontainers/runtime/pull/564 changes the runtime to create the config as:

```
/usr/share/defaults/clear-containers/configuration.toml
```
... rather than the old:

```
/etc/clear-containers/configuration.toml
```

The latter is still supported and in fact takes priority over the former, but since the former is the path the runtime will be installing from now on, this needs to be the path the runtime scripts in this repo work with.

jodh-intel:
If the new runtime config file exists, use it over the legacy version
(which is deleted for safety).

Fixes #542.

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -43,6 +43,17 @@ export LOCALSTATEDIR=/var
 
 runtime_config_path="${SYSCONFDIR}/clear-containers/configuration.toml"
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
This is needed to unblock https://github.com/clearcontainers/runtime/pull/564.

devimc:
lgtm

PR #544, opened 2017-09-18
CI: Fix typo stopping new runtime config being used

### LINKED ISSUE #542: Read runtime config from defaults dir

jodh-intel:
PR #543 attempted to make the tests use the new runtime config file, but
the fix was incorrect due to a silly typo.

Really fixes #542.

### DIFF
--- a/.ci/install_runtime.sh
+++ b/.ci/install_runtime.sh
@@ -43,7 +43,7 @@ export LOCALSTATEDIR=/var
 
 runtime_config_path="${SYSCONFDIR}/clear-containers/configuration.toml"
 
-PKGDEFAULTSDIR="${SHARE_DIR}/defaults/clear-containers"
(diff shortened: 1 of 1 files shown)


grahamwhaley:
lgtm (now waiting for CI to clear)

jodh-intel:
Grrr... this PR is still waiting on semaphore even though we've disabled it. Aslo, I cannot force-merge so will have to close and re-open...

PR #545, opened 2017-09-18
data: ci: jenkins: Save Jenkins configuration

### LINKED ISSUE #540: Store jenkins configuration in this repo
Maybe under `data/ci/jenkins/config/` or similar.

sboeuf:
This commit creates a new folder responsible for holding
configuration information about our Jenkins server. This
way, it is going to be easier if anyone tries to replicate
the same kind of Jenkins CI.

Fixes #540

### DIFF
--- a/data/ci/jenkins/config/config.xml
+++ b/data/ci/jenkins/config/config.xml
@@ -1,0 +1,283 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<hudson>
+  <disabledAdministrativeMonitors>
+    <string>hudson.diagnosis.ReverseProxySetupMonitor</string>
+    <string>OldData</string>
+  </disabledAdministrativeMonitors>
+  <version>2.60.2</version>
+  <numExecutors>2</numExecutors>
+  <mode>NORMAL</mode>
+  <useSecurity>true</useSecurity>
+  <authorizationStrategy class="org.jenkinsci.plugins.GithubAuthorizationStrategy" plugin="github-oauth@0.27">
+    <rootACL>
+      <organizationNameList class="linked-list">
+        <string></string>
+      </organizationNameList>
+      <adminUserNameList class="linked-list">
+        <string>sboeuf</string>
+        <string>clearcontainersbot</string>
+        <string>sameo</string>
+        <string>mcastelino</string>
+        <string>grahamwhaley</string>
+        <string>chavafg</string>
+        <string>jcvenegas</string>
+      </adminUserNameList>
+      <authenticatedUserReadPermission>true</authenticatedUserReadPermission>
+      <useRepositoryPermissions>true</useRepositoryPermissions>
+      <authenticatedUserCreateJobPermission>false</authenticatedUserCreateJobPermission>
+      <allowGithubWebHookPermission>true</allowGithubWebHookPermission>
+      <allowCcTrayPermission>false</allowCcTrayPermission>
+      <allowAnonymousReadPermission>true</allowAnonymousReadPermission>
+      <allowAnonymousJobStatusPermission>false</allowAnonymousJobStatusPermission>
+    </rootACL>
+  </authorizationStrategy>
+  <securityRealm class="org.jenkinsci.plugins.GithubSecurityRealm">
+    <githubWebUri>https://github.com</githubWebUri>
+    <githubApiUri>https://api.github.com</githubApiUri>
+    <clientID>dde1b1de183181e018b7</clientID>
+    <clientSecret>{AQAAABAAAAAwWLmVHCPZQnaZy6txJ2YqPk0kZup4D6uGbL1RDhzMSeUsTsFpkOzKa7rKY+jgUF5sVTGn/aQ8Tpnu1J3kapjgcw==}</clientSecret>
+    <oauthScopes>read:org,user:email</oauthScopes>
+  </securityRealm>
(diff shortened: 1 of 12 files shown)


data/ci/jenkins/config/config.xml
```diff
@@ -0,0 +1,283 @@
...
+<hudson>
+  <disabledAdministrativeMonitors>
+    <string>hudson.diagnosis.ReverseProxySetupMonitor</string>
+    <string>OldData</string>
+  </disabledAdministrativeMonitors>
+  <version>2.60.2</version>
+  <numExecutors>2</numExecutors>
+  <mode>NORMAL</mode>
+  <useSecurity>true</useSecurity>
+  <authorizationStrategy class="org.jenkinsci.plugins.GithubAuthorizationStrategy" plugin="github-oauth@0.27">
```
jodh-intel: I'm guessing that `0.27` is a version number? What is our strategy to ensure we keep these plugins current to minimise any potential security impact?

data/ci/jenkins/config/config.xml
```diff
@@ -0,0 +1,283 @@
...
+      <allowCcTrayPermission>false</allowCcTrayPermission>
+      <allowAnonymousReadPermission>true</allowAnonymousReadPermission>
+      <allowAnonymousJobStatusPermission>false</allowAnonymousJobStatusPermission>
+    </rootACL>
+  </authorizationStrategy>
+  <securityRealm class="org.jenkinsci.plugins.GithubSecurityRealm">
+    <githubWebUri>https://github.com</githubWebUri>
+    <githubApiUri>https://api.github.com</githubApiUri>
+    <clientID>dde1b1de183181e018b7</clientID>
+    <clientSecret>{AQAAABAAAAAwWLmVHCPZQnaZy6txJ2YqPk0kZup4D6uGbL1RDhzMSeUsTsFpkOzKa7rKY+jgUF5sVTGn/aQ8Tpnu1J3kapjgcw==}</clientSecret>
```
jodh-intel: Is this a problem?

data/ci/jenkins/config/config.xml
```diff
@@ -0,0 +1,283 @@
...
+          <isInstallMaven>false</isInstallMaven>
+          <isInstallDocker>true</isInstallDocker>
+          <osType>Linux</osType>
+          <imagePublisher>Canonical</imagePublisher>
+          <imageOffer>UbuntuServer</imageOffer>
+          <imageSku>16.04-LTS</imageSku>
+          <imageVersion>latest</imageVersion>
+          <agentLaunchMethod>SSH</agentLaunchMethod>
+          <preInstallSsh>true</preInstallSsh>
+          <initScript># Install java
```
jodh-intel: Looking over this changeset, there are lots of what appear to be duplicated shell script snippets, which is a potential maintenance problem. Could we instead create a set of stand-alone scripts and have each `<initScript/>` section just call the appropriate script?

data/ci/jenkins/config/config.xml
```diff
@@ -0,0 +1,283 @@
...
+sudo add-apt-repository ppa:openjdk-r/ppa -y
```
jodh-intel: I'd stick the `-y` immediately after the command in all cases as it's easier to parse ;)

jodh-intel:
Hi @sboeuf - thanks for raising this! I'm not familar with setting up Jenkins systems, so how about adding a brief `data/ci/jenkins/README.md` to explain in simple terms:

- How to use this stored config
- How and when these files should be updated

data/ci/jenkins/config/config.xml
grahamwhaley: I concur, it would be nice to have the scripty bits stored in a repo somewhere and not tied directly into the Jenkins UI/xml (which is how I think it currently is). I know we start with a 'bare OS image', so we will always have to have some script in Jenkins to kick off the install process, but maybe that can be more of the form:
```
apt-get install git
git pull clearcontainers/tests
./clearcontainers/tests/data/ci/jenkins/install.sh
```
or similar.
Anyway, that is probably something we can improve in the future - for now having these configs in the repo is a good first step :-)

grahamwhaley: and, yes, having scrolled through the files, they are pretty much replications across the repos and distros - so I agree there seems to be a lot of scope for refactoring here in the future.

data/ci/jenkins/config/jobs/clear-containers-runtime-azure-ubuntu-16-04/config.xml
```diff
@@ -0,0 +1,158 @@
...
+export GOPATH=$HOME/go
+export PATH=$GOPATH/bin:/usr/local/go/bin:$PATH
+
+echo &quot;GOROOT=$GOROOT&quot;
+echo &quot;GOPATH=$GOPATH&quot;
+echo &quot;PATH=$PATH&quot;
+echo &quot;USER=$USER&quot;
+
+export CI=true
+go get github.com/clearcontainers/runtime || true
```
jodh-intel: This should be (here and in the 17.04 script):

```
go get -d github.com/clearcontainers/runtime
```

... because we're seeing errors in the logs as `go get` is trying to build the runtime without calling `make`.

sboeuf:
Closing this PR since this is not what we want to do. Look at #540 for more details.

PR #552, opened 2017-09-19
CI: Only check new runtime config file after installation

### LINKED ISSUE #551: Fix silly logic error in install_runtime.sh
`install_runtime.sh` would only switch to using the new runtime config file if the runtime *was already installed*. This is clearly not the case in a CI environment.

jodh-intel:
The ".ci/install_runtime.sh" script was trying to switch to using the
new config file before the runtime was installed (and hence before the
config file could have existed).

Fixes #551.


devimc:
lgtm

PR #553, opened 2017-09-19
metrics: Re-factorization/Categorization

MarioCarrilloA:
The objective of this group of commits is to improve the following points:

- Remove redundant code
- Fix those tests that do not support `runc` correctly
- Categorize results according to tests in different file
- Improve documentation (includes README files and code comments)

Fixes: #503


metrics/network/network-metrics-iperf3.sh
```diff
@@ -16,46 +16,41 @@
...
-	extra_args="-ti --privileged --rm"
-else
-	extra_args="-ti --rm"
-fi
-
-## Test name for reporting purposes
-test_name="network metrics iperf3"
-
-# This script will perform all the measurements using a local setup using iperf3
+transmit_time=5
```
grahamwhaley: I'd probably be happier if we s/time/count/  `time` always makes me think it is an 'amount of time' measure rather than a 'number of times' one.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -16,46 +16,41 @@
...
+# Arguments to run the client/server
+client_extra_args="-ti --privileged --rm"
+server_extra_args="--privileged"
```
grahamwhaley: I think I mentioned this before somewhere....
iirc, those `--priv` args are for non-CC runtimes only - but I suspect we are passing them in when running CC as well. Officially we do not support that arg (https://github.com/clearcontainers/runtime/blob/master/docs/limitations.md#docker-run---privileged) - and maybe we 'get away' with it right now as we might just ignore it.
I'd be happier if we didn't supply the arg to CC runtimes if possible - and also add a comment on why `--priv` is required above.

metrics/network/network-metrics-memory-pss.sh
```diff
@@ -47,32 +46,42 @@ function pss_memory {
...
+        if [ "$RUNTIME" == "runc" ]; then
+                process="iperf"
+        elif [ "$RUNTIME" == "cor" ] || [ "$RUNTIME" == "cc-runtime" ]; then
+                process="$QEMU_PATH"
+        else
+                die "Unknown runtime: $RUNTIME"
+        fi
+
 	local memory_command="smem --no-header -c pss"
-	${memory_command} -P ${QEMU_PATH} > "$result"
```
grahamwhaley: ah - subtle bug you found/fixed?, missing the '^', so maybe this counted the smem tool footprint as well as the runtime footprint!?

MarioCarrilloA: yeah, and in the case of runc execution, it returns that value :(, then even if there is no any qemu instance running it returned a result :(

metrics/lib/common.bash
```diff
@@ -86,6 +86,18 @@ function init_env()
...
 }
 
+# Clean environment, this function will try to remove all
+# stopped/running containers, it is advisable to use this function
+# in the final of each metrics test.
+function clean_env()
+{
+	containers_running=$(docker ps -qa)
+
+	if [ ! -z "$containers_running" ]; then
```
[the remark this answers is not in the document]
MarioCarrilloA: make sense :)

metrics/network/network-metrics-iperf3.sh
MarioCarrilloA: yeah, that sounds better

MarioCarrilloA: checking tool options it seems that it is a unit time `-t, --time      #        time in seconds to transmit for (default 10 secs)`

metrics/lib/common.bash
```diff
@@ -70,14 +70,16 @@ function check_cmds()
...
-	fi
+	# This clean up is more aggressive, this is in order to
+	# decrease the factors that could affect the metrics results.
+	kill_processes_before_start
+
+	# Remove all stopped containers
+	clean_env
 
 	# Restart services
 	systemctl restart docker
```
grahamwhaley: Hi @MarioCarrilloA Whilst I was running up the metrics code under Jenkins, I realised the network common code does a system level activity here (`systemctl`) without necessarily having raised privs (`sudo`). This was failing for me under Jenkins as the tests are run as the Jenkins user. I had to `sudo` the whole `run_metrics_ci.sh` script to work around it, which was not ideal :-(
I think we should discuss and define what the 'rules for `sudo`' are for the metrics code base. My premise is:
- all commands in scripts that require raised privs should be run with `sudo`
- no script should require to be run, at the top level, as `sudo`

What do you think?
/cc @bwarden @chavafg

bwarden: I agree -- sudo should be applied to the specific commands where needed, and nowhere else. Definitely not at script level.

grahamwhaley: yeah, I meant to also say:
- run all scripts at the lowest priv they need to operate, and have them elevate when and only when necessary :-)

The one place we sort of do away with this for convenience is that we presume the id running the tests is a member of the docker group so we don't have to sudo every docker command - seeing as that is our 'bread and butter', I see that as an acceptable exception :-)

bwarden: We should also report and bail early if we don't have permission for docker.

MarioCarrilloA: perfect thanks :)

metrics/lib/common.bash
```diff
@@ -70,14 +70,16 @@ function check_cmds()
...
+	# check dependencies
 	check_cmds "${cmd[@]}"
 
-	# Verify containers running
-	if [ ! -z "$contr_running" ];then
-		docker rm -f $(docker ps -qa)
-	fi
+	# This clean up is more aggressive, this is in order to
+	# decrease the factors that could affect the metrics results.
+	kill_processes_before_start
```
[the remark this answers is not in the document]
MarioCarrilloA: it seems that `docker rm -f` and `docker kill` is the same

MarioCarrilloA:
Hi @grahamwhaley I have taken account your preliminary feedback thanks for that, however at this moment I think the new changes/commits cover all requirements to fix in the issue #503 , @chavafg, @bwarden @iphutch  I wonder if you could take a look too please :)

metrics/storage/README.md
```diff
@@ -0,0 +1,10 @@
+# Storage metrics tests
+
+This section includes metrics tests focused on measure storage bandwith under different
+conditions/configuration where could be executed different I/O operation such as: read,
+write, random read etc.
+However the use of these should be complemented with more tools in order to find the hot spot.
+
+- `fio_job.sh`: measures the bandwidth measured by a specif I/O operation (read, write,
```
iphutch: s/specif/specific
Also, Im confused by the double use of "measure". Could you replace the second "measure" with a word that better describes what the I/O operations are doing?

metrics/storage/README.md
```diff
@@ -0,0 +1,10 @@
...
+   random-read, etc), and using a default configuration, the operations and configuration
```
iphutch: break this into two sentences. If you are saying that there is a default, it may be best to re-work the sentence order:

Measures storage bandwidth. This test is executed with a specific I/O operation (read, write, random-read, etc) using a default configuration. However, the default operations and configuration can be overwritten.

metrics/time/README.md
```diff
@@ -0,0 +1,10 @@
+# Time metrics tests
+
+This section includes metrics tests focused on measure the time interval[s] of a determined
```
iphutch: Again, "focused on measuring" OR "tests to measure"

metrics/time/README.md
```diff
@@ -0,0 +1,10 @@
...
+number of containers executing a defined workload. This kind of metrics tests could provide
+a general overview about the time spent by a flow, however the use of these should be
+complemented with more tools in order to find the hot spot.
```
iphutch: in order to find the hot spot, these tests should be used in combination with more tools.

metrics/time/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+- `docker_workload_time.sh`: measures the complete workload of a container using docker. The
```
iphutch: Docker\*

metrics/time/README.md
```diff
@@ -0,0 +1,10 @@
...
+   workload of this test is to execute a `true`, the reason behind this is not to add overhead
+   and measure the container flow execution.
```
iphutch: This sentence is confusing. Here's my take:
measures the time taken for a container using Docker\* to complete a workload. In this test, the workload is to execute a `true`. By using this workload, the test does not add overhead when measuring the container flow execution time.

metrics/density/README.md
```diff
@@ -0,0 +1,10 @@
+# Density metrics tests
+
+This section includes metrics tests focused on measure the amount of memory used by
+a determined number of containers or a specific process that is part of the flow.
+This number of containers could be working under different configurations/conditions.
+
+- `docker_memory_usage.sh`: measures Proportional Set Size memory average about a
+   number of containers launched in idle mode, It uses a `sleep` for a configurable
+   period of time to allow any memory optimisations to 'settle' (e.g. KSM execution).
+   This test uses smem tool to get the memory used.
```
iphutch: Some wording changes. Also, have some indicator of what `sleep` is. What is `settle`?/does it need to be literal text?: 
measures the Proportional Set Size (PSS) memory average for a
number of containers launched in idle mode. This test uses the ________ `sleep` for a configurable
period of time to allow any memory optimizations to 'settle' (e.g. KSM execution).

MarioCarrilloA: Hi @iphutch `sleep` is command that pauses a certain time the execution of the tests script, it is very common/popular to use it, and about `settle` word, it makes reference to KSM execution. KSM is a linux module that looks for 2 or more programs running in the same environment, then it compares their memory and if there are identical memory pages, KSM will merged in one, this process takes time and if someone would like to check KSM effects needs to wait KSM ans after take the metric sample(sttle).

metrics/storage/README.md
```diff
@@ -0,0 +1,10 @@
+# Storage metrics tests
+
+This section includes metrics tests focused on measure storage bandwith under different
```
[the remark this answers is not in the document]
MarioCarrilloA: sounds better thanks :)

metrics/storage/README.md
MarioCarrilloA: I am going to rephrase that :)

metrics/density/README.md
iphutch: Great, maybe this will work:

`docker_memory_usage.sh`: measures the Proportional Set Size (PSS) memory average for a
number of containers launched in idle mode. This test uses the`sleep` command to allow any memory optimizations to 'settle' (e.g. KSM execution) for a configurable period of time.

metrics/storage/README.md
```diff
@@ -0,0 +1,11 @@
+# Storage metrics tests
+
+This section includes metrics tests that measure storage bandwith under different
+conditions/configuration. These tests can be executed with different I/O operations
+such as: read, write, random read, etc.
+However, in order to find the hot spot, these tests should be used in combination
+with more tools.
+
+- `fio_job.sh`: measures the bandwidth using a specific I/O operation (read, write,
```
iphutch: Once you rephrase, I will comment. Just need a flag here to remind myself!

metrics/density/README.md
MarioCarrilloA: cool! sounds better :)

MarioCarrilloA:
Hi @iphutch thanks for feedback, I have fixed all commented points :)

metrics/storage/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This section includes metrics tests that measure storage bandwith under different
+conditions/configuration. These tests can be executed with different I/O operations
+such as: read, write, random read, etc.
+However, in order to find the hot spot, these tests should be used in combination
+with more tools.
+
+- `fio_job.sh`: measures the I/O bandwidth using a specific storage operation (read, write,
+   random-read, etc), and this is under a default configuration, the operations and
+   configuration can be overwritten.
```
iphutch: Let's break this up:

random-read, etc) under a default configuration. The default operations
and configuration can be overwritten.

MarioCarrilloA:
@iphutch the point suggested above done :)

metrics/network/network-metrics-memory-pss-1g.sh
```diff
@@ -46,36 +49,50 @@ function pss_memory {
...
+	local server_extra_args="-i --name=$server_name"
+
+	# Initialize/clean environment
+	init_env
 
-	setup
-	local server_command="sleep 30"
-	local server_address=$(start_server "$server_name" "$image" "$server_command")
+	local server_command="sh"
+	local server_address=$(start_server "$image" "$server_command" "$server_extra_args")
```
chavafg: same comment as before, not checking if `server address` has a value

metrics/network/network-nginx-ab-benchmark.sh
```diff
@@ -17,62 +17,54 @@
...
+	total_requests="tmp.log"
+
+	# Initialize/clean environment
+	init_env
 
-	$DOCKER_EXE run -d --name ${container_name} --runtime $RUNTIME -p ${port} ${image} > /dev/null
+	# Launch nginx container
+	$DOCKER_EXE run -d -p $port $image
 	sleep_secs=2
 	echo >&2 "WARNING: sleeping for $sleep_secs seconds (see https://github.com/01org/cc-oci-runtime/issues/828)"
```
chavafg: Can you change this message for something like: "Sleeping for $sleep_secs to let the container start correctly"
Issue 828 is already closed.

metrics/network/network-metrics-iperf3.sh
```diff
@@ -16,46 +16,44 @@
...
 
 # Test single direction TCP bandwith
-function iperf3_bandwidth {
-	setup
+function iperf3_bandwidth() {
+	local test_name="network iperf bandwidth"
 	local server_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -p ${port} -s"
-	local server_address=$(start_server "$server_name" "$image" "$server_command" "$extra_args")
+	local server_address=$(start_server "$image" "$server_command" "$server_extra_args")
```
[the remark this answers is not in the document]
MarioCarrilloA: oh, yes, thanks :)

metrics/network/network-nginx-ab-benchmark.sh
MarioCarrilloA: great...!!!! thanks for update

metrics/lib/common.bash
```diff
@@ -70,19 +70,35 @@ function check_cmds()
...
+		sudo systemctl restart cc-proxy
+	fi
+}
+
+# Clean environment, this function will try to remove all
+# stopped/running containers, it is advisable to use this function
+# in the final of each metrics test.
+function clean_env()
+{
+	containers_running=$(docker ps -qa)
```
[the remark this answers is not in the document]
MarioCarrilloA: oh yes, you are right, thanks!

MarioCarrilloA:
@chavafg all points commented above have been fixed :)

chavafg:
lgtm

metrics/storage/README.md
grahamwhaley: @MarioCarrilloA still need to fix feedback here?

Issue #559, opened 2017-09-20
CI: metrics: disable proxy auto ksm when running metrics


grahamwhaley:
The proxy now does auto KSM management by default, but that is not what we want to do when we are running the metrics CI system, as we need a guaranteed stable, controlled and reproducible environment.
Modify the CI scripts so they can spot when installing under the metrics CI, and take the appropriate actions.

PR #560, opened 2017-09-20
ci: install_proxy: disable proxy ksm mode if on metrics CI

grahamwhaley:
When running under the metrics CI system we do not want the proxy
auto-modifying the KSM modes - we control those from the CI to ensure
we know exactly what we are testing and to get steady state results.

Modify the .ci proxy install script so that it adds the correct
argument to the sytemd file, or copies the correct (added in this
PR) upstart file into place.

Fixes: #559

### DIFF
--- a/.ci/install_proxy.sh
+++ b/.ci/install_proxy.sh
@@ -23,12 +23,29 @@ source "${cidir}/lib.sh"
 
 clone_build_and_install "github.com/clearcontainers/proxy"
 start_proxy_cmd="sudo systemctl start cc-proxy"
+upstart_source_file="${cidir}/data/cc-proxy.conf"
+upstart_dest_file="/etc/init/cc-proxy.conf"
+systemd_extra_flags="-log debug"
 
+
+# If we are running under the metrics CI system then we do not want the proxy
+# to be dynmaically changing the KSM settings under us - we need control of them
+# ourselves
+if [[ $METRICS_CI ]]; then
+	upstart_source_file="${cidir}/data/cc-proxy.conf.noksm"
+	systemd_extra_flags="${systemd_extra_flags} -ksm initial"
+fi
+
+# Are we running upstart or systemd?
 if [[ ! $(ps -p 1 | grep systemd) ]]; then
 	echo "Install proxy service (/etc/init/cc-proxy.conf)"
-	sudo cp "${cidir}/data/cc-proxy.conf" /etc/init/
+	sudo cp "${upstart_source_file}" "${upstart_dest_file}"
 
 	start_proxy_cmd="sudo service cc-proxy start"
+else
+	proxy_systemd_file=$(sudo systemctl show cc-proxy | fgrep FragmentPath | awk 'BEGIN{FS="="}{print $2}')
+	sudo sed -i "s/\(^ExecStart.*$\)/\1 ${systemd_extra_flags}/" "${proxy_systemd_file}"
+	sudo systemctl daemon-reload
 fi
 
 echo "Start proxy service"
--- a/.ci/data/cc-proxy.conf.noksm
+++ b/.ci/data/cc-proxy.conf.noksm
@@ -1,0 +1,10 @@
+description "Clear Containers Proxy"
+
+# start in normal runlevels when disks are mounted and networking is available
+start on runlevel [2345]
+
+# stop on shutdown/halt, single-user mode and reboot
+stop on runlevel [016]
+
+# Do not enable auto KSM - unwanted under the metrics CI system
+exec /usr/libexec/clear-containers/cc-proxy -log debug -ksm initial


grahamwhaley:
@chavafg @MarioCarrilloA I've not run tested this - just an RFC right now... do you think this is the right approach?
I've 'assumed' we can set an env var to be able to tell we are under the metrics CI - I don't think we have that right now - is there any current way you know of?

Reason I think we want this is here: https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh#L44 we set/unset KSM - and what we are seeing right now is that 'unset' is not working, and I suspect because the new proxy ksm auto mode is re-enabling it when we launch a container :-)

.ci/install_proxy.sh
```diff
@@ -23,10 +23,23 @@ source "${cidir}/lib.sh"
...
 start_proxy_cmd="sudo systemctl start cc-proxy"
+upstart_source_file="${cidir}/data/cc-proxy.conf"
+upstart_destr_file="/etc/init/cc-proxy.conf"
+
+
+# If we are running under the metrics CI system then we do not want the proxy
+# to be dynmaically changing the KSM settings under us - we need control of them
+# ourselves
+if [[ $METRICS_CI ]]; then
+	proxy_systemd_file=$(sudo systemctl show cc-proxy | fgrep FragmentPath | awk 'BEGIN{FS="="}{print $"}')
```
jodh-intel: The `print $"` doesn't look quite right? (only a single double-quote and no variable?)

.ci/install_proxy.sh
```diff
@@ -23,10 +23,23 @@ source "${cidir}/lib.sh"
...
+	sudo sed -i 's/\(^ExecStart.*$\)/\1 -ksm initial/' "${proxy_systemd_file}"
```
jodh-intel: Won't this require a `sudo systemctl daemon-reload` after the `sed` (note that Upstart doesn't require any further action)?

.ci/install_proxy.sh
```diff
@@ -23,10 +23,23 @@ source "${cidir}/lib.sh"
...
+
+	upstart_source_file="${cidir}/data/cc-proxy.conf.noksm"
```
jodh-intel: OOI, why a different file for Upstart but a `sed` to the existing one for systemd?

MarioCarrilloA:
Hi @grahamwhaley , I wonder if could be better to handle this inside of `docker_memory_usage.sh` test, in order to someone could reproduce KSM results without using .ci stuff, furthermore I think at this moment KSM is just relevant for `docker_memory_usage.sh` test. However this could involve more changes due to the patch change the systemd unit configuration and maybe the best way is to return again the previous configuration after test.

.ci/install_proxy.sh
grahamwhaley: yep, typo - s/"/2/ (slip of the shift key). thx

.ci/install_proxy.sh
grahamwhaley: ah, so it will! I wonder if that happens elsewhere right now - maybe with the prior clone_build_and_install...
This will be the first launch of the proxy I believe, as this is the 'start from scratch every time' CI - so, maybe the $start_proxy_cmd just 'works' - but, I think a daemon reload is still appropriate - I'll go add.

.ci/install_proxy.sh
grahamwhaley: I just modified what existed - I think this is because the systemd file comes from the proxy code/install itself, but the upstart file comes from the .ci tree (where this script lives).
@chavafg I'll let you comment on that if need be

grahamwhaley:
Hi @MarioCarrilloA Valid point about if this would be better handled in the `docker_memory_usage.sh`. I'm leaning towards no. We probably want to disable KSM auto stuff for all the metrics tests, to remove any unpredictable 'noise' we might get from the extra sporadic CPU usage for instance.
What I think we can do is add an 'assert' to the `docker_memory_usage.sh` to check that KSM really did turn off when we asked it to - at least we will get less surprises then...
As for returning the system back to the default state. I can see that would be nice, but this is an install for the CI system - so, we reset it for every run. The change is also gated on the $METRICS_CI - so, hopefully even if somebody is hand running this for a non-CI use case they won't have that set.

grahamwhaley:
updated with the obvious fixes - thanks @jodh-intel

MarioCarrilloA:
oh if all metrics tests should run with the KSM auto stuff disabled, then I like the idea @grahamwhaley

.ci/install_proxy.sh
```diff
@@ -23,10 +23,24 @@ source "${cidir}/lib.sh"
 
 clone_build_and_install "github.com/clearcontainers/proxy"
 start_proxy_cmd="sudo systemctl start cc-proxy"
+upstart_source_file="${cidir}/data/cc-proxy.conf"
+upstart_destr_file="/etc/init/cc-proxy.conf"
```
chavafg: typo: s/upstart_destr_file/upstart_dest_file/

.ci/install_proxy.sh
```diff
@@ -23,10 +23,24 @@ source "${cidir}/lib.sh"
...
+
+
+# If we are running under the metrics CI system then we do not want the proxy
+# to be dynmaically changing the KSM settings under us - we need control of them
+# ourselves
+if [[ $METRICS_CI ]]; then
+	proxy_systemd_file=$(sudo systemctl show cc-proxy | fgrep FragmentPath | awk 'BEGIN{FS="="}{print $2}')
+	sudo sed -i 's/\(^ExecStart.*$\)/\1 -ksm initial/' "${proxy_systemd_file}"
```
chavafg: what about also adding `-log debug` here?  Just as in `cc-proxy.conf.noksm` file

grahamwhaley: I did notice that we had `-log` in the upstart files we inject, but did not inject it into the systemd files installed by the package already - and wondered if that disparity was deliberate or not?
As the current code did not add `-log` to the systemd files, I went with the same methodology.
@chavafg - what do you think, maybe we should always add `-log` for both upstart and systemd - in which case I'm happy to make the change

grahamwhaley:
dropped RFC and WIP. Pushed with typo fix. Will wait for any feedback on the `-log` item.

.ci/install_proxy.sh
chavafg: Hi @grahamwhaley , I think we didn't added the `-log` option simply because we forgot it and yes, we should always add the option. Please make the change :) 
Thanks.

grahamwhaley: @chavafg added and pushed - please review, as not run-tested here and I suspect CI may only run-test half of the code (not sure if we cover both upstart and systemd).
Note, sed expr param changed from 'x' to "x" as we now use an expansion to build the args.

chavafg:
lgtm

PR #566, opened 2017-09-22
Build runtime as documented

jodh-intel:


jodh-intel:
Both Jenkins builds are showing the correct value for `Clear Containers system build` now.

PR #570, opened 2017-09-25
Openshift: fix tests

chavafg:
This PR contains three commits that fix the Openshift tests.

- Specify `cgroupfs` as the `cgroup-driver` that kubelet will use.
- Call `cc-runtime` using full path.
- Add message when openshift finishes successfully


chavafg:
Can someone take a look at this PR? It is needed to add openshift tests on the CI

devimc:
lgtm

Issue #569, opened 2017-09-25
ci: metrics: make install should run as sudo


grahamwhaley:
In the run metrics script, we build and install some local tools if they do not exist before we run - but, we don't do the install step with root privs (unless the whole script is being run as root), so it fails as the default dest install dir is the system folder `/usr/bin`.
https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh#L31
Fix by adding a `sudo` on the front of the `make install` - like we have `sudo` for other parts of the script that require raised privs.

PR #571, opened 2017-09-25
ci: metrics: run 'make install' with raised privs

grahamwhaley:
The metrics ci script should not have to be run as root, as it uses
sudo to raise privs when necessary - but, it invokes a 'make install'
if there are missing commands, and that needs privs to install into
the default system dir paths. Add the relevant `sudo`.

Fixes: #569


jodh-intel:
lgtm

Issue #572, opened 2017-09-25
metrics: emailreport: is not vendored


grahamwhaley:
The email report tool is not 'vendored':
https://github.com/clearcontainers/tests/tree/master/cmd/emailreport

which means it can fail to build when running automation such as:
https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh#L30
when the package dependencies are not installed already on the system.

Some details of our vendoring method/tools can be found on the pending Issue at:
https://github.com/clearcontainers/runtime/issues/332

Any issues, shout here and we will try to help.
/cc @sboeuf

jodh-intel:
Worth pointing out that `emailreport` also lacks a `--version` flag and/or version print at runtime which could be useful for debugging.

Looks like `fetchbranches`, `checkmetrics` are also missing version details.

MarioCarrilloA:
Hi @jodh-intel , good point I will add another issue for missing `--version` flag

PR #574, opened 2017-09-25
cmd: emailreport: Add vendor dependencies

MarioCarrilloA:
This tool has not vendor dependencies,
this commit adds them.

Fixes: #572

Signed-off-by: Arevalo, Mario Alfredo C <mario.alfredo.c.arevalo@intel.com>


cmd/emailreport/Gopkg.lock
```diff
@@ -0,0 +1,15 @@
+# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
+
+
+[[projects]]
+  name = "github.com/BurntSushi/toml"
+  packages = ["."]
+  revision = "b26d9c308763d68093482582cea63d69be07a0f0"
```
grahamwhaley: until fairly recently `toml` used a license (the WTF license) that we'd rather not have in our project - but, then they updated it :-) Can we thus move the revision/version of `toml` we are tracking, maybe to be the same as that in `checkmetrics` for instance:
https://github.com/clearcontainers/tests/blob/master/cmd/checkmetrics/Gopkg.toml#L24

In fact, generally I believe for our projects, we set `dep` to track the `master` branches rather than fixed versions (unless we have to pin/regress to fix an issue for instance).

grahamwhaley:
I've triggered a rebuild of the Jenkins 16.04 as the output looked like it worked, but somehow failed to register the results back to the Jenkins master - maybe something was in flux - let's see how this cycle goes...

MarioCarrilloA:
@grahamwhaley done :)

PR #584, opened 2017-10-02
.ci: Add shared script for Jenkins jobs

### LINKED ISSUE #583: ci: Add a shared script for Jenkins jobs
Need a shared script to be shared across different Jenkins jobs since they are all very similar.

sboeuf:
All jenkins jobs are very similar since they end up calling into
setup.sh and run.sh. This commit introduces a script that can be
shared across all those jobs/repos.

Fixes #583

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+# Need the repo to know which tests to run.
+cc_repo=$1
+
+tests_repo="github.com/clearcontainers/tests"
+mkdir -p ${HOME}/go
+
+# Export all environment variables needed.
+export GOROOT="/usr/local/go"
+export GOPATH=${HOME}/go
+export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:${PATH}
+export CI=true
+
+# Download and build goveralls binary in case we need to submit the code
+# coverage.
+if [ ${COVERALLS_REPO_TOKEN} ]
+then
+	go get github.com/mattn/goveralls
+fi
+
+# Get the repository and move HEAD to the appropriate commit.
+go get ${cc_repo} || true
+pushd ${GOPATH}/src/${cc_repo}
+if [ ${ghprbPullId} ] && [ ${ghprbTargetBranch} ]
+then
+	git fetch origin pull/${ghprbPullId}/head && git checkout FETCH_HEAD && git rebase origin/${ghprbTargetBranch}
+else
+	git fetch origin && git checkout origin/master
+fi
+
+# Setup and run the tests.
+if [ "${cc_repo}" == "${tests_repo}" ]
+then
+        .ci/setup_tests.sh
+fi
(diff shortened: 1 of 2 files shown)


.ci/jenkins_job_build.sh
```diff
@@ -0,0 +1,61 @@
...
+# Download and build goveralls binary in case we need to submit the code
+# coverage.
+if [ "${COVERALLS_REPO_TOKEN}" != "" ]
+then
+	go get github.com/mattn/goveralls
+fi
+
+# Get the repository and move HEAD to the appropriate commit.
+go get ${cc_repo} || true
+cd ${GOPATH}/src/${cc_repo}
```
MarioCarrilloA: I am not sure if it is necessary to come back to previous dir, if yes, could be better to use `pushd/popd`, and maybe for maintainability.

.ci/jenkins_job_build.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Export all environment variables needed.
+export GOROOT=/usr/local/go
+export GOPATH=${HOME}/go
+export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:${PATH}
+export CI=true
+
+# Download and build goveralls binary in case we need to submit the code
+# coverage.
+if [ "${COVERALLS_REPO_TOKEN}" != "" ]
```
MarioCarrilloA: This is just a comment regarding to style, I have noticed that other scripts inside `.ci` uses `if [ x ]; then`

.ci/jenkins_job_build.sh
```diff
@@ -0,0 +1,61 @@
...
+set -e
+
+# Need the repo to know which tests to run.
+cc_repo=$1
+
+tests_repo="github.com/clearcontainers/tests"
+mkdir -p ${HOME}/go
+
+# Export all environment variables needed.
+export GOROOT=/usr/local/go
```
MarioCarrilloA: maybe could be good to use double quotes

sboeuf: done

.ci/jenkins_job_build.sh
sboeuf: done

.ci/jenkins_job_build.sh
sboeuf: yes I think `pushd/popd` is a good idea. Done

sameo:
lgtm

.ci/jenkins_job_build.sh
```diff
@@ -0,0 +1,63 @@
...
+# Need the repo to know which tests to run.
+cc_repo=$1
+
+tests_repo="github.com/clearcontainers/tests"
+mkdir -p ${HOME}/go
+
+# Export all environment variables needed.
+export GOROOT="/usr/local/go"
+export GOPATH=${HOME}/go
+export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:${PATH}
```
grahamwhaley: `/usr/local/go/bin` should probably be `${GOROOT}/bin`

.ci/jenkins_job_build.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Download and build goveralls binary in case we need to submit the code
+# coverage.
+if [ ${COVERALLS_REPO_TOKEN} ]
+then
+	go get github.com/mattn/goveralls
+fi
+
+# Get the repository and move HEAD to the appropriate commit.
+go get ${cc_repo} || true
```
grahamwhaley: you might want `go get -d` here and to drop the `|| true` ?

sboeuf: `go get -d` never worked, that's why our script are all using `|| true`.

.ci/jenkins_job_build.sh
sboeuf: yep you're right.

.ci/jenkins_job_build.sh
grahamwhaley: hmm, do we know if there is a reason why - it would be nice to fix that. We don't have to block due to it though. /cc @jodh-intel for any thoughts/input

jodh-intel: Huh? `go get -d` never worked? How can that not work?

PR #588, opened 2017-10-03
openshift: Download binaries instead of building them

### LINKED ISSUE #587: Download Openshift binaries instead of building them
To reduce time for our CI, we should download the binaries, instead of building from sources.

chavafg:
This change reduces the time for testing the OpenShift integration.
Instead of clonning the repository and build everything with sources,
we download the binaries they provide in their releases.

Fixes #587.

### DIFF
--- a/integration/openshift/setup.sh
+++ b/integration/openshift/setup.sh
@@ -49,13 +49,15 @@ make binary-local
 sudo -E make install-binary
 popd
 
-echo "Install Openshift"
+echo "Install Openshift Origin"
 openshift_repo="github.com/openshift/origin"
-go get -d "$openshift_repo" || true
-pushd "$GOPATH/src/$openshift_repo"
-git checkout "$origin_version"
-export OS_OUTPUT_GOPATH=1
-make
-sudo cp _output/local/bin/linux/amd64/{openshift,oc,oadm} /usr/bin
-popd
+openshift_tarball="openshift-origin-server-${origin_version}-${origin_commit}-linux-64bit.tar.gz"
+openshift_dir="${openshift_tarball/.tar.gz/}"
+openshift_url="https://${openshift_repo}/releases/download/${origin_version}/${openshift_tarball}"
+
+curl -L -O "$openshift_url"
+tar -xf "$openshift_tarball"
+sudo install ${openshift_dir}/{openshift,oc,oadm} /usr/bin
+rm -rf "$openshift_dir" "${openshift_tarball}"
+
 echo "Openshift + CC Setup finished successfully"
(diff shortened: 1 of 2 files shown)


sboeuf:
LGTM

integration/openshift/setup.sh
```diff
@@ -51,11 +51,13 @@ popd
...
 echo "Install Openshift"
 openshift_repo="github.com/openshift/origin"
-go get -d "$openshift_repo" || true
-pushd "$GOPATH/src/$openshift_repo"
-git checkout "$origin_version"
-export OS_OUTPUT_GOPATH=1
-make
-sudo cp _output/local/bin/linux/amd64/{openshift,oc,oadm} /usr/bin
-popd
+openshift_tarball="openshift-origin-server-${origin_version}-c4dd4cf-linux-64bit.tar.gz"
```
grahamwhaley: I think the `c4dd4cf` is part of, or at least tied to, the `origin_version` - so probably needs to be a var or part of that var

chavafg: Right, I added a new var called `origin_commit`

integration/openshift/setup.sh
```diff
@@ -49,13 +49,15 @@ make binary-local
...
-make
-sudo cp _output/local/bin/linux/amd64/{openshift,oc,oadm} /usr/bin
-popd
+openshift_tarball="openshift-origin-server-${origin_version}-${origin_commit}-linux-64bit.tar.gz"
+openshift_dir="${openshift_tarball/.tar.gz/}"
+openshift_url="https://${openshift_repo}/releases/download/${origin_version}/${openshift_tarball}"
+
+curl -L -O "$openshift_url"
+tar -xvf "$openshift_tarball"
+sudo cp ${openshift_dir}/{openshift,oc,oadm} /usr/bin
```
jodh-intel: `cp` or `install`? This is relying on the archive file having the correct perms on these binaries (but that probably isn't an unreasonable thing to do for a pre-built release archive :).

devimc: imo ```install``` is better

integration/openshift/setup.sh
```diff
@@ -49,13 +49,15 @@ make binary-local
...
-export OS_OUTPUT_GOPATH=1
-make
-sudo cp _output/local/bin/linux/amd64/{openshift,oc,oadm} /usr/bin
-popd
+openshift_tarball="openshift-origin-server-${origin_version}-${origin_commit}-linux-64bit.tar.gz"
+openshift_dir="${openshift_tarball/.tar.gz/}"
+openshift_url="https://${openshift_repo}/releases/download/${origin_version}/${openshift_tarball}"
+
+curl -L -O "$openshift_url"
+tar -xvf "$openshift_tarball"
```
devimc: ```v``` is not needed

integration/openshift/setup.sh
chavafg: Agree, changed to install.

devimc:
lgtm

PR #597, opened 2017-10-05
ci: integrate OpenShift setup and test execution into the CI process 

chavafg:
This change will allow to setup Openshift environment and execute tests 
through the CI if envar OPENSHIFT_CI=true. The purpose of this change is
to make minimal changes in our jenkins setup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,12 @@ metrics:
 integration: ginkgo
 	./ginkgo ./integration/docker/ -- -timeout ${TIMEOUT}
 
+openshift:
+	bash .ci/install_bats.sh
+	cd integration/openshift && \
+	./init.sh && \
+	bats hello_world.bats
+
 swarm:
 	bats integration/swarm/swarm.bats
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -39,6 +39,13 @@ metrics:
 integration: ginkgo
 	./ginkgo ./integration/docker/ -- -timeout ${TIMEOUT}
 
+openshift:
+	cd integration/openshift
+	./init.sh
+	# Sleep 30 seconds to let the master and node initialize correctly
+	sleep 30
```
jodh-intel: This looks a bit odd and potentially fragile. Is there no way to check when the nodes have initialised fully (a command to run, particular log message(s) to look for, etc)?

chavafg: not sure, but will take a look

chavafg: changed the approach to verify if the node is up every 5 secs until a maximum of 30s.

jodh-intel:
lgtm

devimc:
lgtm

PR #609, opened 2017-10-13
.ci: Copy logs instead of displaying them

### LINKED ISSUE #608: ci: Don't dump logs at teardown if a destination path is provided
The logs generated by the different components are pretty huge, and we don't want them to be included with the default console output in Jenkins. Otherwise the browser has to process more than 70MiB of logs to display...
Instead we should copy those logs and retrieve them as files directly.

Related to https://github.com/clearcontainers/jenkins/issues/7

sboeuf:
If a destination directory is provided to the teardown.sh script,
it will now copy the logs instead of displaying them through the
standard output. By default, if no path is provided, we keep the
original behavior, meaning we're dumping the logs.

Also, because our CI is not testing on Ubuntu 14.04 anymore,
there is no reason for keeping the upstart case handled by our
scripts.

Fixes #608

### DIFF
--- a/.ci/teardown.sh
+++ b/.ci/teardown.sh
@@ -17,22 +17,53 @@
 # This script will get any execution log that may be useful for
 # debugging any issue related to Clear Containers.
 
-runtime_log_file="/var/lib/clear-containers/runtime/runtime.log"
+log_copy_dest="$1"
 
-echo "Clear Containers Runtime Log:"
-sudo cat "$runtime_log_file"
+runtime_log_location="/var/lib/clear-containers/runtime/runtime.log"
+runtime_log_filename="cc-runtime.log"
+runtime_log_path="${log_copy_dest}/${runtime_log_filename}"
 
-if [[ ! $(ps -p 1 | grep systemd) ]]; then
-	upstart_logs_path="/var/log/upstart"
-	echo "Clear Containers Proxy Log:"
-	sudo cat "${upstart_logs_path}/cc-proxy.log"
-	echo "CRI-O Log:"
-	sudo cat "${upstart_logs_path}/crio.log"
+proxy_log_filename="cc-proxy.log"
+proxy_log_path="${log_copy_dest}/${proxy_log_filename}"
+
+shim_log_filename="cc-shim.log"
+shim_log_path="${log_copy_dest}/${shim_log_filename}"
+
+crio_log_filename="crio.log"
+crio_log_path="${log_copy_dest}/${crio_log_filename}"
+
+upstart_logs_path="/var/log/upstart"
+upstart_proxy_path="${upstart_logs_path}/${proxy_log_filename}"
+upstart_crio_path="${upstart_logs_path}/${crio_log_filename}"
+
+# Copy log files if a destination path is provided, otherwise simply
+# display them.
+if [ ${log_copy_dest} ]; then
+	sudo cp "${runtime_log_location}" "${runtime_log_path}"
+
+	if [[ -n $(ps -p 1 | grep systemd) ]]; then
+		sudo journalctl --no-pager -u cc-proxy > "${proxy_log_path}"
+		sudo journalctl --no-pager -t cc-shim > "${shim_log_path}"
+		sudo journalctl --no-pager -u crio > "${crio_log_path}"
+	else
+		sudo cp "${upstart_proxy_path}" > "${proxy_log_path}"
+		sudo cp "${upstart_crio_path}" > "${crio_log_path}"
+	fi
 else
-	echo "Clear Containers Proxy Log:"
-	sudo journalctl --no-pager -u cc-proxy
-	echo "Clear Containers Shim Log:"
-	sudo journalctl --no-pager -t cc-shim
-	echo "CRI-O Log:"
-	sudo journalctl --no-pager -u crio
+	echo "Clear Containers Runtime Log:"
+	sudo cat "${runtime_log_location}"
+
+	if [[ -n $(ps -p 1 | grep systemd) ]]; then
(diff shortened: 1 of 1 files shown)


.ci/teardown.sh
```diff
@@ -17,22 +17,50 @@
...
-	sudo cat "${upstart_logs_path}/crio.log"
+
+# Copy log files if a destination path is provided, otherwise simply
+# display them.
+if [ ${log_copy_dest} ]; then
+	sudo cp "${runtime_log_location}" "${runtime_log_path}"
+
+	if [[ ! $(ps -p 1 | grep systemd) ]]; then
+		sudo cp "${upstart_proxy_log_path}" "${proxy_log_path}"
+		sudo cp "${upstart_crio_log_path}" "${crio_log_path}"
```
jodh-intel: We should be able to extract the shim debug from `/var/log/syslog` or similar I think?

.ci/teardown.sh
```diff
@@ -17,22 +17,50 @@
...
 
-runtime_log_file="/var/lib/clear-containers/runtime/runtime.log"
+log_copy_dest=$1
 
-echo "Clear Containers Runtime Log:"
-sudo cat "$runtime_log_file"
+runtime_log_location="/var/lib/clear-containers/runtime/runtime.log"
+runtime_log_filename="cc-runtime.log"
+runtime_log_path="${log_copy_dest}/${runtime_log_filename}"
+proxy_log_filename="cc-proxy.log"
```
jodh-intel: Could you add blank lines between each block (runtime, proxy, shim, etc) to make this a little easier to read?

.ci/teardown.sh
```diff
@@ -17,22 +17,50 @@
 # This script will get any execution log that may be useful for
 # debugging any issue related to Clear Containers.
 
-runtime_log_file="/var/lib/clear-containers/runtime/runtime.log"
+log_copy_dest=$1
```
jodh-intel: Best to quote `$1` for safety.

.ci/teardown.sh
```diff
@@ -17,22 +17,50 @@
...
-	sudo cat "${upstart_logs_path}/cc-proxy.log"
-	echo "CRI-O Log:"
-	sudo cat "${upstart_logs_path}/crio.log"
+
+# Copy log files if a destination path is provided, otherwise simply
+# display them.
+if [ ${log_copy_dest} ]; then
+	sudo cp "${runtime_log_location}" "${runtime_log_path}"
+
+	if [[ ! $(ps -p 1 | grep systemd) ]]; then
```
jodh-intel: I appreciate you're modifying what was already there, but I'd remove the `!` and switch the blocks around as both cases are covered and this reverse logic isn't necessary:

```
if [[ -n $(ps -p 1 | grep systemd) ]]; then
  # systemd
else
  # upstart
fi
```

.ci/teardown.sh
```diff
@@ -17,22 +17,50 @@
...
-	echo "CRI-O Log:"
-	sudo journalctl --no-pager -u crio
+	echo "Clear Containers Runtime Log:"
+	sudo cat "${runtime_log_location}"
+
+	if [[ ! $(ps -p 1 | grep systemd) ]]; then
+		echo "Clear Containers Proxy Log:"
+		sudo cat "${upstart_proxy_log_path}"
+		echo "CRI-O Log:"
+		sudo cat "${upstart_crio_log_path}"
```
jodh-intel: Again, missing shim details.

jodh-intel:
Hi @sboeuf - thanks for raising - this is an essential change given the amount of log data we're generating now.

.ci/teardown.sh
sboeuf: done

.ci/teardown.sh
sboeuf: done

.ci/teardown.sh
sboeuf: done

sboeuf:
@jodh-intel I have addressed your comments and I have removed the code specific to upstart (Ubuntu 14.04) as we have discussed on IRC.

sboeuf:
@grahamwhaley yes you're right, I am gonna split this as two separate patches and mention it on the PR.

sboeuf:
@grahamwhaley I have updated as asked ;)

.ci/teardown.sh
```diff
@@ -17,22 +17,53 @@
...
+if [ ${log_copy_dest} ]; then
+	sudo cp "${runtime_log_location}" "${runtime_log_path}"
+
+	if [[ -n $(ps -p 1 | grep systemd) ]]; then
+		sudo journalctl --no-pager -u cc-proxy > "${proxy_log_path}"
+		sudo journalctl --no-pager -t cc-shim > "${shim_log_path}"
+		sudo journalctl --no-pager -u crio > "${crio_log_path}"
+	else
+		sudo cp "${upstart_proxy_path}" > "${proxy_log_path}"
+		sudo cp "${upstart_proxy_path}" > "${crio_log_path}"
```
grahamwhaley: Is this meant to be crio_path, not proxy_path? (copy/paste error?). Or at least not copying the proxy path to the crio log path :-)

sboeuf: oh yeah I did a mistake here !

sboeuf:
@grahamwhaley yeah you're right I could have gone the other way, but I don't want to change my PR again. If this latest update works for you, please merge it ;)

chavafg:
lgtm

sboeuf:
@grahamwhaley please merge this if that's okay for you, I'll update the jenkins jobs accordingly ;)

PR #514, opened 2017-09-06
CI: Install packages from clearlinux.org instead of OBS

chavafg:
This change install clear-containers-image, linux-container and
qemu-lite packages from download.clearlinux.org using `alien` as
the packages in the site are RPM.

This change is to not rely on OBS platform as sometimes the service
is down.

Fixes #513.

Signed-off-by: Salvador Fuentes <salvador.fuentes@intel.com>

PR #610, opened 2017-10-13
CI: Install packages from clearlinux.org instead of OBS (fedora)

### LINKED ISSUE #604: CI: fedora Install packages from OBS
fedora ci script installs clear-containers-image, linux-container and
qemu-lite packages from OBS which is wrong. We need to install from download.clearlinux.org 

This change is required to not rely on OBS platform as sometimes the service
is down.

gorozco1:
This change install clear-containers-image, linux-container and
qemu-lite packages from download.clearlinux.org.

This change is to not rely on OBS platform as sometimes the service
is down.

Fixes #604

Signed-off-by: Geronimo Orozco <geronimo.orozco@intel.com>

### DIFF
--- a/.ci/install_qemu_lite.sh
+++ b/.ci/install_qemu_lite.sh
@@ -16,14 +16,15 @@
 
 set -e
 
-if [ "$#" -ne 2 ]; then
-    echo "Usage: $0 <CLEAR_RELEASE> <QEMU_LITE_VERSION>"
+if [ "$#" -ne 3 ]; then
+    echo "Usage: $0 <CLEAR_RELEASE> <QEMU_LITE_VERSION> <DISTRO>"
     echo "       Install the QEMU_LITE_VERSION from clear CLEAR_RELEASE."
     exit 1
 fi
 
 clear_release="$1"
 qemu_lite_version="$2"
+distro="$3"
 qemu_lite_bin="qemu-lite-bin-${qemu_lite_version}.x86_64.rpm"
 qemu_lite_data="qemu-lite-data-${qemu_lite_version}.x86_64.rpm"
 
@@ -33,9 +34,14 @@ echo -e "Install qemu-lite ${qemu_lite_version}"
 curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/x86_64/os/Packages/${qemu_lite_bin}"
 curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/x86_64/os/Packages/${qemu_lite_data}"
 
-# install packages using alien
-sudo alien -i "./${qemu_lite_bin}"
-sudo alien -i "./${qemu_lite_data}"
+# install packages
+if [ "$distro" == "fedora" ]; then
+    sudo rpm -ihv "./${qemu_lite_bin}" --nodeps
+    sudo rpm -ihv "./${qemu_lite_data}" --nodeps
+elif [ "$distro" == "ubuntu" ];  then
+    sudo alien -i "./${qemu_lite_bin}"
+    sudo alien -i "./${qemu_lite_data}"
+fi
 
 # cleanup
 rm -f "./${qemu_lite_bin}"
--- a/.ci/setup_env_fedora.sh
+++ b/.ci/setup_env_fedora.sh
@@ -19,6 +19,7 @@ set -e
 cidir=$(dirname "$0")
 source "$cidir/../test-versions.txt"
 source /etc/os-release
+cc_kernel_path="/usr/share/clear-containers"
 
 if grep -q "N" /sys/module/kvm_intel/parameters/nested; then
 	echo "enable Nested Virtualization"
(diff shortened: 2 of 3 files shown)


.ci/install_qemu_lite.sh
```diff
@@ -34,8 +35,13 @@ curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/x86_64
 curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/x86_64/os/Packages/${qemu_lite_data}"
 
 # install packages using alien
-sudo alien -i "./${qemu_lite_bin}"
-sudo alien -i "./${qemu_lite_data}"
+if [ "$distro" == fedora ]; then
```
sboeuf: I think it could be better for bash syntax to write:
```bash
if [ "$distro" == "fedora" ]; then
```

.ci/install_qemu_lite.sh
```diff
@@ -34,8 +35,13 @@ curl -LO "https://download.clearlinux.org/releases/${clear_release}/clear/x86_64
...
+    rpm -ihv "./${qemu_lite_bin}" --nodeps
+    rpm -ihv "./${qemu_lite_data}" --nodeps
+elif [ "$distro" == ubuntu ];  then
```
sboeuf: Ditto

sboeuf:
LGTM

jodh-intel:
Hi @gorozco1 - If we can't rely on OBS, we need to explore other options but I'm not wholly convinced about using `alien` for Debian-based systems: we should be building and installing distro packages using  the standard tooling for each distro.

Do you have any further details on the OBS service being "down"? Do you mean fully inaccessible? Or just that the OBS files are not available from the site for periods of time? It looks like https://github.com/clearcontainers/runtime/issues/724 may been an instance of what you are describing. Could it be that we are being hit by scheduled OBS outages, or is this seemingly random downtime?

If 724 is the same problem you describe I think we need a general solution (for all users), not just one for the tests imho.

.ci/setup_env_fedora.sh
```diff
@@ -27,23 +28,20 @@ if grep -q "N" /sys/module/kvm_intel/parameters/nested; then
...
-
 sudo -E dnf makecache
 
 echo "Install clear containers dependencies"
 sudo -E dnf -y groupinstall "Development tools"
 sudo -E dnf -y install libtool automake autoconf bc
 
 echo "Install qemu-lite binary"
-sudo -E dnf -y install qemu-lite
+"${cidir}/install_qemu_lite.sh" "${qemu_clear_release}" "${qemu_lite_sha}" "fedora"
```
chavafg: what about changing `"fedora"` with `"$ID"` as we already sourced `/etc/os-release`

.ci/setup_env_ubuntu.sh
```diff
@@ -37,7 +37,7 @@ echo "Install clear containers dependencies"
 chronic sudo -E apt install -y libtool automake autotools-dev autoconf bc alien libpixman-1-dev
 
 echo "Install qemu-lite binary"
-"${cidir}/install_qemu_lite.sh" "${qemu_clear_release}" "${qemu_lite_sha}"
+"${cidir}/install_qemu_lite.sh" "${qemu_clear_release}" "${qemu_lite_sha}" "ubuntu"
```
chavafg: same as above

chavafg:
Seems like you forgot to add a dependency here. From: http://cc-jenkins-ci.westus2.cloudapp.azure.com/job/clear-containers-tests-fedora-26-PR/62/consoleText

```
time="2017-10-16T15:47:45.384315804Z" pid=68663 name="cc-runtime" level="error" source="virtcontainers" msg="Unable to launch qemu: exit status 127"
time="2017-10-16T15:47:45.409541818Z" pid=68663 name="cc-runtime" level="error" source="virtcontainers" msg="/usr/bin/qemu-lite-system-x86_64: error while loading shared libraries: libpixman-1.so.0: cannot open shared object file: No such file or directory\n"
time="2017-10-16T15:47:55.326487968Z" pid=68663 name="cc-runtime" level="error" source="runtime" msg="Did not receive the pod started notification (timeout 10000000000s)"
```

gorozco1:
@jodh-intel That ship sailed a month ago :P #514

gorozco1:
@chavafg updated PR

jodh-intel:
@gorozco1 - Yikes! I must have missed that. Well, we're going to have to be careful here since if we start using more dpkg-specific features for example, the CI won't be running them and so won't be representative of what users are installing. I'm thinking of things like dpkg's maintainer scripts which might at some point be required to tweak upgrade behaviour or similar.

jcvenegas:
@gorozco1 I agree @jodh-intel , If the qemu from Clear Linux works, I am not totally if this will have a good behavior the binary is linked to things provided in Clear Linux. I don't think the the image and kernel are a problem those are components that will run as part of the guest not a program in the ubuntu/fedora system.

gorozco1:
@jcvenegas the tests pass

jcvenegas:
@gorozco1 I did not seen that qemu is already installed in the ubuntu setup. I guess we could merge to have parity with it. @jodh-intel  @gorozco1  as alternative we could try to compile qemu-lite as an static binary and provide it in github using a tag/release.

chavafg:
lgtm, opened issue #635 to address @jodh-intel concern and @jcvenegas recommendation. 
merging this one.

Issue #611, opened 2017-10-14
cmd: checkmetrics: it forces to use a basefile for working


MarioCarrilloA:
checkmetrics tools should provide the option of "parse" the metrics results without a basefile, this is in order to get a fast overview about results.

PR #612, opened 2017-10-14
cmd: checkmetrics: Add nobaseline option

MarioCarrilloA:
This option allows to parse the metrics CSV files without
a baseline file, the report about these metrics just
includes the mean, the standard deviation and the number
of samples used for those results.

Fixes: #611

### DIFF
--- a/cmd/checkmetrics/README.md
+++ b/cmd/checkmetrics/README.md
@@ -47,7 +47,7 @@ maxval = 1.5
 ## Options
 `checkmetrics` takes a number of options. Some are mandatory.
 
-### TOML basefile path (mandatory)
+### TOML basefile path (mandatory unless `nobaseline` is used)
 
 ```
 --basefile value    path to baseline TOML metrics file
@@ -61,6 +61,11 @@ maxval = 1.5
 ### Log file path
 ```
 --log value         set the log file path
+```
+
+### No baseline
+```
+--nobaseline        enable parsing metrics without a basefile
 ```
 
 ### Metrics CSV directory path
--- a/cmd/checkmetrics/csv.go
+++ b/cmd/checkmetrics/csv.go
@@ -39,6 +39,7 @@ type csvRecord struct {
 	MaxVal        float64   // Largest value we saw
 	SD            float64   // Standard Deviation
 	CoV           float64   // Co-efficient of Variation
+	TestName      string    // Test name field
 }
 
 // load reads in a CSV 'Metrics' results file from the file path given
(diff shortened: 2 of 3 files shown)


chavafg:
lgtm

grahamwhaley:
@MarioCarrilloA I like the idea, that we can then do stats over results and data (rather than having to hand craft some scripts or import into a spreadsheet etc. every time) - but, I think we can lay the code out better.
It looks to me like `StatMetrics` should be broken up/re-factored, as that func pretty much looks like a copy of `load()`, and then carries some 'table logic' into it as well.
I'd be much happier to see your new `processMetrics` in main.go to be calling `load()` to get the `csvRecord`s out, and then forming the table directly - that way we avoid duplicating most of the `load()` code, and we lift all the table formatting and string processing knowledge out into main.go.
What do you think?

MarioCarrilloA:
hey @grahamwhaley , I agree with that feedback, expect a version 2 :)

MarioCarrilloA:
hi @grahamwhaley thanks for suggestion, new version done

MarioCarrilloA:
Hi @iphutch could you please take a look :)

cmd/checkmetrics/README.md
```diff
@@ -47,7 +47,7 @@ maxval = 1.5
 ## Options
 `checkmetrics` takes a number of options. Some are mandatory.
 
-### TOML basefile path (mandatory)
+### TOML basefile path (mandatory unless nobaseline is used)
```
iphutch: use literal text for `nobaseline`

MarioCarrilloA: looks better, thanks @iphutch

cmd/checkmetrics/csv.go
```diff
@@ -75,6 +76,9 @@ func (c *csvRecord) load(filepath string) error {
 	// The Results column for the CSV metrics files is expected to be in the 5th column
 	const resultsColumn = 4
 
+	// The Test Name column for the CSV metrics files is expected to be in the 3th column
```
grahamwhaley: s/3th/3rd/

cmd/checkmetrics/main.go
```diff
@@ -128,6 +133,45 @@ func processMetrics(context *cli.Context) (err error) {
...
+	for _, csvf := range csvFiles {
+		var thisCsv csvRecord
+		filepath := path.Join(mtrdir, csvf.Name())
+		err = thisCsv.load(filepath)
+		if  err != nil {
+			return err
+		}
+
+		row := []string{thisCsv.TestName,
+			strconv.FormatFloat(thisCsv.Mean, 'f', -1, 64),
```
grahamwhaley: Might be nice to have the Min and Max here as well, as those are what we actually code back into the baseline toml files - what do you think?

MarioCarrilloA: make sense, I will send a new version :)

MarioCarrilloA:
@grahamwhaley @iphutch done, thanks :)

Issue #617, opened 2017-10-17
metrics: network: missing tests between a container and the host


MarioCarrilloA:
There are not metrics tests that can give us information between a container and the host

PR #618, opened 2017-10-17
metrics: network: Add host/container test

MarioCarrilloA:
This commits adds a test that measures the bandwidth
between a client/server interconnection, where
the container takes the role of server and the host
takes the role of client.

Fixes: #617

Signed-off-by: Arevalo, Mario Alfredo C <mario.alfredo.c.arevalo@intel.com>

### DIFF
--- a/metrics/network/network-metrics-iperf3.sh
+++ b/metrics/network/network-metrics-iperf3.sh
@@ -21,9 +21,12 @@
 # - bandwith duplex
 # - jitter
 #
-# These metrics/results will be got from the interconnection between a container
-# server and client using the iperf tool.
-# container-server <----> container-client
+# These metrics/results will be got from the interconnection between
+# a client and a server using iperf tool.
+# The following cases are covered:
+#
+# case 1:
+#  container-server <----> container-client
 
 set -e
 
@@ -43,13 +46,22 @@ transmit_timeout=5
 # "privileged" argument enables access to all devices on
 # the host and it allows to avoid conflicts with AppArmor
 # or SELinux configurations.
-client_extra_args="-ti --privileged --rm"
-server_extra_args="--privileged"
+if [ "$RUNTIME" == "runc" ]; then
+	extra_capability="--privileged"
+fi
+
+# Client/Server extra configuration
+client_extra_args="-ti $extra_capability --rm"
+server_extra_args="$extra_capability"
+
+# Iperf server configuration
+mount_ramfs="mount -t ramfs -o size=20M ramfs /tmp"
+server_command="$mount_ramfs && iperf3 -s"
+
 
 # Test single direction TCP bandwith
 function iperf3_bandwidth() {
 	local test_name="network iperf bandwidth"
-	local server_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -p ${port} -s"
 	local server_address=$(start_server "$image" "$server_command" "$server_extra_args")
 
 	# Verify server IP address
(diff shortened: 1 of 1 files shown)


metrics/network/network-metrics-iperf3.sh
```diff
@@ -100,6 +106,48 @@ function iperf3_jitter() {
...
+	local init_time=1
+
+	# Verify server IP address
+	if [ -z "$server_address" ];then
+		clean_env
+		die "server: ip address no found"
+	fi
+
+	# This time is in order to wait the container initialization
+	sleep $init_time
```
jodh-intel: This isn't going to be reliable - can you run a command, or check a file, or `grep` for a log pattern (with a timeout) instead maybe?

metrics/network/network-metrics-iperf3.sh
```diff
@@ -100,6 +106,48 @@ function iperf3_jitter() {
...
+# Test/measure the bandwidth between the interconnection of a server/client.
+# A container takes the role of server and the host takes the role of client.
+function iperf3_host_container() {
+	local test_name="network host to container"
+
+	# Checks iperf3 tool installed in host
+	local cmds=("iperf3")
+	check_cmds "${cmds[@]}"
+
+	local server_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -s -V"
```
jodh-intel: General comment: all these tests are using the same mount args so it would make sense to define a variable for them to minimise the cost of maintaining the script. Also, I think there is scope to create a `start_iperf3` function that all the `iperf3` tests could use.

metrics/network/network-metrics-iperf3.sh
grahamwhaley: I agree, if we can find a way to wait/check (even if we have a sleep and then check - possibly in a loop), that would be better. Maybe we can do something with `netstat` or similar to check for the server open socket?

metrics/network/network-metrics-iperf3.sh
grahamwhaley: Agree, a re-factor could help here - @MarioCarrilloA are you game to re-factor some of the common parts?

MarioCarrilloA:
@jodh-intel and @grahamwhaley , agree, expect a new version :)

metrics/network/network-metrics-iperf3.sh
```diff
@@ -100,6 +106,48 @@ function iperf3_jitter() {
...
+
+	# Verify server IP address
+	if [ -z "$server_address" ];then
+		clean_env
+		die "server: ip address no found"
+	fi
+
+	# This time is in order to wait the container initialization
+	sleep $init_time
+	result=$(iperf3 -c $server_address -t $transmit_timeout)
```
mcastelino: We need to run the tests in both direction.
The RX direction Host sending to VM
```+	result=$(iperf3 -c $server_address -t $transmit_timeout)```
and the TX direction VM sending to host

```result=$(iperf3 -c $server_address -t $transmit_timeout) -R```

Also add multiqueue tests where you run the test with -P X.
Where X is 2,4,8

Comment (MarioCarrilloA) [reply metrics/network/network-metrics-iperf3.sh]: Hi @mcastelino thanks for feedback, I will add that

**Comment by MarioCarrilloA** [comment]
@grahamwhaley @jodh-intel @mcastelino done :)

File: metrics/network/network-metrics-iperf3.sh [hunk 4]
```diff
@@ -21,9 +21,15 @@
...
 #
-# These metrics/results will be got from the interconnection between a container
-# server and client using the iperf tool.
-# container-server <----> container-client
+# These metrics/results will be got from the interconnection between
+# a client and a server using iperf tool.
+# The following cases are covered:
+#
+# case 1:
+#  container-server <----> container-client
```
mcastelino: @MarioCarrilloA looks like a typo here. I assume you meant host-server

jodh-intel:
lgtm

metrics/network/network-metrics-iperf3.sh
MarioCarrilloA: This script includes multiple network tests using iperf with different configurations, one of them is `container-server<->container-client` and I have added some of them where the server is the container  `container-server<->host-client` there are not yet cases about `host-server<->container-client`

chavafg:
lgtm

Issue #619, opened 2017-10-18
Do not `systemctl restart` for each and every test


grahamwhaley:
Currently we `systemctl restart` both docker and the proxy for every init_env() call, which is pretty much for every test.
This is a bit 'heavy handed', and on my unbuntu 16.04 system it actually causes systemd to kill docker.service as it sees it as 'respawning too frequently' :-(

Modify the metrics tests so they restart docker and the proxy once before running the tests - we'd like to start on a 'known level footing', but restarting for every test is a bit heavy and has issues right now.

PR #620, opened 2017-10-18
metrics: Only systemctl restart once per cycle

grahamwhaley:
Rather than doing a systemctl restart on docker and the proxy
for every test invocation, lets just do it the once at the
start before we begin the test cycle.

Fix up the invocation scripts to make the one off call.

Predominanlty because running many frequent restarts on my
ubuntu machine results in systemd disabling docker.service
as it sees it as restarting too often.

Fixes: #619

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -65,6 +65,26 @@ function check_cmds()
 	done
 }
 
+# A one time (per uber test cycle) init that tries to get the
+# system to a 'known state' as much as possible
+function onetime_init()
+{
+	# The onetime init must be called once, and only once
+	if [ ! -z "$onetime_init_done" ]; then
+		die "onetime_init() called more than once"
+	fi
+
+	# Restart services
+	sudo systemctl restart docker
+	if [[ "${RUNTIME}" == "cor" || "${RUNTIME}" == "cc-runtime" ]];then
+		sudo systemctl restart cc-proxy
+	fi
+
+	# We want this to be seen in sub shells as well...
+	# otherwise init_env() cannot check us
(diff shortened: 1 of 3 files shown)


metrics/lib/common.bash
```diff
@@ -65,6 +65,26 @@ function check_cmds()
 	done
 }
 
+# A one time (per uber test cycle) init that tries to get the
+# system to a 'known state' as much as possible
+function onetime_init()
+{
+	# The onetime init must be called once, and only once
+	if [ ! -z "$onetime_init_done" ]; then
+		die "onetime_init() called more than once"
```
bwarden: Should we die, or just warn and return?

jodh-intel:
lgtm

metrics/lib/common.bash
grahamwhaley: Hi @bwarden . My theory is that if somebody is calling this more than once then they are potentially heading back to the situation we have right now, where we are calling the restart frequently, which upsets systemd. I'd rather enforce that it is only called once, and hence the `die`. A `warn` I suspect will just go ignored in the logs.

grahamwhaley: Ah, and I see what you mean - should we 'just warn and return', and be more benign. Hmm, yes, that would work, but I think I'd rather make it more blatant that somebody has done something they should not, rather than fairly benign (and if they do not spot the warning in the logs, they may continue under the misbelief that we have done something when actually we have not)

bwarden: Makes sense, lgtm.

chavafg:
lgtm

PR #626, opened 2017-10-19
20171019 clarify logs

grahamwhaley:
Add function and call from init_env to notify which tests we are running when,
and fix up relevant places in tests to set the TEST_NAME var and call init_env as early as practical
in each test.


jodh-intel:
lgtm

chavafg:
lgtm

PR #628, opened 2017-10-20
.ci: Split log files to avoid getting too large ones

### LINKED ISSUE #627: .ci: Split log files to 5 MiB smaller ones
In case of Jenkins, we need to export/move some log files generated on the VM. Jenkins has some issues doing this with large log files (10MiB+). The size of 5MiB has been chosen so that each file can be properly opened from a web browser without getting it crashing or freezing.

sboeuf:
In case of Jenkins CI, we need to have log files not too large
so that they can be properly moved from the VM to the Jenkins
server as build artifacts. Also, 5 MiB is a reasonable size to
check those logs from a web browser.

Notice that "sudo" has been removed from this script since none
of them were really needed. Moreover, the copy of the runtime
logs with "sudo" was preventing the regular user to modify the
file previously copied.

Fixes #627

### DIFF
--- a/.ci/teardown.sh
+++ b/.ci/teardown.sh
@@ -22,30 +22,45 @@ log_copy_dest="$1"
 runtime_log_location="/var/lib/clear-containers/runtime/runtime.log"
 runtime_log_filename="cc-runtime.log"
 runtime_log_path="${log_copy_dest}/${runtime_log_filename}"
+runtime_log_prefix="cc-runtime_"
 
 proxy_log_filename="cc-proxy.log"
 proxy_log_path="${log_copy_dest}/${proxy_log_filename}"
+proxy_log_prefix="cc-proxy_"
 
 shim_log_filename="cc-shim.log"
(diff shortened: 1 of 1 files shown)


.ci/teardown.sh
```diff
@@ -22,30 +22,43 @@ log_copy_dest="$1"
...
-	sudo journalctl --no-pager -u crio > "${crio_log_path}"
+	# Create the log files
+	cp "${runtime_log_location}" "${runtime_log_path}"
+	journalctl --no-pager -u cc-proxy > "${proxy_log_path}"
+	journalctl --no-pager -t cc-shim > "${shim_log_path}"
+	journalctl --no-pager -u crio > "${crio_log_path}"
+
+	# Split them in 5 MiB subfiles to avoid too large files.
+	pushd ${log_copy_dest}
+	split -b 5242880 -d ${runtime_log_path} ${runtime_log_prefix}
```
grahamwhaley: Can we add a var for the 5242880 at the top of the file pls - easier to edit later, saves the duplication here.

Issue #637, opened 2017-10-24
Integration test dont use CC_RUNTIME makefile variable


jcvenegas:
Functional tests uses CC_RUNTIME to know what runtime use. But integration test dont so if you run:

`make integration`

Will rely on the guess the cc-runtime is the default runtime in docker and was registered with as `cc-rumtime`

grahamwhaley:
Just to note, I believe the metrics currently use `RUNTIME` - we should probably define and unify this across all the tests at some point :-)

jcvenegas:
`RUNTIME` make more sense to me. It  should be possible to use any runtime. @chavafg do you know CC_RUNITME is being used as part of the validation somewhere ? 

A quick grep under clearcontainers repositories show that is used only here:
```
tests/Makefile:CC_RUNTIME ?= cc-runtime
tests/Makefile: RUNTIME=${CC_RUNTIME} ./test/test_runner.sh crio.bats
tests/Makefile: ./ginkgo -v functional/ -- -runtime ${CC_RUNTIME} -timeout ${TIMEOUT}
tests/Makefile: RUNTIME=${CC_RUNTIME} ./metrics/run_all_metrics.sh
tests/Makefile: ./ginkgo -v -focus "${FOCUS}" ./integration/docker/ -- -runtime=${CC_RUNTIME} -timeout ${TIMEOUT}
tests/.git/COMMIT_EDITMSG:Use CC_RUNTIME variable from Makefile instead of expect

```

chavafg:
@jcvenegas yes, that Makefile must be the only place where we use it.

jcvenegas:
@chavafg @grahamwhaley  ok, changed the name .

PR #638, opened 2017-10-24
integration: Use docker with user defined runtime

jcvenegas:
Use CC_RUNTIME variable from Makefile instead of expect
the cc-runtime defined as default runtime in docker.

The PR also  adds two ginkgo options.

- verbose:

Add `-v` option to ginkgo in order to know what commands
are being running on each tests and not only the lines of
ginkgo.

- focus:

Added -fucus option allow to only run some tests.

Example:
```
make integration FOCUS="docker kill"
```
will run only docker kill tests

- Changed `CC_RUNTIME` variable name to `RUNTIME`

Fixes: #637

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Path of Clear Containers Runtime
-CC_RUNTIME ?= cc-runtime
+RUNTIME ?= cc-runtime
 
 # The time limit in seconds for each test
 TIMEOUT ?= 15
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -31,13 +31,13 @@ ginkgo:
 	unlink vendor/src
 
 functional: ginkgo
-	./ginkgo functional/ -- -runtime ${CC_RUNTIME} -timeout ${TIMEOUT}
+	./ginkgo -v functional/ -- -runtime ${CC_RUNTIME} -timeout ${TIMEOUT}
```
grahamwhaley: nit - obviously you also snuck the `-v` in here as well - I presume that is verbose, and not version output ?

jcvenegas: hehe - I missed to split the commit, I partially  modify the makefile to see the commands that are being running ( `-v`) and found really useful to only tests I want `-focus`. I think are useful let me split the commit to have that merged in this PR.

jcvenegas: @grahamwhaley  now is explained why is added -v and -focus

chavafg:
can you fix typos in the commit messages: `fucus` and `raname`

PR #639, opened 2017-10-25
Azure: Fixes to docker and crio setup

mcastelino:
- Use daemon.json to setup the runtime instead of a systemd unit file.
  This allows the distro defaults to be preserved.

- Set the default graph driver to devicemapper

- Also ensure that each VM gets its own unique set of disks.
  This allows the script to be run multiple times as long as the
  VM name is unique

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

### DIFF
--- a/Azure/cloud-init.txt
+++ b/Azure/cloud-init.txt
@@ -73,17 +73,9 @@ apt-get install -y thin-provisioning-tools
 
 # Configure the clear containers runtime using devicemapper as
 # storage driver
-mkdir -p /etc/systemd/system/docker.service.d/
-cat <<EOF | tee /etc/systemd/system/docker.service.d/clr-containers.conf
-[Service]
-ExecStart=
-ExecStart=/usr/bin/dockerd -D --add-runtime cc-runtime=/usr/bin/cc-runtime --default-runtime=cc-runtime
-EOF
-
-# Add devicemapper configuration
 mkdir -p /etc/docker
-
 mkdir -p /etc/lvm/profile
+
 cat <<EOF | tee /etc/docker/daemon.json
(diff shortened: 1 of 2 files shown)


Azure/cloud-init.txt
```diff
@@ -73,17 +73,9 @@ apt-get install -y thin-provisioning-tools
...
-ExecStart=
-ExecStart=/usr/bin/dockerd -D --add-runtime cc-runtime=/usr/bin/cc-runtime --default-runtime=cc-runtime
-EOF
-
-# Add devicemapper configuration
 mkdir -p /etc/docker
-
 mkdir -p /etc/lvm/profile
+
 cat <<EOF | tee /etc/docker/daemon.json
```
jodh-intel: Not strictly part of this PR, but this should probably be `tee -a` to avoid stamping on whatever was there before.

sameo:
LGTM

grahamwhaley:
May be related to the WIP title - the CIs are all failing as there is no 'Fixes' in the commit, just to note.
Not that the CI will run this script anyhow....

Azure/cloud-init.txt
mcastelino: @jodh-intel I think we need to write a little utility to insert our runtime into the json file. This is also eliminate the one manual step we have today to install CC. We should build it into cc-runtime itself to help install itself.

GabyCT:
lgtm

Azure/cloud-init.txt
jodh-intel: #763 raised to discuss this idea.

Issue #630, opened 2017-10-20
measure cpu consumption when doing disk io tests


grahamwhaley:
We measure the CPU utilisation when doing the network tests (https://github.com/clearcontainers/tests/blob/master/metrics/network/network-metrics-cpu-consumption.sh#L87) - we should do similar for the disk io tests (https://github.com/clearcontainers/tests/tree/master/metrics/storage)

The results can be useful when comparing new optimisations, particularly when trying to gauge their effectiveness around scalability.

grahamwhaley:
And whilst at it - grab the PSS and RSS memory consumption values as well

PR #653, opened 2017-10-27
metrics: storage: measure cpu and memory as well as bandwidth

grahamwhaley:
Whilst running the storage (fio) tests, also measure the CPU
consuption and the memory footprint being used.
To do this we re-structure the test to set up the initial
test environment (get the container up and then init the
test file - which takes longer than the test itself), and then
we can know exactly when we are running the actual test.

Fixes: #630

### DIFF
--- a/metrics/storage/fio_job.sh
+++ b/metrics/storage/fio_job.sh
@@ -37,16 +37,21 @@ NUM_JOBS="1"
 OPERATION="randread"
 TEST_NAME="storage fio test"
 EXTRA_ARGS=""
+TEST_TIME="30"
 
 # This docker image includes FIO tool.
 FIO_IMAGE="clusterhq/fio-tool"
+
+# We need to know which QEMU to monitor
+QEMU_PATH=${QEMU_PATH:-$(get_qemu_path)}
 
 # The FIO tool included in clusterhq/fio-tool docker image
 # works in "/tmp" by default. "/tmp" is overridden by /home
 # mount point in order to avoid tmpfs and use the disk.
 FILE_NAME="/home/fio_test"
 
-declare FIO_JOB
+declare FIO_RUN_JOB
+declare FIO_PREP_JOB
 
 function help()
 {
@@ -72,10 +77,33 @@ EOF
 
 # Create the job configuration for FIO
 # this job will run inside of the container.
-function create_fio_job()
+# This job sets up the image - creates the test file
+function create_fio_prep_job()
 {
-	FIO_JOB="$(cat << EOF
-	#!/bin/bash
+	FIO_PREP_JOB="$(cat << EOF
+	# This fio job will run inside of
+	# the container.
+	# It only runs for 1s, its job is to create
+	# the test file
+
+	fio --ioengine="$ENGINE" \
+		--name="$FIO_TEST_NAME" \
+		--filename="$FILE_NAME" \
+		--bs="$BLOCK_SIZE" \
+		--size="$FILE_SIZE" \
+		--readwrite="$OPERATION" --max-jobs=$NUM_JOBS \
+		--runtime=1 \
+		${EXTRA_ARGS}
+EOF
+)"
+}
+
+# Create the job configuration for FIO
+# this job will run inside of the container.
+# This job runs the actual test
+function create_fio_run_job()
+{
+	FIO_RUN_JOB="$(cat << EOF
 	# This fio job will run inside of
 	# the container.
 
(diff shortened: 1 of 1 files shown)


metrics/storage/fio_job.sh
```diff
@@ -72,24 +77,49 @@ EOF
...
+# this job will run inside of the container.
+# This job runs the actual test
+function create_fio_run_job()
+{
+	FIO_JOB="$(cat << EOF
+	#!/bin/bash
+	# This fio job will run inside of
+	# the container.
+
+	fio --ioengine="$ENGINE" \
```
jodh-intel: We might not care (and I see there are already existing functions similar to this), but you could make that `exec fio` to avoid creating a 2nd process. In fact, why don't we just call `fio` directly? I've just noticed that `$FIO_JOB` itself is called via `bash -c` so we have 3 processes I think?

metrics/storage/fio_job.sh
```diff
@@ -72,24 +77,49 @@ EOF
 
 # Create the job configuration for FIO
 # this job will run inside of the container.
-function create_fio_job()
+# This job sets up the image - creates the test file
+function create_fio_prep_job()
 {
 	FIO_JOB="$(cat << EOF
```
grahamwhaley: Just found a couple of uses of `FIO_JOB`, which got deprecated when I split it into RUN and PREP phases ... think I got bit by a rebase.... new push coming.

metrics/storage/fio_job.sh
```diff
@@ -125,7 +155,7 @@ function main()
 {
 	cmds=("bc" "awk")
```
MarioCarrilloA: we can ask for `smem` tool

metrics/storage/fio_job.sh
```diff
@@ -161,14 +194,42 @@ function main()
...
+	docker exec -ti ${CONTAINER_ID} bash -c "$FIO_PREP_JOB"
+
+	# Kick off background tasks to measure the CPU, PSS and RSS
+	cpu_temp=$(mktemp fio_job_cpu.XXX)
+	pss_temp=$(mktemp fio_job_pss.XXX)
+	rss_temp=$(mktemp fio_job_rss.XXX)
+
+	# We know the test will run for 30s (unless somebody changed it...)
+	# So spread the measures out over the test run
+	(sleep 10; ps --no-headers -o %cpu -C $(basename ${QEMU_PATH}) > ${cpu_temp})&
```
MarioCarrilloA: I made a test and it shows and error using `runc`, it seems that the test is looking for a `qemu` instance, maybe we could check the runtime and in the case of `runc` we can measure the `fio` process running in host

GabyCT:
@grahamwhaley while running the test I am getting using CC

===== starting test [storage IO random write bs 16k] =====
command: docker: yes
command: bc: yes
command: awk: yes
Units: KB
389936.00 KB/s
ERROR: result argument not supplied

metrics/storage/fio_job.sh
grahamwhaley: Yeah. I knew this was a gap when I did the PR - I was going to ask @GabyCT what 'process' she measured when not running a VM runtime - and I guess the logical answer is the `fio` process - so, sure, let me make that modification...

grahamwhaley:
Hi @GabyCT Yeah, I found an issue where the write test seems to not honour the `--runtime=30` request if it, probably, manages to cover the whole file - so the test quits before the measurements take place, and then we end up with an empty result, which baulks :-(
I have made a modification to take the samples at 3/4/5s - my write test was completing in 13s for reference. What a pain ;-)
I have some other mods coming as well.
Oh, and another note - if we use the `-x magic args`, they include a `--ramp_time=60 --runtime=300` - which messes the test as well - in that case I set the sample sleep times to 65/66/67 to get past the ramp_time.

So, now I have give us the `-x` argument, that sort of can mess with this PRs intention - I think 'such is life', as we can never make the script smart enough to handle everything...

metrics/storage/fio_job.sh
mcastelino: @GabyCT can you confirm which process is used in the case of runc?

GabyCT:
@grahamwhaley and @mcastelino for runc it is the fio process

metrics/storage/fio_job.sh
grahamwhaley: Heh - we probably don't care much, but I took out the `#!/bin/bash`, and the world seems fine...
I've not injected the `exec`, as I've not seen that as a common use case on a docker command line.
Mostly I'd like to keep the script readable - I think that is how the original layout came about for those commands.

grahamwhaley:
Updated with feedback. Fixed up some niggles. Rebased and re-pushed.

GabyCT:
lgtm

GabyCT:
@grahamwhaley could you please check this?, I want to merge this but it has a conflict. Thanks

grahamwhaley:
rebased, resolved the merge issue with my other 'rm the fio containers' that already merged, and re-pushed...

PR #659, opened 2017-10-30
ci: integration: Fix docker kill tests

### LINKED ISSUE #658: ci: integration: Fix docker kill tests
The CI fails sporadically when running `docker kill` commands, because it does not get the expected exit code. The reason behind that is that we should not rely on a Sleep() to ensure that we can actually read the exit code.

sboeuf:
Recently, the CI has been failing pretty often when running docker
kill tests. This is due to the way the tests are written. Indeed,
because the exit code is retrieved after a time.Sleep(), it does
not ensure the container has properly exited.

This patch inspects State.Status from the container, that way it
makes sure to retrieve the container exit code only after the
container has exited. With this way to proceed, we will never
read an exit code if it is not ready to be read.

Fixes #658

### DIFF
--- a/docker.go
+++ b/docker.go
@@ -67,8 +67,64 @@ func StatusDockerContainer(name string) string {
 	return state[0]
 }
 
+// hasExitedDockerContainer checks if the container has exited.
+func hasExitedDockerContainer(name string) (bool, error) {
+	args := []string{"--format={{.State.Status}}", name}
+
+	stdout, _, exitCode := runDockerCommand("inspect", args...)
+
+	if exitCode != 0 || stdout == "" {
+		return false, fmt.Errorf("failed to run docker inspect command")
+	}
+
+	status := strings.TrimSpace(stdout)
+
+	if status == "exited" {
+		return true, nil
+	}
+
+	return false, nil
+}
+
 // ExitCodeDockerContainer returns the container exit code
-func ExitCodeDockerContainer(name string) (int, error) {
+func ExitCodeDockerContainer(name string, waitForExit bool) (int, error) {
+	// It makes no sense to try to retrieve the exit code of the container
+	// if it is still running. That's why this infinite loop takes care of
+	// waiting for the status to become "exited" before to ask for the exit
+	// code.
+	// However, we might want to bypass this check on purpose, that's why
+	// we check waitForExit boolean.
+	if waitForExit {
+		errCh := make(chan error)
+		exitCh := make(chan bool)
+
+		go func() {
+			for {
+				exited, err := hasExitedDockerContainer(name)
+				if err != nil {
+					errCh <- err
+				}
+
+				if exited {
+					break
+				}
+
+				time.Sleep(time.Second)
+			}
+
+			close(exitCh)
+		}()
+
+		select {
+		case <-exitCh:
+			break
+		case err := <-errCh:
+			return -1, err
+		case <-time.After(time.Duration(Timeout)*time.Second):
+			return -1, fmt.Errorf("Timeout reached after %ds", Timeout)
+		}
+	}
+
 	args := []string{"--format={{.State.ExitCode}}", name}
 
 	stdout, _, exitCode := runDockerCommand("inspect", args...)
--- a/docker.go
+++ b/docker.go
@@ -53,6 +53,19 @@ func runDockerCommand(command string, args ...string) (string, string, int) {
 	return runDockerCommandWithTimeout(time.Duration(Timeout), command, args...)
 }
 
+// LogsDockerContainer returns the container logs
+func LogsDockerContainer(name string) (string, error) {
+	args := []string{name}
+
+	stdout, _, exitCode := runDockerCommand("logs", args...)
+
+	if exitCode != 0 {
+		return "", fmt.Errorf("failed to run docker logs command")
+	}
+
+	return strings.TrimSpace(stdout), nil
+}
+
 // StatusDockerContainer returns the container status
 func StatusDockerContainer(name string) string {
 	args := []string{"-a", "-f", "name=" + name, "--format", "{{.Status}}"}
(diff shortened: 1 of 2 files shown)


sboeuf:
@chavafg @grahamwhaley please take a look and merge it, this PR is gonna stabilize our CI, at least for the failure we were seeing on docker kill tests.

docker.go
```diff
@@ -67,8 +80,48 @@ func StatusDockerContainer(name string) string {
...
-func ExitCodeDockerContainer(name string) (int, error) {
+func ExitCodeDockerContainer(name string, waitForExit bool) (int, error) {
+	// It makes no sense to try to retrieve the exit code of the container
+	// if it is still running. That's why this infinite loop takes care of
+	// waiting for the status to become "exited" before to ask for the exit
+	// code.
+	// However, we might want to bypass this check on purpose, that's why
+	// we check waitForExit boolean.
+	if waitForExit {
+		for {
```
jcvenegas: Here is need a timeout

integration/docker/kill_test.go
```diff
@@ -65,32 +66,37 @@ var _ = Describe("docker kill", func() {
...
 			}
 
 			DockerRun(args...)
 
-			// we have to wait for the container workload
-			// to process the trap.
-			time.Sleep(5 * time.Second)
-			
 			if signal > 0 {
+				for {
```
jcvenegas: here as well ideally the test will print the traptag but if better have a timeout.

docker.go
```diff
@@ -53,6 +53,19 @@ func runDockerCommand(command string, args ...string) (string, string, int) {
 	return runDockerCommandWithTimeout(time.Duration(Timeout), command, args...)
 }
 
+// LogsDockerContainer returns the container logs
+func LogsDockerContainer(name string) (string, error) {
+	args := []string{name}
+
+	stdout, _, exitCode := runDockerCommand("logs", args...)
+
+	if exitCode != 0 || stdout == "" {
```
jcvenegas: This function fails if not not have stdout, it is possible that check for a container logs should not have any log ? 
Also this function is only used once and the error is ignored.  can you better not fail on non stdout and properly handle the error where is called?

integration/docker/kill_test.go
```diff
@@ -65,32 +66,37 @@ var _ = Describe("docker kill", func() {
...
+					logs, _ := LogsDockerContainer(id)
```
jcvenegas: here is ignoring error, if the container not exist and always fails this will be an infinite loop.

docker.go
sboeuf: Yes you're right, I should not consider an empty stdout as an error.

integration/docker/kill_test.go
sboeuf: I have added a timeout handler.

integration/docker/kill_test.go
sboeuf: Ignoring this err is on purpose, I have added a comment to explain.

sboeuf:
@jcvenegas I think I have addressed all your comments, please take another look

docker.go
```diff
@@ -67,8 +67,64 @@ func StatusDockerContainer(name string) string {
...
+	}
+
+	return false, nil
+}
+
 // ExitCodeDockerContainer returns the container exit code
-func ExitCodeDockerContainer(name string) (int, error) {
+func ExitCodeDockerContainer(name string, waitForExit bool) (int, error) {
+	// It makes no sense to try to retrieve the exit code of the container
+	// if it is still running. That's why this infinite loop takes care of
```
grahamwhaley: nitpick: you could mention the timeout of the loop in the comment as well - as 'infinite loop' sounds dangerous otherwise :-)

grahamwhaley:
Some failures - unrelated I believe - in 16.04 and Fedora CIs.
Here is a snip from 16.04:
```
05:13:57 • Failure in Spec Setup (BeforeEach) [17.718 seconds]
05:13:57 restart
05:13:57 /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/restart_test.go:51
05:13:57   restart with docker [BeforeEach]
05:13:57   /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/restart_test.go:50
05:13:57     restart a container
05:13:57     /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/restart_test.go:49
05:13:57       should be running
05:13:57       /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/restart_test.go:48
05:13:57 
05:13:57       Expected
05:13:57           <int>: -1
05:13:57       to equal
05:13:57           <int>: 0
05:13:57 
05:13:57       /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/restart_test.go:32
05:13:57 ------------------------------
05:13:57 
05:13:57 
05:13:57 Summarizing 5 Failures:
05:13:57 
05:13:57 [Fail] inspect [BeforeEach] inspect with docker inspect with '{{json .Config}}' 
05:13:57 /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/inspect_test.go:39
05:13:57 
05:13:57 [Fail] users and groups running container [It] as ':postgres' user without additional groups 
05:13:57 /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/user_test.go:77
05:13:57 
05:13:57 [Fail] run container with docker [It] with 'false' as workload 
05:13:57 /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/run_test.go:54
05:13:57 
05:13:57 [Fail] terminal terminal with docker Check that pseudo tty is setup properly when allocating a tty [It] should display the pseudo tty's name 
05:13:57 /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/terminal_test.go:57
05:13:57 
05:13:57 [Fail] restart [BeforeEach] restart with docker restart a container should be running 
05:13:57 /home/jenkins/go/src/github.com/clearcontainers/tests/integration/docker/restart_test.go:32
05:13:57 
05:13:57 Ran 85 of 87 Specs in 628.606 seconds
05:13:57 FAIL! -- 80 Passed | 5 Failed | 0 Pending | 2 Skipped --- FAIL: TestIntegration (644.12s)
05:13:57 FAIL
```

PR #661, opened 2017-10-31
ci: Allow larger timeout for functional and integration tests

### LINKED ISSUE #660: ci: Increase functional/integration timeout to 60s
Because we are seeing some sporadic failures with a 15s timeout, let's increase it to make those tests pass. Tests can sometimes be slow because they run on Azure VM (meaning nested env).

sboeuf:
Our CI runs on Azure virtual machines and sometimes a simple docker
command can take some time because of the VM nesting. Let's increase
this timeout to avoid any false failure, which is going to make our
CI more stable.

Fixes #660

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@
 RUNTIME ?= cc-runtime
 
 # The time limit in seconds for each test
-TIMEOUT ?= 15
+TIMEOUT ?= 60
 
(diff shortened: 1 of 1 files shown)


sboeuf:
@grahamwhaley oh yeah you're actually right. Let's make it 60s since we want to see other things failing too. I mean that right now with 15s, this timeout might be hidding some internal timeout of 15s...

sameo:
LGTM

Issue #667, opened 2017-11-01
CI: Jenkins script does not use Jenkins ${WORKSPACE} for GOPATH


grahamwhaley:
Currently in our Jenkins setup script:
https://github.com/clearcontainers/tests/blob/master/.ci/jenkins_job_build.sh#L23
we set the GOPATH like:
```
mkdir -p ${HOME}/go

# Export all environment variables needed.
export GOROOT="/usr/local/go"
export GOPATH=${HOME}/go
```

which works fine in our single shot/build transient VMs - but, might not work so well for persistent Jenkins slaves or slaves that perform parallel builds.

Jenkins provides a per-job scratch area, and sets the `WORKSPACE` env var to point at it, as described in the pipeline docs that can be found on a local Jenkins instance:
localhost:8080/pipeline-syntax/
```
WORKSPACE
    The absolute path of the directory assigned to the build as a workspace.
```

I believe that `WORKSPACE` var will also be set for the `freestyle` projects that we currently use for Jenkins config, but we should check that... (let me confirm)...

The proposal is thus to move the GOPATH (and any other scratch items) to use the WORKSPACE area.

grahamwhaley:
confirmed - an `env` in the pre-build step of an empty `freestyle` project showed WORKSPACE is set...

```
Building in workspace /var/lib/jenkins/workspace/scratch
[scratch] $ /bin/sh -xe /tmp/jenkins1610530303306197489.sh
+ env
JENKINS_HOME=/var/lib/jenkins
MAIL=/var/mail/jenkins
LANGUAGE=en_GB:en
J2SDKDIR=/usr/lib/jvm/java-8-oracle
USER=jenkins
RUN_CHANGES_DISPLAY_URL=http://unconfigured-jenkins-location/job/scratch/1/display/redirect?page=changes
NODE_LABELS=master
J2REDIR=/usr/lib/jvm/java-8-oracle/jre
SHLVL=1
HOME=/var/lib/jenkins
HUDSON_COOKIE=xxxxxxxxxxxx
JENKINS_SERVER_COOKIE=xxxxxxxxxxx
WORKSPACE=/var/lib/jenkins/workspace/scratch
LOGNAME=jenkins
...
```

PR #668, opened 2017-11-01
ci: jenkins: Use Jenkins workspace for GOPATH

grahamwhaley:
We had our GOPATH being created/set to our home dir. That worked
fine for short lived Jenkins VMs currently utilised for the QA CI,
but might not be suitable for longer lived Jenkins slaves or ones
that perform parallel runs.
Jenkins provides a defined ${WORKSPACE} for each job - use that
location for GOPATH - which will also give us the benefit of
dropping and cleaning up properly if we ever use the 'remove
workspace after completion' feature of Jenkins.

Fixes: #667

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -20,11 +20,22 @@ set -e
 cc_repo=$1
 
 tests_repo="github.com/clearcontainers/tests"
-mkdir -p ${HOME}/go
+
+# This script is intended to execute under Jenkins
+# If we do not know where the Jenkins defined WORKSPACE area is
+# then quit
+if [ -z "${WORKSPACE}" ]
+then
+	echo "Jenkins WORKSPACE env var not set - exiting" >&2
+	exit 1
+fi
+
+# Put our go area into the Jenkins job WORKSPACE tree
+export GOPATH=${WORKSPACE}/go
+mkdir -p ${GOPATH}
 
 # Export all environment variables needed.
 export GOROOT="/usr/local/go"
-export GOPATH=${HOME}/go
 export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:${PATH}
 
 # Download and build goveralls binary in case we need to submit the code
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -26,7 +26,6 @@ mkdir -p ${HOME}/go
 export GOROOT="/usr/local/go"
 export GOPATH=${HOME}/go
 export PATH=${GOPATH}/bin:/usr/local/go/bin:/usr/sbin:${PATH}
-export CI=true
 
 # Download and build goveralls binary in case we need to submit the code
 # coverage.
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -45,23 +45,31 @@ then
 	go get github.com/mattn/goveralls
 fi
 
-# Get the repository and move HEAD to the appropriate commit.
+# Get the repository and move to the correct commit
 go get ${cc_repo} || true
 pushd ${GOPATH}/src/${cc_repo}
 if [ ${ghprbPullId} ] && [ ${ghprbTargetBranch} ]
 then
+	# For PRs we rebase the PR commits onto the defined target branch
 	git fetch origin pull/${ghprbPullId}/head && git checkout FETCH_HEAD && git rebase origin/${ghprbTargetBranch}
 else
+	# Othewise we test the master branch
 	git fetch origin && git checkout origin/master
 fi
 
-# Setup and run the tests.
+# All repos apart from the test repo run some setup/checks from their own
+# .ci/setup.sh before invoking the test repo .ci/setup.sh. Thus, the test
+# repo has its own pre-setup.sh script that needs running
 if [ "${cc_repo}" == "${tests_repo}" ]
 then
         .ci/setup_tests.sh
 fi
 
+# Set up the distro environment. Get, build and install all the latest
+# components
 .ci/setup.sh
+
+# Run the test suite
 .ci/run.sh
 
 # Publish the code coverage if needed.


grahamwhaley:
@sboeuf @chavafg this is a bit of an RFC right now - what do you think?
For my metrics CI I will have more permanent (not single shot VM) jenkins slaves, so feel this is a move in the right direction - I may (not quite decided yet) set 'remove workspace' on my Jenkins so I don't just gather up lots of go/src/repos for all my old PRs for instance.

grahamwhaley:
@chavafg , just out of interest, as I suspect this is unrelated to the PR - how is the kubernetes qa for stability - is the above 'failure' semi-expected at present for instance?

.ci/jenkins_job_build.sh
```diff
@@ -20,11 +20,13 @@ set -e
 cc_repo=$1
 
 tests_repo="github.com/clearcontainers/tests"
-mkdir -p ${HOME}/go
+
+# Use the Jenkins workspace for our work area
+export GOPATH=${WORKSPACE}/go
```
jodh-intel: Might be worth adding a check to ensure `WORKSPACE` is set? If not, either error, or fall back to the old directory with a message?

grahamwhaley: Indeed, would not be a bad thing - I did briefly ponder what would happen if it were not set :-)
If we accept `WORKSPACE` is the right thing to do - and this file is titled 'jenkins', then I'd probably make it abort if not set.

chavafg: I agree with @jodh-intel  that we should fall back to `${HOME}/go` if the `WORKSPACE` does not exists as I (and maybe others) usually run the CI scripts on other machines to setup my CC environment.

chavafg:
Hi @grahamwhaley 
k8s tests are more stable now, but I checked the logs of the k8s execution and found that this was an unusual problem with the network, so we don't need to be worried about this:
```
ERROR: Get https://api.github.com/repos/clearcontainers/tests/pulls/668: net/http: TLS handshake timeout
```

.ci/jenkins_job_build.sh
grahamwhaley: Sure, I'll add a check.
I'm not wholly sold on the fallback to `${HOME}`, mostly as the script is called `jenkins_job_build.sh` - thus inferring it is made for Jenkins. I know @sboeuf won't like renaming it and making it more 'jenkins neutral/unaware', as we'd have to modify all the jenkins setups to change the script name they are calling - but - something else that bugs me a little is that it is called `build`, yet it also blatantly does the run as well (I am forming a PR to stop it doing the run in the case of the metrics CI at least - but maybe we should lift the run out into another script, and wrap/rename this one...).

@sboeuf @chavafg thoughts on copy/splitting this script into two that do `build` and `run_qa`, and then making this a wrapper that invokes those two (or run_metrics for the metrics CI case).
I think that will then give us a clearer definition of what the scripts actually do and a touch more flexibility in how they are called either from different CI systems or by hand etc.

grahamwhaley: and - I'm happy to put that PR together, of course

sboeuf: @grahamwhaley well I gave this name to the script because it is executed in the `Build` section of the Jenkins job.

sboeuf: BTW, we can do the fallback to `$HOME` but `$WORKSPACE` must be set in case of Jenkins. So I think you might consider to return an error if `$WORKSPACE` is not set.

sboeuf: @grahamwhaley of course you can split the script, I don't mind (as long as we keep the name of this one as a wrapper as you mentioned). If you think splitting will make this clearer and more reusable, that's fine ;)

jodh-intel: How about erroring if `CI` is set and fallback if not?

jodh-intel: If the Jenkins script is going to change, could we maybe add a comment explaining why it has its name, and maybe throw in a reference to https://github.com/clearcontainers/jenkins.

grahamwhaley: @jodh-intel Looks like this script sets `CI` :-) - and I don't see `CI` set in my Jenkins env. But, sure, we'll figure something out.

@chavafg just for clarity, to help me try to restructure and break up the scripts maybe - you run the scripts locally just to set up your CC env - you don't use then to actually run QA tests presumably?
That is sort of implying to me that we need some setup scripts for hand-running - and then I wonder how they are different from the distro install scripts we already have... your thoughts welcome here...

@mcastelino you were re-creating and running the CI locally recently I think - were you using Jenkins or running by hand (I suspect by hand) - and if so, did you use this script, and thus did you have to hack it or set up a bunch of env settings to make it work?

My plan is to see if we can break this apart to make it work for everybody (QA CI Jenkins, metrics CI Jenkins and hand-run cases) - and document that a bit.

Oh, and @jodh-intel yes, I intend to add more comments to the scripts to note why we have to do all the little bits in them we do (like why the test repo has its own little bit of magic etc.) - saves me having to work it all out again in another 6 months ;-)

chavafg: @grahamwhaley yes, it is to setup the CC environment only. Not sure what you mean with the distro install scripts? I use this script because it install latest versions of each component.

mcastelino: @grahamwhaley I was not running the full jenkins setup. Just running what jenkins would run which was 
- Create VM
- Install Go
```
cd $GOHOME/src/github.com/clearcontainers/runtime/
sudo -E .ci/setup.sh
sudo -E .ci/run.sh
```
- comment out qemu-lite installation
- Switch branches
- Revendor virtcontainers into runtime (if virtcontainers was modified)

grahamwhaley:
Hi @sboeuf @chavafg @jodh-intel . I've updated the PR after staring some more an taking on the feedback:
1) Added the check for WORKSPACE being set
1) Annotated the script
1) Dropped the CI var as I don't think we actually use it

@chavafg - afaict, to do the 'install all latest components', all you actually have to do is run `.ci/setup.sh` is it not? It looks like that is maybe the only bit of this script you are really utilising when running it by hand?, so I've taken the presumption that I can leave the `WORKSPACE` code and check in here, and you can invoke `.ci/setup.sh` to get your local repos and install up to date.

... now I see if the CI passes with the changes....

grahamwhaley:
Oh, and after some more contemplation I think I won't need to split the script - post this merging I will add a check around `.ci/run.sh` to check for `METRICS_CI`, and do an if/else to pick `run_metrics_ci.sh` for the METRICS Jenkins...

jodh-intel:
lgtm

grahamwhaley:
oops, forgot the Fixes....

PR #689, opened 2017-11-08
tests: Add MTU test using swarm

### LINKED ISSUE #688: add mtu test using swarm
This test will check the mtu interfaces as well as their values

GabyCT:
This will check MTU interfaces as well as their values.

Fixes #688

### DIFF
--- a/integration/swarm/swarm.bats
+++ b/integration/swarm/swarm.bats
@@ -25,7 +25,7 @@ nginx_image="gabyct/nginx"
 # Name of service to test swarm
 SERVICE_NAME="testswarm"
 # Maximum number of replicas that will be launch (just a quick test)
-number_of_replicas=4
+number_of_replicas=1
 # Saves the name of the replicas
 declare -a REPLICAS_UP
 # Saves the hostname of the replicas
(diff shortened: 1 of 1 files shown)


grahamwhaley:
To my eyes the test looked sane, but then it fails on the CI:
```
23:13:57 not ok 4 check mtu values in different interfaces
23:13:57 # (in test file integration/swarm/swarm.bats, line 100)
23:13:57 #   `[ "$mtu_value_eth0" = "$mtu_value_tap0" ]' failed
```
Is that expected/intentional?
Also makes me think maybe we can add a bit of logging to the failure cases so we can see the values of the `x != y` check etc. ?

jodh-intel:
Agreed. The simplest way that I've found to get debug out of bats when a test fails is this rather cute hack:

```
teardown() {
    [ "$BATS_ERROR_STATUS" -eq 0 ] || set
}
```
Or you could get elaborate and do what we did with some of the cor tests:

- https://github.com/01org/cc-oci-runtime/blob/master/tests/functional/state.bats#L21..L32
- https://github.com/01org/cc-oci-runtime/blob/master/tests/functional/common.bash.in#L387..L422

GabyCT:
@grahamwhaley and @jodh-intel thanks for the feedback, yes I will add a debug information. I see that swarm is failing mainly because swarm replicas are not running , they remained in created status

PR #704, opened 2017-11-13
tests: Add MTU test for swarm

### LINKED ISSUE #688: add mtu test using swarm

GabyCT:
Check MTU interfaces and values using swarm.

Fixes #688

### DIFF
--- a/integration/swarm/swarm.bats
+++ b/integration/swarm/swarm.bats
@@ -90,6 +90,21 @@ setup() {
 	export http_proxy="$proxy"
 }
 
+@test "check mtu values in different interfaces" {
+	REPLICAS_UP=$($DOCKER_EXE ps -q --filter name="${SERVICE_NAME}")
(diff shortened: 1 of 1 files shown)


integration/swarm/swarm.bats
```diff
@@ -90,6 +90,21 @@ setup() {
 	export http_proxy="$proxy"
 }
 
+@test "check mtu values in different interfaces" {
+	REPLICAS_UP=$($DOCKER_EXE ps -q --filter name="${SERVICE_NAME}")
```
grahamwhaley: Just a couple of thoughts:
- we repeat the `REPLICAS_UP` idiom through this file - i wonder if that can be lifted out/re-factored?
- we could check that REPLICAS_UP is not empty here (and maybe other places) as a sanity check.

Issue #710, opened 2017-11-14
CI: Jenkins: Do not run the QA checks if we are running under the metrics CI


grahamwhaley:
If we are running under the metrics CI, then we want to use the jenkins script to get the system set up, but we don't actually want to run the QA checks - as we will be running the metrics checks after this script (and we only run the metrics CI once we know the QA CI has passed - no point in trying to measure something you've already found out to be broken...).

Basically, we want to caveat this line on !metrics:
https://github.com/clearcontainers/tests/blob/master/.ci/jenkins_job_build.sh#L76

PR #711, opened 2017-11-14
CI: metrics: Jenkins: Do not run QA checks for metrics CI

grahamwhaley:
When we are running the metrics CI we presume (or have checked)
that the QA tests already work. We do not need to run them again
before we run the metrics CI.
Thus, only run the QA checks when we are not under the METRICS_CI.

Fixes: #710

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -72,8 +72,15 @@ fi
 # components
 .ci/setup.sh
 
-# Run the test suite
-.ci/run.sh
+# The metrics CI does not need to do the QA checks - it only runs once
+# it knows the QA CI has passed already.
+if [ -z "${METRICS_CI}" ]
+then
+	# Run the test suite
+	.ci/run.sh
+else
+	echo "Under METRICS_CI - skipping test run"
+fi
 
 # Publish the code coverage if needed.
 if [ ${COVERALLS_REPO_TOKEN} ]


.ci/jenkins_job_build.sh
```diff
@@ -72,8 +72,15 @@ fi
...
-# Run the test suite
-.ci/run.sh
+# The metrics CI does not need to do the QA checks - it only runs once
+# it knows the QA CI has passed already.
+if [ -z "${METRICS_CI}" ]
+then
+	# Run the test suite
+	.ci/run.sh
+else
+	echo "Under METRICS_CI - skipping test run"
```
chavafg: generally lgtm, 
just a question, shouldn't here be a call to `run_metrics_ci.sh`, or you call it directly in your jenkins setup?

grahamwhaley: Fine question.
I intend to run it from my Jenkins script.
Ideally I would have either:
- lifted `.ci/run.sh` out of this script all togehter
- inverted the test to only run `.ci/run.sh` if we know we are in the QA CI

but.... to do either of those would involve editing the many many Jenkins configs we have - so the easier route was to just special case the metrics.
If later we find we are having to do this again for some reason then we should consider 'biting the bullet', and updating all CIs at the same time to make this script more generically useful etc.

chavafg:
Seems like a network issue too on Fedora...

```
14:30:40 Cloning into '/home/jenkins/jenkins_slave/workspace/clear-containers-tests-fedora-26-PR/go/src/github.com/clearcontainers/runtime'...
14:30:40 fatal: unable to access 'https://github.com/clearcontainers/runtime/': Could not resolve host: github.com
```

rebuilding

Issue #724, opened 2017-11-15
ci: metrics: different CIs need different 'check' methods


grahamwhaley:
Our different CIs use different methods to check and report if the metrics run (https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh#L105) has passed or not.
Currently the script is hard wired for `localCI` - modify it to check which CI it is running under and do the appropriate action.

PR #725, opened 2017-11-15
ci: metrics: Choose check method according to CI

grahamwhaley:
Depending on which CI system we are running (such as localCI
or Jenkins etc.), we want to use different 'tail end' checks.
Modify the script to check the CI env variables, and do the
appropriate thing.
Note, the 'generic' thing is to use the checkmetrics tool, and
have the config file determined by the name of the host, thus
easily allowing different config files per host (that are not
all called the same thing).

Fixes: #724

### DIFF
--- a/.ci/run_metrics_ci.sh
+++ b/.ci/run_metrics_ci.sh
@@ -14,6 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


grahamwhaley:
ping @chavafg or @devimc - pls review - blocked on 'approver' review.

PR #736, opened 2017-11-17
integration/docker: set and check environment variables in the container

### LINKED ISSUE #735: integration/docker: check if environment variables are set correctly
using ```docker run -e``` check if environment variables are set correctly in the container

devimc:
check environment variables are set correctly in the container

fixes #735


GabyCT:
lgtm

grahamwhaley:
lgtm.
I guess we could have also checked the inverse (that an env is not set if we did not set it), but that sort of feels somewhat artificial and tbh I can't really think up a good way to make that sane :-)

Issue #741, opened 2017-11-21
jenkins: PATH: /sbin required for modprobe for cc-check


grahamwhaley:
Running Jenkins on ubuntu 16.04.3 LTS, the `cc-runtime cc-check` step began failing (I'm sure it worked before - maybe something changed with a recent Ubuntu update...?).
Anyhow, I tracked this down to the `jenkins` user not being a priv'd user, and thus not having `/sbin` in their path, which `cc-runtime cc-check` requires when it invokes `modprobe` to check we have all the relevant kernel bits in place.
Here is the failure output:

```
00:00:52.859 time="2017-11-21T15:50:41Z" level=info msg="CPU property found" description="Intel Architecture CPU" name=GenuineIntel source=runtime type=attribute
00:00:52.859 time="2017-11-21T15:50:41Z" level=info msg="CPU property found" description="Virtualization support" name=vmx source=runtime type=flag
00:00:52.859 time="2017-11-21T15:50:41Z" level=info msg="CPU property found" description="64Bit CPU" name=lm source=runtime type=flag
00:00:52.859 time="2017-11-21T15:50:41Z" level=info msg="CPU property found" description=SSE4.1 name=sse4_1 source=runtime type=flag
00:00:52.859 time="2017-11-21T15:50:41Z" level=info msg="kernel property found" description="Kernel-based Virtual Machine" name=kvm source=runtime type=module
00:00:52.859 time="2017-11-21T15:50:41Z" level=info msg="kernel property found" description="Intel KVM" name=kvm_intel source=runtime type=module
00:00:52.859 time="2017-11-21T15:50:41Z" level=info msg="Kernel property value correct" description="Intel KVM" name=kvm_intel source=runtime type=module
00:00:52.859 time="2017-11-21T15:50:41Z" level=info msg="Kernel property value correct" description="Intel KVM" name=kvm_intel source=runtime type=module
00:00:52.859 time="2017-11-21T15:50:41Z" level=error msg="kernel property not found" description="Host kernel accelerator for virtio" name=vhost source=runtime type=module
00:00:52.859 time="2017-11-21T15:50:41Z" level=error msg="kernel property not found" description="Host kernel accelerator for virtio network" name=vhost_net source=runtime type=module
00:00:52.859 time="2017-11-21T15:50:41Z" level=error msg="ERROR: System is not capable of running Intel® Clear Containers" source=runtime
00:00:52.859 ERROR: System is not capable of running Intel® Clear Containers
```

[...]

PR #742, opened 2017-11-21
ci: jenkins: add /sbin to the PATH

grahamwhaley:
cc-runtime cc-check started failing for me - it looks like
if we are not a priv'd user then we need /sbin in our PATH
to find modprobe, or we fail the checks (for vhost enabled
for instance).
Add /sbin to our existing PATH manipulation.

Fixes: #741


jodh-intel:
lgtm

Issue #745, opened 2017-11-22
remove sudo to build crio 


chavafg:
We shouldn't build `cri-o` using sudo as there are issues when trying to cleanup the environment because some directories/files get owned by root instead of the current user.

PR #746, opened 2017-11-22
ci: Do not use sudo when building cri-o

chavafg:
We don't need sudo to build cri-o and this leads to having files
owned by root, which is an issue when trying to clean up the
environment.

Fixes: #745.


chavafg:
@grahamwhaley I tested locally and seems that we do not need use sudo for building cri-o. 
This hopefully resolves the issue you have when cleaning your environment.

devimc:
lgtm

GabyCT:
lgtm

PR #749, opened 2017-11-23
metrics: Improve warn and killing of unexpected processes

grahamwhaley:
We were silently killing stray hypervisors and shim processes, but we were not doing that with `sudo`, so sometimes we would get stray `process not found`errors whilst running the metrics.
Make it now a more blatant warning with some process details, and do the killall with sudo.

It is debatable if we should warn or die here - having stray processes hanging around might be a strong indicator of a bug...


GabyCT:
lgtm

PR #751, opened 2017-11-24
conformance/posixfs: Add posix filesystem conformance test

### LINKED ISSUE #750: Add posix filesystem conformance test
Add posix filesystem conformance test[1]

[1] https://sourceforge.net/p/ntfs-3g/pjd-fstest/ci/master/tree/

devimc:
This new test checks if the container's filesystem supports
most common filesystem operations

fixes #750

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,10 @@ openshift:
 swarm:
 	bats integration/swarm/swarm.bats
 
-check: functional crio integration swarm
+conformance:
+	bats conformance/posixfs/fstest.bats
+
+check: functional crio integration swarm conformance
 
 all: functional checkcommits integration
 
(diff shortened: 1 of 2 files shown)


grahamwhaley:
I like it - very nice :-)
I've noted on the Issue over in the runtime that we might also consider if we can use this as well:
http://posixtest.sourceforge.net/
which is available as a package under Ubuntu at least.

Can you check the Makefile is doing the right thing for me? I can't see any mention of conformance or fstests in the CI logs, and when I run it by hand: `make conformance`, it says `nothing to do`. I slapped it in the .PHONY, and then it did something, but I note swarm is not in the phony but swarm builds... most odd.

Also, when it did run the `bats` failed in your proxy check code - probably as I have no proxy set here - can you check that as well pls?

So, looking good, but can we check it does run as we expect under the CIs?

(I nudged the F26 CI for a rebuild btw, as it looked like a crio or swarm failure - but I see it has failed again. Worth a check if it is related to this PR, but I suspect it is not).

GabyCT:
lgtm

PR #755, opened 2017-11-27
metrics: Add save_results and baremetal support to remote tests

### LINKED ISSUE #754: Add save_results and baremetal support for remote networking tests
Add the functionality of save_results as well as support runc in the remote networking tests.

GabyCT:
Add the functionality of save_results as well as support runc
in the remote networking tests.

Fixes #754

### DIFF
--- a/metrics/network/remote_network/remote-networking-iperf3.sh
+++ b/metrics/network/remote_network/remote-networking-iperf3.sh
@@ -26,6 +26,8 @@ source "${SCRIPT_PATH}/remote-networking-test-common.sh"
 
 # Time to run the server using iperf3
 server_time=30
+# Arguments to run the client
+extra_args="-ti"
 
 # This function describes how to use this script
 function help {
@@ -39,23 +41,26 @@ Usage: $0 "[options]"
 		- Interface name where swarm will run.
 		- User of the host B.
 		- IP address of the host B
+		This is an example of how to run this script:
+		./remote-networking-iperf3.sh [options] -i <interface_name> -u <user> -a <ip_address>
 	Options:
-		-h	Shows help
-		-b	Run remote bandwidth
-		-j	Run remote jitter
-		-p	Run remote parallel bandwidth (-P4)
-		-l	Run remote latency
-		-t	Run all remote tests
-		-i	Interface name to run Swarm (mandatory)
-		-u	User of host B (mandatory)
 		-a	IP address of host B (mandatory)
-
+		-b	Run remote bandwidth
+		-h	Shows help
+		-i	Interface name to run Swarm (mandatory)
+		-j	Run remote jitter
+		-l	Run remote latency
+		-p	Run remote parallel bandwidth (-P4)
+		-t	Run all remote tests
+		-u	User of host B (mandatory)
 EOF
 )"
 }
 
 # This function will measure the bandwidth using iperf3
 function remote_network_bandwidth_iperf3 {
+	test_name="Remote network bandwidth"
+	get_runtime
 	setup_swarm
 	client_replica_status
 	server_replica_status
@@ -64,16 +69,21 @@ function remote_network_bandwidth_iperf3 {
 	start_server
 
 	client_id=$($DOCKER_EXE ps -q)
-	client_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -c $server_ip_address -t $server_time"
-	result=$(start_client "$client_id" "$client_command")
-	total_bandwidth=$(echo "$result" | tail -n 3 | head -1 | awk '{print $(NF-2), $(NF-1)}')
-	echo "Network bandwidth is : $total_bandwidth"
+	command="iperf3 -c $server_ip_address -t $server_time"
+	check_iperf3_client_command "$command"
+	result=$(start_client "$extra_args" "$client_id" "$client_command")
+	total_bandwidth=$(echo "$result" | tail -n 3 | head -1 | awk '{print $(NF-2)}')
+	units=$(echo "$result" | tail -n 3 | head -1 | awk '{print $(NF-1)}')
+	echo "Network bandwidth is : $total_bandwidth $units"
 
+	save_results "$test_name" "Remote Bandwidth" "$total_bandwidth" "$units"
 	clean_environment
 }
 
 # This function will measure the jitter using iperf3
 function remote_network_jitter_iperf3 {
+	test_name="Remote network jitter"
+	get_runtime
 	setup_swarm
 	client_replica_status
 	server_replica_status
@@ -82,16 +92,21 @@ function remote_network_jitter_iperf3 {
 	start_server
 
 	client_id=$($DOCKER_EXE ps -q)
-	client_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -c $server_ip_address -u -t $server_time"
-	result=$(start_client "$client_id" "$client_command")
-	total_jitter=$(echo "$result" | tail -n 4 | head -1 | awk '{print $(NF-3), $(NF-2)}')
-	echo "Network jitter is : $total_jitter"
+	command="iperf3 -c $server_ip_address -u -t $server_time"
+	check_iperf3_client_command "$command"
+	result=$(start_client "$extra_args" "$client_id" "$client_command")
+	total_jitter=$(echo "$result" | tail -n 4 | head -1 | awk '{print $(NF-4)}')
+	units=$(echo "$result" | tail -n 4 | head -1 | awk '{print $(NF-3)}')
+	echo "Network jitter is : $total_jitter $units"
 
+	save_results "$test_name" "Remote jitter" "$total_jitter" "$units"
 	clean_environment
 }
 
 # This function will measure parallel bandwidth using iperf3
 function remote_network_parallel_iperf3 {
+	test_name="Remote network parallel bandwidth"
+	get_runtime
 	setup_swarm
 	client_replica_status
 	server_replica_status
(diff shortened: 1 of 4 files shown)


metrics/network/remote_network/remote-networking-test-common.sh
```diff
@@ -44,6 +44,28 @@ second_service="testswarm2"
 port_first_service="8080:80"
 # Port where the second service will run
 port_second_service="8081:80"
+# Mount command for iperf3 (cc-runtime)
+mount="mount -t ramfs -o size=20M ramfs /tmp"
+
+# This function will retrieve the runtime from
+# host A and host B
+function get_runtime {
```
grahamwhaley: I was going to suggest that we don't need this and we use the 'common' set $RUNTIME, but I see this is somewhat specialised as we check both the local and remote machines - so, np :-)

metrics/network/remote_network/remote-networking-iperf3.sh
```diff
@@ -64,16 +68,26 @@ function remote_network_bandwidth_iperf3 {
 	start_server
 
 	client_id=$($DOCKER_EXE ps -q)
-	client_command="mount -t ramfs -o size=20M ramfs /tmp && iperf3 -c $server_ip_address -t $server_time"
-	result=$(start_client "$client_id" "$client_command")
-	total_bandwidth=$(echo "$result" | tail -n 3 | head -1 | awk '{print $(NF-2), $(NF-1)}')
-	echo "Network bandwidth is : $total_bandwidth"
-
+	if [ "$runtime_client" == "runc" ]; then
```
grahamwhaley: there is a lot of repetition here, both for the runtime check code and the iperf invocation code - it would be nice if we could factor those out into one or two more generic functions - but, if we can't then fine.

metrics/network/remote_network/remote-networking-test-common.sh
```diff
@@ -59,11 +81,11 @@ function setup_swarm {
 	$DOCKER_EXE service create \
 		--name "$first_service" --replicas "$number_of_replicas" \
 		--constraint 'node.labels.machine == machine1' \
-		--publish "$port_first_service" "$network_image" sh
+		--publish "$port_first_service" "$network_image" sh -c "sleep 60"
```
grahamwhaley: if these are meant to effectively 'run forever', and we kill them off/shut them down by hand later, maybe it is better to use something like `fail -f /dev/null` as the payload, rather than the `sleep 60` - having sleeps in the tests always makes us a little nervous for future-proofness.

GabyCT:
@grahamwhaley changes applied

grahamwhaley:
Hi @GabyCT . I tried to run up the test of remote iperf3 bandwidth, but had some difficultly with the command line setup. i think there are some things to fix with the arg getopt parsing etc.
- I think there is a `shift` in the getopt look that is messing up the parsing - it failed to get my -u and -a bits correct for instance: https://github.com/GabyCT/tests/blob/0dd601f0507dd19eb9ff35c9cfb9e954f63609c4/metrics/network/remote_network/remote-networking-iperf3.sh#L187
- can you sort the optargs string and the switch table into alphabetical - it makes it a little easier to debug. I know currently they are grouped more semantically, but I had trouble checking all was in order as they are not presently ordered: https://github.com/GabyCT/tests/blob/0dd601f0507dd19eb9ff35c9cfb9e954f63609c4/metrics/network/remote_network/remote-networking-iperf3.sh#L152
- I think you have your optarg ':' string with the ':'s in the wrong places - they should follow the opt that has args - so, 't' needs to lose its ':', and 'a' needs to gain one. That stopped me running the test ;-)
- I'm not sure the default arg settings in the common file are correct? For instance, I did not set an interface name on the command line for the iperf3 bandwidth test, but it did not complain as it picked up the (incorrect) default from the common file afaict: https://github.com/GabyCT/tests/blob/0dd601f0507dd19eb9ff35c9cfb9e954f63609c4/metrics/network/remote_network/remote-networking-test-common.sh#L28
- there seems to be a hard dependancy that the test type args (such as '-b') must follow the config args (such as '-a') - but that is not enforced in the script. I had my '-b' as the first arg on the command line, and strange things happen..... Maybe we can have a 'sanity_check_args()' function that does those -z checks and call that before each test invocation, such as:
```
  b)
    sanity_check_args
    remote_network_bandwidth_iperf3
```

hopefully that will make mis-typed runs easier to catch. We should probably make it clear in the in-script help that some arguments must come before others (I don't think optargs can do that for us, so we will just have to code it up I suspect).

GabyCT:
@grahamwhaley could you please tell me how do you put your command line?

GabyCT:
@grahamwhaley also which docker version are you using?

GabyCT:
@grahamwhaley you need to run it like this
bash remote-networking-iperf3.sh -b -i interface_name -u user -a ip_address

grahamwhaley:
Hi. I ran it like this:
`bash network/remote_network/remote-networking-iperf3.sh -b -u gwhaley -a 192.168.0.xxx`
yes, I didn't put the interface argument there - and that is probably why I saw the issues I had to debug.
I'm running docker 17.09.0-ce with overlay2 on Ubuntu server 16.04 - pretty recent CC components. Ah, I see in the docs it only works with Docker 1.12.1 - but, I don't think that effects the problems I reported - did not get that far yet ;-)

Sure, I added the `-i en01` to my command, and it got a lot further, and then probably hit what are the docker incompatibility issues.

I think there are bugs to fix (the `:` in the getopts looks pretty wrong to me as does the `shift` ?).

I also think if we have a hard requirement for mandatory arguments to be in a mandatory order then we either have to:
 - document that very clearly - in the docs and the script help output
 - and error out clearly in all other cases (which I don't think we do right now)
or
 - remove that strict ordering requirement, which looks entirely feasible, and that is what I expected the script to have.

GabyCT:
@grahamwhaley changes applied

grahamwhaley:
Hi @GabyCT ,
 I think it's still not working correctly for me. I ran it as:
`bash -x ./network/remote_network/remote-networking-iperf3.sh -b -a 192.168.0.111 -u gwhaley`
(yes, deliberately still with no `-i` argument)
and
`bash -$- ./network/remote_network/remote-networking-iperf3.sh -t -a 192.168.0.111 -u gwhaley -i eno1`
(with `-i`, but using `-t` instead of `-b`)

Neither of those worked. The first as I think there is an issue still with the global/common file picking up bad default arguments. The second as I think the `t` is still wrong in the optargs.

 I think what we need to do (pls check though) is:
- Remove the `:` after the `t` in the optargs? (as the `t` flag does not take an arg?)
- Remove the global arg parse/grab code, as we parse all of those in the getopts I think, and afaict they are wrong as they have fixed positions encoded which I think is wrong (and currently if you miss a commandline arg these falsely then pick up an incorrect setting - which is what I see for `-i` being missing): https://github.com/GabyCT/tests/blob/e26fd0ad838b742fde5055318a50782eb3400e53/metrics/network/remote_network/remote-networking-test-common.sh#L25-L32
- Add back in the `-z` checks that we have all the args set that we expect before we start invoking the actual tests (thanks for moving the test invocation later - but I think we still need to check the params are set).

GabyCT:
@grahamwhaley as the README and the help says could you please put the arguments in order and all of them thanks

grahamwhaley:
I could, of course, and I'm sure that would work until I hit my incorrect version of docker - but, if the arguments are not all present or in the correct order then the script should abort cleanly, not go wrong. That's all I'm really asking for...

GabyCT:
@grahamwhaley if you mess the order the script will not work that is the way is designed

GabyCT:
@grahamwhaley let me do the changes

MarioCarrilloA:
Hi @GabyCT I am a little confused, the arguments for the script should be put in certain order?, if yes, that is no good, `getopt` should solve that.

GabyCT:
@grahamwhaley changes applied

metrics/network/remote_network/remote-networking-iperf3.sh
```diff
@@ -100,81 +115,103 @@ function remote_network_parallel_iperf3 {
...
 	total_latency=$(echo $result | grep avg | awk '{print $(NF-1)}' | cut -d '/' -f 2)
-	echo "Network latency is : $total_latency ms"
+	echo "Network latency is : $total_latency $units"
 
+	save_results "$test_name" "Remote latency" "$total_latency" "$units"
 	clean_environment
 }
 
 function main {
+	[[ $# -ne 7 ]]&& help && die "Illegal number of parameters."
```
grahamwhaley: I still think it would be better to check the individual items have been set after the getopts has finished. With this arg counting method you could still pass if you passed the same arg twice for instance etc.

GabyCT:
@grahamwhaley I am adding an extra condition that will check that all arguments are there so if we repeat -a for example and we just set -a and -u, it will tell you that -i is missing

metrics/network/remote_network/remote-networking-iperf3.sh
MarioCarrilloA: I agree with this point, it is better to validate individually arguments e.g. if it is mandatory to have an IP address as an argument you could do something like this:
```
if [ -z "IP_ADDRESS" ]; then
         die "no input ip addres: mandatory argument"
fi
```

GabyCT: @MarioCarrilloA it has that already

metrics/network/remote_network/remote-networking-iperf3.sh
```diff
@@ -100,81 +115,107 @@ function remote_network_parallel_iperf3 {
...
 		esac
-		shift
 	done
 	shift $((OPTIND-1))
 
-	[ -z "$ssh_address" ] && help && die "Mandatory IP address of host B not supplied"
-	[ -z "$ssh_user" ] && help && die "Mandatory user of host B not supplied"
-	[ -z "$interface_name" ] && help && die "Mandatory interface name to run Swarm not supplied"
-
+	[[ -z "$interface_name" ]] && help && die "Missing IP Address."
```
GabyCT: @grahamwhaley and @MarioCarrilloA here it is that validation

MarioCarrilloA: great, then maybe is unnecessary this `[[ $# -ne 7 ]]&& help && die "Illegal number of parameters."`, and the order of the arguments should be irrelevant for this test execution.

GabyCT: @MarioCarrilloA yes the order of the arguments is not relevant

Issue #759, opened 2017-11-28
Remove test-versions.txt and rely on the clearcontainers/runtime versions.txt


chavafg:
Today we have 2 files on different repos which contain information of the dependencies versions we support, but this duplicates information and it is fragile since we can end up updating one, but not the other.

My proposal is to remove `test-versions.txt` from this repository and move all missing info to the `versions.txt` on the runtime repository: https://github.com/clearcontainers/runtime/blob/master/versions.txt 

So instead of sourcing `test-version.txt` in the CI scripts we will need to... 
```
go get github.com/clearcontainers/runtime
source $GOPATH/src/github.com/clearcontainers/runtime/version.txt
```

What do you think ? 
/cc @sboeuf @jodh-intel @grahamwhaley @jcvenegas

jodh-intel:
I agree that we should define all these versions in a single location to avoid the update/sync issue. On  https://github.com/clearcontainers/runtime/issues/608, I mentioned a possible circular dependency issue, but I don't think there really is one given that:

- all repos use `tests`.
- `tests` will always clone the runtime.

Therefore, if, for example, a proxy PR was going to be tested by the CI, the following would happen:

- proxy's CI scripts clone `tests` repo and run setup.
- `tests` setup scripts clone the `runtime` repo.
- `tests` can now access `versions.txt`.

The only problem would be if the tests needed to query `versions.txt` *before* calling `install_runtime.sh` I think. If we ever think that might happen, we should put `versions.txt` somewhere else: a repo that isn't a main component of the system and thus required by the CI. That could either be a new repo entirely or maybe the `packaging` repo?

Related:

- https://github.com/clearcontainers/runtime/issues/608
- https://github.com/clearcontainers/tests/issues/666

grahamwhaley:
Completely agree we should de-duplicate this.
I also agree that if the proposal to place it solely in the runtime repo does not work out, then maybe the packaging repo makes a lot of sense.
+1 that we try the runtime and see if that flies.

jodh-intel:
:airplane: or :boom:

chavafg:
I didn't notice that there was already an issue opened to address this, sorry for opening this one.
@jodh-intel I think we use`versions.txt` before running `install_runtime.sh` and I think we can fix it by running a `go get` the first time we use versions.txt, does this make sense?

[...]

Issue #666, opened 2017-11-01
install repo-supported version of docker


jodh-intel:
The `.ci/setup_env_ubuntu.sh` and `.ci/setup_env_fedora.sh` scripts  install the `docker-ce` package if docker is not already installed. However, they should be installing the newest version of docker that Clear Containers supports for the repo they are testing.

For example, for the runtime this should be `v17.09-ce`. See `docker_version=` and `docker_swarm_version=` in:

- https://github.com/clearcontainers/runtime/blob/master/versions.txt

This would translate into something like the following (ignore actual version numbers below - it's just showing the package-manager command to install a named version):

# Ubuntu

```
$ sudo apt-get -y install docker-ce=17.05.0~ce-0
```

# Fedora

```
$ sudo dnf -y install docker-ce-17.09.0.ce-1.fc26
```

A complication is that the version of docker used for swarm may be different to this version. Today, swarm will only work with docker `1.12.1`.

This also might have a CI impact.

/cc @sboeuf, @mcastelino, @grahamwhaley.

jodh-intel:
Darn. The number of the beast :japanese_ogre:. I should have raised this yesterday! :smile:

jodh-intel:
Idea: we could create an `manage_ctr_mgr.sh` script:

```
$ manage_ctr_mgr.sh [docker|kubernetes] [install|list|configure|remove]
```

... which when run as:

```
$ manage_ctr_mgr.sh docker install
```

... would do the following:

- if docker already installed, `exit 0`.
- if docker not installed, attempt to find `versions.txt` and install `docker_version=`.

But if run as, say,

```
$ manage_ctr_mg.sh docker install --tag swarm --force
```

... it would:

- remove an existing docker (we'd need a regex to find `docker.io`, `docker-engine`, `docker-ce` packages, etc).
- attempt to find `versions.txt` and install `docker_${tag}_version=`.

It will need to deal with a few things:

- `sudo`.
- the `docker` group (maybe?)
- it will need a `remove` option.
- it could also have a `configure --runtime=...` option.
- it should call `logger -s -t $program_name" ...` for any operation that modifies the version of docker on the system (`install`, `configure`, `remove`) so we have a trace of what it did.
- also handle k8s! ;)

Thoughts?

sboeuf:
@jodh-intel This proposal sounds good to me. All cases seem to be covered here!

jodh-intel:
Fixed by #800.

PR #762, opened 2017-11-30
ci: remove test-versions.txt

chavafg:
This commit removes `test-versions.txt` file, which we used for
obtaining versions for components and dependencies. We now will
rely on the clearcontainers/runtime versions.txt.

Also, some variables names in the CI scripts have been updated
to match the ones that reside in versions.txt

Fixes: #759.


.ci/openshift_setup.sh
```diff
@@ -16,9 +16,14 @@
 
 set -e
 
-SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source /etc/os-release
-source "$SCRIPT_PATH/../test-versions.txt"
+
+# source versions.txt from clearcontainers/runtime repository
+runtime_repo="github.com/clearcontainers/runtime"
+if [ ! -f "$GOPATH/src/${runtime_repo}/versions.txt" ]; then
```
MarioCarrilloA: Hi @chavafg , how about if `$GOPATH/src/${runtime_repo}/versions.txt` file exists, but the repo is old, should the repo do a `git pull` instead?

jodh-intel: That's a good point. For CI systems, this shouldn't happen as we create a new environment for every PR, but we probably should do this anyway for the developer system scenario. In fact, how about changing this PR slightly to reduce the boiler-plate code. We can "edge" slowly towards a [`manage_ctr_mgr.sh`](#666) by creating a script (called `foo.sh` below but call it whatever you think is appropriate at this stage) that is just:

```
set -e

_runtime_repo="github.com/clearcontainers/runtime"
go get -d -u -v "$_runtime_repo" || true

[ ! -f "$_versions_file" ] && { echo >&2 "ERROR: cannot find $_versions_file"; exit 1; }
_versions_file="$GOPATH/src/${_runtime_repo}/versions.txt"

source "$_versions_file"

unset _versions_file
unset _runtime_repo
```
Then, all the other scripts just need to do:

```
source foo.sh
```
... and this should DTRT (TM) :smile:

That will minimise the amount of code required in the calling scripts and also "hide" the location of `versions.txt` so that we can change it at will later if required with no impact on the callers. In fact, I wonder if it would be worth wrapping the guts of `foo.sh` into a function so it becomes something like:

```
set -e

_runtime_repo="github.com/clearcontainers/runtime"
_versions_file="$GOPATH/src/${_runtime_repo}/versions.txt"

get_cc_versions()
{
    go get -d -u -v "$_runtime_repo" || true

    [ ! -f "$_versions_file" ] && { echo >&2 "ERROR: cannot find $_versions_file"; exit 1; }

    source "$_versions_file"
}

unset _versions_file
unset _runtime_repo
```

and for the callers:

```
source foo.sh
get_cc_versions

# use $docker_version, etc
```

The advantage of that being we could start to introduce parameters for `get_cc_versions()` at a later date.

chavafg: I like the idea, will add the function to  `.ci/lib.sh`

chavafg:
@jodh-intel, @MarioCarrilloA addressed your comments, please re-review.

PR #772, opened 2017-12-05
Fix metrics tests for headless operation

grahamwhaley:
A number of the metrics tests do not work under headless (such as some CI systems or running under `nohup`) systems.
Fix up a number of the tests to work (predominantly by not using `-ti` on docker run or trying to use long-running 'shell's to keep a container 'open' (switch to using `tail -f /dev/null`).


chavafg:
lgtm

PR #773, opened 2017-12-06
metrics: update/refresh docker images before running tests

grahamwhaley:
Most of the tests did not check to see if there was an update to the docker images they were using.
Fix this by adding a generic function to do the check, and invoke it from the tests.

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -73,6 +73,29 @@ function check_cmds()
 	done
 }
 
+# This function performs a docker pull on the image names
+# passed in (notionally as 'about to be used'), to ensure
+#  - that we have the most upto date images
+#  - that any pull/refresh time (for a first pull) does not
+#    happen during the test itself.
+#
+# The image list can be received standalone or as an array, e.g.
+#
+# images=(“img1” “img2”)
+# check_imgs "${images[@]}"
+function check_images()
+{
+	local img req_images=( "$@" )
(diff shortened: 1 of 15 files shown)


grahamwhaley:
@GabyCT once this has landed, when you are adding the `init_env` calls to the remote network tests, maybe you can also add the `check_images` calls as well.

metrics/lib/common.bash
```diff
@@ -73,6 +73,29 @@ function check_cmds()
...
+# The image list can be received standalone or as an array, e.g.
+#
+# images=(“img1” “img2”)
+# check_imgs "${images[@]}"
+function check_images()
+{
+	local img req_images=( "$@" )
+	for img in "${req_images[@]}"; do
+		echo "docker pull'ing: $img"
+		if ! docker pull "$img" > /dev/null 2>&1; then
```
jodh-intel: How about we drop the `stdout` redirect so that the tests always log the image revision (which would potentially help with problem diagnosis).

grahamwhaley: Agreed - I'd considered that more for the 'show me what went wrong' factor as well - will update.

grahamwhaley: updated and pushed - there was no trivial way to just get a nice synopsis from `docker pull`, so I've removed all redirect and now it is a touch verbose. Let's see how that pans out - and if we have to in the future we can filter it with some greppy'ness maybe.

Issue #775, opened 2017-12-06
change cc-proxy log gathering 


chavafg:
Now that `cc-proxy` will no longer be a service, we need to update how to gather the proxy log in the `teardown.sh` script
This now should be: `sudo journalctl -t cc-proxy`

PR #776, opened 2017-12-06
ci: Update how journalctl gets the cc-proxy log

chavafg:
Now that the cc-proxy will not be a service, we need to use
option -t in journalctl to get the logs from the proxy.
Relates to:
https://github.com/clearcontainers/runtime/pull/835
https://github.com/clearcontainers/proxy/pull/177

Fixes: #775


devimc:
lgtm

PR #777, opened 2017-12-07
metrics: Add init_env function to remote tests

### LINKED ISSUE #766: metrics: remote networking: no calls to `init_env`
None of the remote networking metrics tests call `init_env` from the common metrics routines. They probably should to ensure we have a clean and stable environment, at least on the host, before we invoke the tests.
It is unclear what to do on the server side, if anything, though.

GabyCT:
Add init_env to the remote networking tests.

Fixes #766

### DIFF
--- a/metrics/network/remote_network/remote-networking-iperf3.sh
+++ b/metrics/network/remote_network/remote-networking-iperf3.sh
@@ -61,6 +61,8 @@ EOF
 function remote_network_bandwidth_iperf3 {
 	test_name="Remote network bandwidth"
 	get_runtime
+	init_env
+	init_env_remote
 	setup_swarm
 	client_replica_status
 	server_replica_status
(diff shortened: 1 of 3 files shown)


metrics/network/remote_network/remote-networking-iperf3.sh
```diff
@@ -149,8 +157,6 @@ function remote_network_latency {
 }
 
 function main {
-	[[ $# -ne 7 ]]&& help && die "Illegal number of parameters."
```
grahamwhaley: Did you mean to sneak a couple of other changes in here?
If so, can you break them out into a separate commit pls?

metrics/network/remote_network/remote-networking-iperf3.sh
```diff
@@ -61,6 +61,8 @@ EOF
 function remote_network_bandwidth_iperf3 {
 	test_name="Remote network bandwidth"
 	get_runtime
+	init_env
```
grahamwhaley: Unless it happens from a generic lib function, can you consider adding the `check_images` calls as well as added in #773

PR #783, opened 2017-12-08
metrics: Changes to the remote networking tests

GabyCT:
Add init_env and init_env_remote functions with the purpose of ensure a clean testing environment.
Add check_images function to retrieve the latest of the image from Docker Hub.
Changes to the main functions exit codes as well error messages.

### DIFF
--- a/metrics/network/remote_network/remote-networking-cpu.sh
+++ b/metrics/network/remote_network/remote-networking-cpu.sh
@@ -62,8 +62,6 @@ function remote_network_cpu_consumption {
 	test_name="Remote Network CPU Consumption"
 	units="%"
 	get_runtime
-	init_env
-	init_env_remote
 	setup_swarm
 	client_replica_status
 	server_replica_status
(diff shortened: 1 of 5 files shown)


devimc:
lgtm

GabyCT:
@grahamwhaley, I did all the changes as you mentioned thanks

metrics/network/remote_network/remote-networking-cpu.sh
```diff
@@ -62,8 +62,6 @@ function remote_network_cpu_consumption {
 	test_name="Remote Network CPU Consumption"
 	units="%"
 	get_runtime
-	init_env
```
grahamwhaley: /me mumbles that we don't normally wrap the subject line of commits into the body text... ;-)

grahamwhaley: /me then mumbles at github gui for showing that as a split line, when in fact it is not (although it may be a little long).... :-/

Issue #787, opened 2017-12-11
metrics: common: do not fail if pgrep fails


grahamwhaley:
We use `pgrep` in the common routines to report information on any errant/hung processes - but, sometimes by the time we call `pgrep`, the process has vanished (it would seem there is a window/race where we see a process around just as it is quitting...). currently the `-e` in bash means we quit, which we need not do, as this is not really a fatal error.
Make the `pgrep` failure non-fatal.

PR #788, opened 2017-12-11
metrics: common: do not fail if pgrep fails

grahamwhaley:
We use pgrep to report diagnostics on errant processes,
but sometimes those processes have gone by the time we
go looking for them. In that case pgrep fails, but is non
fatal.
Code it as so.

Fixes: #787

### DIFF
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -270,14 +270,14 @@ function kill_processes_before_start() {
 		# Sometimes we race and the process has gone by the time we list
 		# it - so make a pgrep fail non-fatal
 		pgrep -a -f "$HYPERVISOR_PATH" || true
-		sudo killall "$HYPERVISOR_PATH" || true
+		sudo killall -9 "$HYPERVISOR_PATH" || true
 	fi
 
 	result=$(check_active_process "$CC_SHIM_PATH")
 	if [[ $result -ne 0 ]]; then
 		warning "Found unexpected ${CC_SHIM_PATH} processes present"
 		pgrep -a -f "$CC_SHIM_PATH" || true
-		sudo killall "$CC_SHIM_PATH" || true
+		sudo killall -9 "$CC_SHIM_PATH" || true
 	fi
 }
 
--- a/metrics/lib/common.bash
+++ b/metrics/lib/common.bash
@@ -267,15 +267,17 @@ function kill_processes_before_start() {
 	result=$(check_active_process "$HYPERVISOR_PATH")
 	if [[ $result -ne 0 ]]; then
 		warning "Found unexpected ${HYPERVISOR_PATH} processes present"
-		pgrep -a -f "$HYPERVISOR_PATH"
-		sudo killall "$HYPERVISOR_PATH"
+		# Sometimes we race and the process has gone by the time we list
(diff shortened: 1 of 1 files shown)


grahamwhaley:
marking WIP - I missed a case - will update...

metrics/lib/common.bash
```diff
@@ -267,7 +267,9 @@ function kill_processes_before_start() {
 	result=$(check_active_process "$HYPERVISOR_PATH")
 	if [[ $result -ne 0 ]]; then
 		warning "Found unexpected ${HYPERVISOR_PATH} processes present"
-		pgrep -a -f "$HYPERVISOR_PATH"
+		# Sometimes we race and the process has gone by the time we list
+		# it - so make a pgrep fail non-fatal
+		pgrep -a -f "$HYPERVISOR_PATH" || true
```
jodh-intel: If this races due to the process(es) exiting, won't the `killall` below also fail (so doesn't that also need `|| true`?)

grahamwhaley: you're probably right - let me check the man for killall, but I'll probably add it anyway. thx

grahamwhaley: Yep, even with the `sudo` wrapper I think you are right, the killall could fail (although I have not see that in testing (mostly due to pgrep failing first)). Fixed, pushed.

metrics/lib/common.bash
```diff
@@ -267,15 +267,17 @@ function kill_processes_before_start() {
 	result=$(check_active_process "$HYPERVISOR_PATH")
 	if [[ $result -ne 0 ]]; then
 		warning "Found unexpected ${HYPERVISOR_PATH} processes present"
-		pgrep -a -f "$HYPERVISOR_PATH"
-		sudo killall "$HYPERVISOR_PATH"
+		# Sometimes we race and the process has gone by the time we list
+		# it - so make a pgrep fail non-fatal
+		pgrep -a -f "$HYPERVISOR_PATH" || true
+		sudo killall "$HYPERVISOR_PATH" || true
```
devimc: ```killall -9``` ?

grahamwhaley: Presumably you mean as a 'stronger kill', and not a way to avoid non-0 return codes from `killall` if the process has already vanished?
I'm torn - on the one hand I'd like to be pretty sure any stray things are torn down, but not given them a chance to die semi-gracefully could leave some of our CC state in a mess, maybe.

I'm going to leave as-is for now, as I think if we start seeing a stream of things we really cannot kill then we'd like to know. At that point then I'll consider moving to something like:
```
killall $blah || true
sleep3
kill -9 $blah || true
```
OK?

devimc: you are killing to the hypervisor, I guess you'll see a mess in the runtime, I'd prefer to use -9 (just in case the process is a zombie and ignore the term signal)

grahamwhaley: hmmmmm - ok - I'll bite. Yes, it's the runtime that will get more messed up than anything. I'll apply `-9` to both the hypervisor and the shim code below it. Expect a re-push ...

devimc:
lgtm

PR #790, opened 2017-12-11
metrics: Remove mount commands for iperf3.

### LINKED ISSUE #789: metrics: remove mount commands for iperf3 in remote networking tests
Change mount commands in the remote networking tests

GabyCT:
Remove mount commands for iperf3 in remote networking tests.

Fixes #789

### DIFF
--- a/metrics/network/remote_network/remote-networking-test-common.sh
+++ b/metrics/network/remote_network/remote-networking-test-common.sh
@@ -36,8 +36,9 @@ second_service="testswarm2"
 port_first_service="8080:80"
 # Port where the second service will run
 port_second_service="8081:80"
-# Mount command for iperf3 (cc-runtime)
-mount="mount -t ramfs -o size=20M ramfs /tmp"
+# Set the TMPDIR to an existing tmpfs mount to avoid
+# a 9p unlink error
+init_cmds="export TMPDIR=/dev/shm"
 
 # This function will retrieve the runtime from
 # host A and host B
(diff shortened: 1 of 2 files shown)


metrics/network/remote_network/remote-networking-test-common.sh
```diff
@@ -135,8 +136,8 @@ function start_server {
 	elif [ "$runtime_server" == "cc-runtime" ]; then
 		ssh "$ssh_user"@"$ssh_address" 'DOCKER_EXE=docker; \
 			server_id=$($DOCKER_EXE ps -q); \
-			mount="mount -t ramfs -o size=20M ramfs /tmp"; \
-			server_command="$mount && iperf3 -s"; \
+			init_cmds="export TMPDIR=/dev/shm"; \
```
devimc: you have already defined ```init_cmds``` in the line 41

GabyCT: @devimc yes, however, here the command will run in another host, so I need to define again the variable

devimc: yep, but I think you could do something like this

```
init_cmds="$init_cmds"
```

if TMPDIR change in the future, you only will change its path in the line 41

devimc:
lgtm

metrics/network/remote_network/remote-networking-test-common.sh
GabyCT: @devimc the problem is that the commands should be using "" however that will take the variables of the host a

PR #802, opened 2017-12-14
RFC: add sysbench cpu and memperf metrics

grahamwhaley:
Add a couple of simple in-container cpu performance and memory bandwidth tests based around `sysbench`.

### DIFF
--- a/Dockerfiles/sysbench/Dockerfile
+++ b/Dockerfiles/sysbench/Dockerfile
@@ -1,0 +1,22 @@
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Set up an Ubuntu image with 'sysbench' installed
+
+FROM ubuntu
+
+RUN apt-get update && \
+	apt-get install -y sysbench
+
+CMD ["/bin/bash"]
--- a/metrics/cpu/sysbench_cpu.sh
+++ b/metrics/cpu/sysbench_cpu.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


grahamwhaley:
@MarioCarrilloA @chavafg I've marked this WIP/RFC as I'd like your input on the way I've added a dockerfile and the (not quite perfect) way the test will auto-build and inject the image locally if it does not exist before the test is run.

metrics/cpu/sysbench_cpu.sh
```diff
@@ -0,0 +1,68 @@
...
+CMD="sysbench --test=cpu run"
+
+# Check if we have our local image installed, and if not, install it
+# from the dockerfile
+function local_check_images()
+{
+	local exists=$(docker image inspect $IMAGE > /dev/null; echo $?)
+
+	if [ $exists -ne 0 ]; then
+		docker build --label "$IMAGE" --tag "${IMAGE}:latest" "$DOCKERFILE"
```
MarioCarrilloA: this is great!

grahamwhaley: Thanks :-)  The only thing it does not do is manage 'versions', so would not update the image if we changed the Dockerfile... I've not looked into how we'd do that - if we can embed a version into the Dockerfile (and have to remember to update it if we modify it) or not.
Another thought might be that maybe an Image contains some short of SHA of the dockerfile it came from, and then we could check that. I dunno - anybody?
Unless we know of a moderately simple way to manage updates of Dockerfile/images (without using a hub), then I suspect this will do 'for now'.

MarioCarrilloA (approved):

Dockerfiles/sysbench/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2017 Intel Corporation
```
devimc: 2018

metrics/cpu/sysbench_cpu.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
```
devimc: 2018

metrics/memperf/sysbench_memory.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
```
devimc: 2018

grahamwhaley: well.... nope - the patches were submitted in Dec 2017 ;-)

grahamwhaley: sure, if I need to re-write them, I'll try to remember to update the copyright

metrics/cpu/sysbench_cpu.sh
```diff
@@ -0,0 +1,68 @@
...
+function main()
+{
+	cmds=("awk")
+
+	init_env
+	check_cmds "${cmds[@]}"
+	local_check_images "$IMAGE"
+
+	# Now we run the prep task
+	local output=$(docker run --rm $IMAGE $CMD)
```
grahamwhaley: I was about to nudge this along (as I was copying some of the magic from it), when I realised it does not use `$RUNTIME` during the docker run - so, I'll need to do a touch of rework here anyhow.... update soon then I hope.

grahamwhaley:
Closing this, as all work has moved to https://github.com/kata-containers.

PR #808, opened 2017-12-21
integration: Add tests for "docker exec --user"

### LINKED ISSUE #805: Create a docker exec test specifying an explicit user
We need some new tests in `integration/docker/exec_test.go` that run

```
docker exec --user $user $container ...
```
Where `$user` is:

- A username
- A UID
- A username + groupname
- A UID + GID

See: https://github.com/clearcontainers/runtime/issues/867

jodh-intel:
Add some new ginkgo integration tests to ensure
`docker exec --user= ...` works as expected.

Fixes #805.

### DIFF
--- a/integration/docker/exec_test.go
+++ b/integration/docker/exec_test.go
@@ -15,10 +15,17 @@
 package docker
 
 import (
+	"fmt"
+
 	. "github.com/clearcontainers/tests"
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 )
+
+func withUser(user, regexp string) TableEntry {
+	return Entry(fmt.Sprintf("with --user='%s'", user), user, regexp)
+}
 
 var _ = Describe("docker exec", func() {
 	var (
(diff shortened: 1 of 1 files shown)


jodh-intel:
Note: these tests should fail until https://github.com/clearcontainers/runtime/pull/873 lands.

jodh-intel:
And they are indeed failing as expected since currently the runtime will always run all `exec --user=` as `root:root`:

```
    Expected
        <string>: uid=0(root) gid=0(root) groups=10(wheel)
        
    to match regular expression
        <string>: uid=\d+\(daemon\) gid=\d+\(daemon\)

```

integration/docker/exec_test.go
```diff
@@ -79,4 +81,91 @@ var _ = Describe("docker exec", func() {
...
+
+		It(fmt.Sprintf("should run the command with uid %s and gid %s", uid, gid),
+			func() {
+				args = []string{"-t", "--user", fmt.Sprintf(":%s", gid), id, "id"}
+				stdout, stderr, exitCode = DockerExec(args...)
+				Expect(exitCode).To(Equal(0))
+				Expect(stderr).To(BeEmpty())
+				Expect(stdout).To(MatchRegexp(fmt.Sprintf("uid=%s.*gid=%s", uid, gid)))
+			})
+	})
```
devimc: I think a table is the best option for this kind of tests

```go
function withUser(user, regexp string) TableEntry {
    return Entry(fmt.Sprintf("with --user='%s'", user), user, regexp)
}

DescribeTable("check exec honours '--user'", func(user, regexp string) {
        args = []string{"-t", "--user", user, id, "id"}
        stdout, stderr, exitCode = DockerExec(args...)
        Expect(exitCode).To(Equal(0))
        Expect(stderr).To(BeEmpty())
        Expect(stdout).To(MatchRegexp(regexp))
    },
    withUser(fmt.Sprintf(":%s", 999), fmt.Sprintf("uid=%s.*gid=%s", 0, 999)),
    withUser(fmt.Sprintf("root:adm"), fmt.Sprintf(`uid=\d+\(root\) gid=\d+\(adm\)`)),
    .............
}
```

jodh-intel: Neat, thanks. I didn't know you could do that! Branch updated :)

devimc:
@jodh-intel  cool, even you added more tests :)

lgtm

PR #822, opened 2018-01-05
integration/docker: Add CPU constraint test

### LINKED ISSUE #821: Add test to ensure CPU quota and period are honoured
using docker, run a container with next options ```--cpu-period``` and ```--cpu-quota``` and check the CPU constraint

devimc:
this test ensures CPU quota and period are honoured

fixes #821

### DIFF
--- a/integration/docker/run_test.go
+++ b/integration/docker/run_test.go
@@ -17,7 +17,9 @@ package docker
 import (
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
+	"strings"
 
 	. "github.com/clearcontainers/tests"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


integration/docker/run_test.go
```diff
@@ -204,3 +204,34 @@ var _ = Describe("run", func() {
...
+		vCPUs  uint
+	)
+
+	BeforeEach(func() {
+		quota = 50000
+		period = 20000
+		vCPUs = (quota + period - 1) / period
+		id = RandID(30)
+		args = []string{"--rm", "--name", id, "--cpu-quota", fmt.Sprintf("%d", quota),
+			"--cpu-period", fmt.Sprintf("%d", period), Image, "nproc"}
```
jodh-intel: This looks fine, but I think it would be good to test the vcpu count for various quota and period values including values which would nominally result in a vcpu count of `< 0`, `0`, `1`, `$cpus` and `$cpus + 1`, where `$cpus == $number_of_host_cpus`. That will test boundary conditions, ensure that a reasonable default gets chosen (or a reasonable error gets returned for unreasonable quota/period values), and also will ensure that we didn't just "get lucky" where `nproc` in the container happens to match `nproc` on the host.

integration/docker/run_test.go
```diff
@@ -204,3 +204,34 @@ var _ = Describe("run", func() {
...
+		quota  uint
+		vCPUs  uint
+	)
+
+	BeforeEach(func() {
+		quota = 50000
+		period = 20000
+		vCPUs = (quota + period - 1) / period
+		id = RandID(30)
+		args = []string{"--rm", "--name", id, "--cpu-quota", fmt.Sprintf("%d", quota),
```
grahamwhaley: Just to note, according to the Docker docs, it seems `quota` and `period` may be semi-deprecated
https://docs.docker.com/engine/admin/resource_constraints/#cpu
https://blog.docker.com/2017/01/cpu-management-docker-1-13/

Maybe we should re-write, or at least add, `--cpus` based tests (which I believe will be translated by docker into `quota` and `period` values to the runtime itself?)

devimc: ohh I see, thanks @grahamwhaley

devimc:
@grahamwhaley  @jodh-intel  changes applied, thanks

chavafg:
lgtm

Issue #824, opened 2018-01-08
checkmetrics: nobaseline: prints out results with infinite precision


grahamwhaley:
well, maybe not infinite, but is uses '-1' as the precision to `FormatFloat`, whereas for the other report generating we use `3`.
https://github.com/clearcontainers/tests/blob/master/cmd/checkmetrics/main.go#L163

To make the report more readable, let's move to using `3` there as well.

PR #825, opened 2018-01-08
checkmetrics: nobaseline: set precison of output table data

grahamwhaley:
The output table data was being printed with 'infinite' precision,
which leads to a very noisy table (like 16+ digits of precision).
Let's make the table more readable (and more likely to fit on an
80 column screen) by setting precision to 3 - we really/rarely need
to see more than that.

Fixes: #824


devimc:
lgtm

chavafg:
lgtm

PR #828, opened 2018-01-09
ci: Remove legacy code related to one proxy per host

### LINKED ISSUE #827: ci: Remove legacy code related to one proxy per host
We don't want to keep some legacy code which will let several instances of the proxy running.

sboeuf:
This commit removes legacy code when Clear Containers was relying on
one proxy per host. Now that we rely on one proxy per VM, there is no
point in keeping this.

Fixes #827

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

### DIFF
--- a/.ci/install_proxy.sh
+++ b/.ci/install_proxy.sh
@@ -22,17 +22,3 @@ cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
 
 clone_build_and_install "github.com/clearcontainers/proxy"
-systemd_extra_flags="-log debug"
-
-# If we are running under the metrics CI system then we do not want the proxy
-# to be dynmaically changing the KSM settings under us - we need control of them
-# ourselves
-if [[ $METRICS_CI ]]; then
-	systemd_extra_flags="${systemd_extra_flags} -ksm initial"
-fi
-
-proxy_systemd_file=$(sudo systemctl show cc-proxy | fgrep FragmentPath | awk 'BEGIN{FS="="}{print $2}')
-sudo sed -i "s/\(^ExecStart.*$\)/\1 ${systemd_extra_flags}/" "${proxy_systemd_file}"
-sudo systemctl daemon-reload
-echo "Start proxy service"
-sudo systemctl restart cc-proxy


jodh-intel:
This looks fine. However, we potentially want to apply these settings to the ksm-throttler now.

but note: https://github.com/clearcontainers/packaging/issues/225.

Also, the throttler doesn't seem to have the KSM throttler option that the proxy has?

Another approach for QA/PnP might be to simply uninstall the throttler (once the packaging bug above is resolved). What do you think @chavafg, @grahamwhaley ?

devimc:
lgtm

grahamwhaley:
I've not stared at the throttler - is it some sort of systemd service that we could just (even temporarily) disable with a `systemctl` whilst we run the metrics - if so, that would work even better than before I think...

jodh-intel:
@grahamwhaley - yes, that's another option of course. Note that currently it needs to be manually installed and manually enabled due to the packaging issue. Hence, perfect currently for QA/PnP, but soon not to be ;)

sboeuf:
@gwhaley @jodh-intel I don't think this change impacts the metrics because this code is not used anymore since we have switched to one proxy per VM, right ?
Also, notice that I have remove this code since I wanted to make sure we don't introduce some potential issues by having one proxy running but *not* used along with other proxies (started by the runtime for every VM.)

jodh-intel:
What I meant is that ideally we'd have a PR that switched these changes from `cc-proxy` to `ksm-cc-throttler`. However, we're essentially blocked on the ability to do that due to the packaging bug.

I guess we can land this now, but we need to ensure we don't forget to apply these changes to the throttler once the packaging bug is resolved.

lgtm

sboeuf:
@jodh-intel oh yes I get it. What do you think about commenting out this code then ?

jodh-intel:
I guess we could do that with a comment explaining why. But as long as we've got an issue to remind us to apply these changes in reverse to the throttler that'd wfm.

sboeuf:
Ok I'll open this issue

sboeuf:
@jodh-intel here is the issue: https://github.com/clearcontainers/tests/issues/829

sboeuf:
@chavafg if you're okay with this patch, feel free to force merge it since F26 failure is due to Openshift.

chavafg:
lgtm

PR #836, opened 2018-01-10
test: Add docker integration no-new-privileges test.

### LINKED ISSUE #811: Add test to check a container process cannot gain additional privileges when no-new-privileges flag is passed on the docker command line.
Related changes that need to land:
https://github.com/containers/virtcontainers/pull/529
https://github.com/clearcontainers/agent/pull/186

The test would include starting a container as a non-root user, and then on run a setuid program. The process would not be able to gain root privileges. 
Please refer to this blog that details this:
https://www.projectatomic.io/blog/2016/03/no-new-privs-docker/

GabyCT:
This test will verify that a container process do not gain
additional privileges while running no-new-privileges with docker.

Fixes #811


GabyCT:
This test is following this approach  https://www.projectatomic.io/blog/2016/03/no-new-privs-docker/

integration/docker/privileges_test.go
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2017 Intel Corporation
```
jodh-intel: Nit: 2018.

integration/docker/privileges_test.go
```diff
@@ -0,0 +1,88 @@
...
+		Expect(ExistDockerContainer(id)).NotTo(BeTrue())
+		Expect(RemoveDockerContainer(secondId)).To(BeTrue())
+		Expect(ExistDockerContainer(secondId)).NotTo(BeTrue())
+		_, _, exitCode := DockerRmi(testImage)
+		Expect(exitCode).To(Equal(0))
+	})
+
+	Context("check no-new-privileges flag", func() {
+		It("should display the correct effective uid", func() {
+			Skip("Issue https://github.com/clearcontainers/runtime/issues/880")
```
jodh-intel: Nice. And this has inspired me to raise #837 which would be a super-useful little facility :smile:

integration/docker/privileges_test.go
```diff
@@ -0,0 +1,88 @@
...
+			id = randomDockerName()
+			args = []string{"-d", "--name", id, GccImage, "bash", "-c", "sleep 30"}
+			_, _, exitCode = DockerRun(args...)
+			Expect(exitCode).To(Equal(0))
+			args = []string{id, "bash", "-c", "echo -e '#include<stdio.h>\n#include <unistd.h>\n#include <sys/types.h>\nint main (int argc, char *argv[])\n{printf(\"Effective uid: %d\", geteuid());return 0;}' > demo.c && make demo"}
```
jodh-intel: I'm not really clear why it's necessary to build our own binary. Can't we just use a standard image with a standard version of `/bin/id` (which shows the EUID) and just create a container, copy `/bin/id` to `/bin/id-foo` or something, `chmod -s` it, commit the image, and re-run the test using that new image and passing `--security-opt=no-new-privileges`?

GabyCT:
Hi @jodh-intel the problem is that the /usr/bin/id is not working correctly, if I do the chmod +s /usr/bin/id of a Fedora container and then I export the container and then import the image, I obtained the following
$ docker run -t --user 1000  --security-opt=no-new-privileges intento /usr/bin/id
uid=1000 gid=0(root) groups=0(root)

$ docker run -t --user 1000 intento /usr/bin/id
uid=1000 gid=0(root) euid=0(root) groups=0(root)

In both cases the uid is 1000 (running with the flag or without).

GabyCT:
@jodh-intel , this was verified using
cc-runtime  : 3.0.12
commit   : 9014fe377ba9db875e08b28a801164febad43264

amshinde:
@GabyCT The above behaviour is actually correct. /usr/bin/uid reports the effective user id only if its different from the real user id (https://docs.oracle.com/cd/E19683-01/816-0211/6m6nc66s5/index.html) So,
```
$ docker run -t --user 1000  --security-opt=no-new-privileges intento /usr/bin/id
uid=1000 gid=0(root) groups=0(root)
```
Means that the effective user id (euid) is 1000, which is what we expect from the no-new-privileges flag.
So, in your test using a modified /usr/bin/uid with setuid bit set, you should check that there is no euid string present in the output when no-new-privileges flag is passed and uid is ofcourse 1000.

Also, this PR to vendor virtcontainers in the proxy needs to be merged:
https://github.com/clearcontainers/proxy/pull/196

jodh-intel:
:electric_plug: Shameless plug :electric_plug:

If you want a clearer confirmation of which user you are running as, you could use a `debian`, `ubuntu`, `fedora` image and install my `procenv` tool:

```
$ sudo apt-get -y install procenv
$ cp `which procenv` /tmp
$ sudo chown root:root /tmp/procenv
$ sudo chmod +s /tmp/procenv
$ /tmp/procenv --process|grep uid
    real user id (uid): 1000 ('james')
    effective user id (euid): 0 ('root')
    saved set-user-id (suid): 0 ('root')
```

Issue #841, opened 2018-01-12
metrics: launch_times: don't dmesg for non-VM runtimes


grahamwhaley:
We grab a `dmesg` from the container during the launch_times.sh test, but we only use this for data gathering for VM runtimes.
The dmesg from a non-VM runtime may be large (as it is actually from the host, not the container boot), and we have seen having the dmesg in place there in the test cause some issues.

Thus, do not invoke the dmesg in the test if we don't need it.
This does mean the tests between runtimes may be running slightly different container commands, but given the dmesg will act differently between VM and non-VM runtimes anyhow, it probably makes little actual change to the test.

PR #842, opened 2018-01-12
metrics: launch_times: do not use dmesg for non-VM runtimes

grahamwhaley:
We don't use the output of the container `dmesg` if we are not
under a known VM runtime, and we have seen it's inclusion cause
some issues on other runtimes (apparent hangups), so let's drop
the dmesg if we are not under a known VM runtime.

Fixes: #841


jcvenegas:
lgtm

devimc:
lgtm

PR #847, opened 2018-01-12
tests: Add docker exit code tests

### LINKED ISSUE #846: Add docker tests for checking exit status codes
Verify that container gives the correct exit code when running in interactive and non-interactive mode.

chavafg:
These tests verify that container gives the correct exit code when
running in interactive and non-interactive mode.

Fixes #846.

### DIFF
--- a/integration/docker/exit_code.go
+++ b/integration/docker/exit_code.go
@@ -1,0 +1,68 @@
+// Copyright (c) 2018 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


devimc:
lgtm

GabyCT:
lgtm

jodh-intel:
This looks fine to me. But how about adding in a range of exit codes to check (for both non-interactive and interactive)?

Basically adding in a "table" as @devimc suggested I do for #808. That would allow you to do some boundary testing to ensure we handle `exit` values like `-1` (should return 254), `0`, `1`, `55`, `255`, `256` (should return `0`).

chavafg:
Agree, changes applied :)

PR #854, opened 2018-01-16
tests: Add docker run with stdin.

### LINKED ISSUE #853: add stdin test for docker run
add a docker running integration test with stdin

GabyCT:
Test stdin while running docker run.

Fixes #853

### DIFF
--- a/command.go
+++ b/command.go
@@ -55,11 +55,20 @@ func NewCommand(path string, args ...string) *Command {
 
 // Run runs a command returning its stdout, stderr and exit code
 func (c *Command) Run() (string, string, int) {
+	return c.RunWithPipe(nil)
+}
+
+// Run runs a command with stdin as an input and returning its stdout, stderr and exit code
+func (c *Command) RunWithPipe(stdin *bytes.Buffer) (string, string, int) {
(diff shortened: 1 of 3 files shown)


command.go
```diff
@@ -92,3 +92,44 @@ func (c *Command) Run() (string, string, int) {
 		return stdout.String(), stderr.String(), exitCode
```
devimc: could you please modify ```Run() ``` to use the new function ```RunWithPipe() ```

```go
func (c *Command) Run() (string, string, int) {
    return c.RunWithPipe(nil)
}
```

docker.go
```diff
@@ -51,10 +52,24 @@ func runDockerCommandWithTimeout(timeout time.Duration, command string, args ...
 	return cmd.Run()
```
devimc: also please update ```runDockerCommandWithTimeout```

```
func runDockerCommandWithTimeout(timeout time.Duration, command string, args ...string) (string, string, int) {
   return runDockerCommandWithTimeoutAndPipe(nil, timeout, command, args...)
}
```

GabyCT:
@devimc changes applied

devimc:
lgtm

Issue #867, opened 2018-01-22
Fix cri-o integration tests


chavafg:
CRI-O integration tests rely on `crioctl`, which was deprecated. Now `crictl` should be used.

chavafg:
The fix for this should land at the same time of https://github.com/clearcontainers/runtime/pull/942

chavafg:
branch_runtime: https://github.com/chavafg/runtime/tree/topic/update-versions

PR #870, opened 2018-01-23
Update scripts and test to work with latest CRI-O, k8s and openshift

chavafg:
Now that we are moving to CRI-O v1.9, the crioctl tool has
been deprecated, instead we need to use crictl for our tests.

Also, k8s v1.9.1 and openshift v3.7.1 changed the configuration slightly.
This PR addresses all issues to be able to move to these newer
versions. 

Depends on: 

Fixes: #867.

### DIFF
--- a/integration/kubernetes/init.sh
+++ b/integration/kubernetes/init.sh
@@ -18,7 +18,7 @@ set -e
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/lib.sh"
 
-sudo -E kubeadm init --pod-network-cidr 10.244.0.0/16
+sudo -E kubeadm init --pod-network-cidr 10.244.0.0/16 --cri-socket=/var/run/crio/crio.sock
 export KUBECONFIG=/etc/kubernetes/admin.conf
 
 sudo -E kubectl get nodes
--- a/integration/kubernetes/nginx.bats
+++ b/integration/kubernetes/nginx.bats
@@ -20,12 +20,15 @@ load "${BATS_TEST_DIRNAME}/lib.sh"
 setup() {
 	nginx_image="nginx"
 	busybox_image="busybox"
-	service_name="${nginx_image}-service"
+	service_name="nginx-service"
 	export KUBECONFIG=/etc/kubernetes/admin.conf
 	master=$(hostname)
 	sudo -E kubectl taint nodes "$master" node-role.kubernetes.io/master:NoSchedule-
-	sudo -E crioctl image pull "$busybox_image"
-	sudo -E crioctl image pull "$nginx_image"
+	# Pull the images before launching workload. This is mainly because we use
+	# a timeout and in slow networks it may result in not been able to pull the image
+	# successfully.
+	sudo -E crictl pull "$busybox_image"
+	sudo -E crictl pull "$nginx_image"
 }
 
 @test "Verify nginx connectivity between pods" {
--- a/.ci/install_crio.sh
+++ b/.ci/install_crio.sh
@@ -71,6 +71,9 @@ crio_config_file="/etc/crio/crio.conf"
 echo "Set runc as default runtime in CRI-O for trusted workloads"
 sudo sed -i 's/^runtime =.*/runtime = "\/usr\/local\/bin\/crio-runc"/' "$crio_config_file"
 
+echo "Add docker.io registry to pull images"
+sudo sed -i 's/^registries = \[/registries = \[ "docker.io"/' /etc/crio/crio.conf
+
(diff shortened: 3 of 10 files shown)


chavafg:
branch_runtime: https://github.com/chavafg/runtime/tree/topic/update-versions

integration/kubernetes/init.sh
```diff
@@ -18,7 +18,7 @@ set -e
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/lib.sh"
 
-sudo -E kubeadm init --pod-network-cidr 10.244.0.0/16
+sudo -E kubeadm init --pod-network-cidr 10.244.0.0/16 --ignore-preflight-errors=CRI
```
jcvenegas: why it is need to ignore CRI errors ?

integration/kubernetes/nginx.bats
```diff
@@ -18,14 +18,15 @@
...
+	image_registry="docker.io"
+	nginx_image="${image_registry}/nginx"
+	busybox_image="${image_registry}/busybox"
+	service_name="nginx-service"
 	export KUBECONFIG=/etc/kubernetes/admin.conf
 	master=$(hostname)
 	sudo -E kubectl taint nodes "$master" node-role.kubernetes.io/master:NoSchedule-
-	sudo -E crioctl image pull "$busybox_image"
-	sudo -E crioctl image pull "$nginx_image"
+	sudo -E crictl pull "$busybox_image"
```
jcvenegas: This should not be handled by k8s? (by calling  cri-o api)

integration/kubernetes/init.sh
chavafg: Because kubelet searches for `dockershim.sock`. As we do not use that socket and instead we use CRI-O, we need to ignore it.

integration/kubernetes/nginx.bats
chavafg: Yes, it can be handled by k8s, but as we have a timeout below, sometimes when running on a slow network, you may be able to see failures because the image couldnt be downloaded on time, that is why I pull them before launching the workload.

integration/kubernetes/init.sh
jcvenegas: oh thats true, I had the sam problem, I added `--cri-socket`
kubeadm init --pod-network-cidr=10.244.0.0/16 --cri-socket /var/run/crio.sock

jcvenegas: Same for reset:
kubectl drain $(hostname) --delete-local-data --force --ignore-daemonsets
kubectl delete node --all
kubeadm reset  --cri-socket /var/run/crio.sock

integration/kubernetes/nginx.bats
jcvenegas: Thanks for clarify, also would be good to add a comment inside.

integration/kubernetes/init.sh
```diff
@@ -18,7 +18,11 @@ set -e
 SCRIPT_PATH=$(dirname "$(readlink -f "$0")")
 source "${SCRIPT_PATH}/lib.sh"
 
-sudo -E kubeadm init --pod-network-cidr 10.244.0.0/16
+# Run init ignoring CRI preflight check.
+# the CRI preflight check verifies that /var/run/dockershim.sock exists. But we don't
+# need it as we use CRI-O for launching pods, then we need to disable this check,
+# otherwise, kubeadm will fail.
+sudo -E kubeadm init --pod-network-cidr 10.244.0.0/16 --ignore-preflight-errors=CRI
```
egernst: I would add ```--cri-socket=/var/run/crio/crio.sock``` to the init and the reset.  This tells kubeadm to use that crio socket instead of docker to reset the system.  Without this (at least on the reset), you'll need to do painful cleanup (as we have to do today) after tearing down the cluster.

chavafg: Agree @egernst @jcvenegas, changes applied.

egernst:
Thanks -- LGTM!

integration/kubernetes/nginx.bats
```diff
@@ -18,14 +18,18 @@
 load "${BATS_TEST_DIRNAME}/lib.sh"
 
 setup() {
-	nginx_image="nginx"
-	busybox_image="busybox"
-	service_name="${nginx_image}-service"
+	image_registry="docker.io"
```
egernst: @chavafg  -- should we move this to the crio.conf the we install?  If folks want to be able to follow the script to install CRIO and expect it to work (with K8S), they'll want this set.

chavafg: @egernst fixed :)

sboeuf:
LGTM

GabyCT:
lgtm

Issue #885, opened 2018-02-02
Donwload assets in tmpdir


jcvenegas:
`install_assets.sh` downloads assets in the current working directory use a tmpdir.

PR #886, opened 2018-02-02
ci: install assets use tmpdir

jcvenegas:
SHASUMS file exist in both assets.

To avoid issues when download file from kernel or image
use a tmpdir to download assets.

Fixes: #885


chavafg:
lgtm

PR #891, opened 2018-02-06
ctr_mgr: Remove newgroup command from installation

### LINKED ISSUE #890: remove newgrp command from the docker installation tool
newgrp command was introduced to login into the docker group and allow a normal user execute the runtime unit tests which require the use of docker without root, but this command didn't had the desired effect on the CI and when running under a terminal, it creates a new bash session, stopping the installation of all other components.

chavafg:
newgrp command was introduced to login into the docker
group and allow a normal user execute the runtime
unit tests which require the use of docker without
root, but this command didn't had the desired effect
on the CI and when running under a terminal it creates
a new bash session, stopping the installation of all
other components. For this reason, lets remove this
command from the docker installation.

Fixes #890.

### DIFF
--- a/cmd/container-manager/manage_ctr_mgr.sh
+++ b/cmd/container-manager/manage_ctr_mgr.sh
@@ -157,7 +157,6 @@ install_docker(){
 	sudo systemctl restart docker
 	sudo gpasswd -a ${USER} docker
 	sudo chmod g+rw /var/run/docker.sock
-	newgrp docker
 }
 
 # This function removes the installed docker package.


jodh-intel:
Presumably, all tests will now use `sudo` to access docker (which is probably safer anyway)?

lgtm

grahamwhaley:
just for reference, on the metrics CI we run the tests under the user 'jenkins', who is in the docker group on the CI machines - so, we probably won't notice if a test is not run using sudo.
running all the tests as sudo makes me a little nervous. OK, sure, the docker daemon has root privs, and if we only run the docker command as sudo then we should not litter the workspace with root owned/created files - it still makes me feel a little nervous though...

jcvenegas:
lgtm

GabyCT:
lgtm

Issue #897, opened 2018-02-07
pnp jenkins: make sure runc is default runtime to build image


jcvenegas:
Configure runc as default runtime to create new images.

PR #898, opened 2018-02-07
ci: Set runc as default runtime for image creation

jcvenegas:
jenkins may use cc-runtime as default runtime.
This is an issue when an image need to be created with osbuilder.

This commit tries to set runc as default runtime. If is not possible
(because docker not installed), the setup will continue.

Fixes: #897

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -79,6 +79,11 @@ then
         .ci/setup_tests.sh
 fi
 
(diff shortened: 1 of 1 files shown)


jcvenegas:
This is  need to have https://github.com/clearcontainers/runtime/pull/980 passing.

jodh-intel:
lgtm

/cc @grahamwhaley.

Issue #907, opened 2018-02-13
jenkins: manage_ctr_mgr.sh: No such file or directory


jcvenegas:
Jenkins setup fails to find manage_ctr_mgr.sh see:
https://clearlinux.org/cc-ci/job/clear-containers-runtime-16.04-PR/116/console

PR #908, opened 2018-02-13
ci: jenkins: fix path to manage_ctr_mgr.sh

jcvenegas:
Fix path to manage_ctr_mgr.sh

Fixes: #907

### DIFF
--- a/.ci/jenkins_job_build.sh
+++ b/.ci/jenkins_job_build.sh
@@ -82,7 +82,7 @@ fi
 # Make sure runc is default runtime.
 # This is needed in case a new image creation.
 # See https://github.com/clearcontainers/osbuilder/issues/8
-"${GOPATH}/${tests_repo}/cmd/container-manager/manage_ctr_mgr.sh" docker configure -r runc  -f || true
+"${GOPATH}/src/${tests_repo}/cmd/container-manager/manage_ctr_mgr.sh" docker configure -r runc -f
 
 # Call the repo-specific setup script.
 #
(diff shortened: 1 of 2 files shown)


.ci/jenkins_job_build.sh
```diff
@@ -82,7 +82,7 @@ fi
 # Make sure runc is default runtime.
 # This is needed in case a new image creation.
 # See https://github.com/clearcontainers/osbuilder/issues/8
-"${GOPATH}/${tests_repo}/cmd/container-manager/manage_ctr_mgr.sh" docker configure -r runc  -f || true
+"${GOPATH}/src/${tests_repo}/cmd/container-manager/manage_ctr_mgr.sh" docker configure -r runc  -f
```
grahamwhaley: If I were being picky :-), I think you have a 'double space' at the end before the `-f` :-)

grahamwhaley:
oh, that's interesting, the metrics ci failed (which is something this is fixing...)

jcvenegas:
@grahamwhaley  I see still fail for the same reason... 
.ci/jenkins_job_build.sh: line 85: /srv/jenkins/workspace/workspace/clear-containers-tests-16.04-PR/go/github.com/clearcontainers/tests/cmd/container-manager/manage_ctr_mgr.sh: No such file or directory

the path is not updated yet , because cloned master?

jcvenegas:
@chavafg  what is the issue on  k8s ci `kubernetes qa-failed`?

grahamwhaley:
@jcvenegas - heh heh, I think it is even a little more evil than that....
we start by checking out the master branch, and then run the `.ci/jenkins_job_build.sh` from that branch - and that script then switches branches.... but, it is 'too late', as we are already running that script, so we won't notice the update from the branch switch.

Thus, I'm happy the change is correct, and we understand why the metrics CI is not passing - I'm happy if we merge this, and re-visit if we are wrong and have to fix again...

PR #909, opened 2018-02-14
Refine metrics CI invocations

grahamwhaley:
Refine our metrics CI invocations somewhat:
- stop running tests that we don't currently check the results for
- refine the iperf3 test run to only test what we do check
    - which involved adding a commandline to the iperf3 test so we could tweak it

### DIFF
--- a/.ci/run_metrics_ci.sh
+++ b/.ci/run_metrics_ci.sh
@@ -125,18 +125,8 @@ pushd "$CURRENTDIR/../metrics"
 	# ops/second
 	bash network/network-nginx-ab-benchmark.sh
 
-	# ping latency
-	bash network/network-latency.sh
-
-	# qperf latency
-	bash network/network-latency-qperf.sh
-
-	# Bandwidth and jitter
+	# iperf3 bandwidth between host and container
 	bash network/network-metrics-iperf3.sh -B -i 8
-
-	# UDP bandwidths and packet loss
-	bash network/network-metrics-nuttcp.sh
-
 
 	#
 	# Run some IO tests
--- a/.ci/run_metrics_ci.sh
+++ b/.ci/run_metrics_ci.sh
@@ -132,7 +132,7 @@ pushd "$CURRENTDIR/../metrics"
 	bash network/network-latency-qperf.sh
 
 	# Bandwidth and jitter
-	bash network/network-metrics-iperf3.sh
+	bash network/network-metrics-iperf3.sh -B -i 8
 
 	# UDP bandwidths and packet loss
 	bash network/network-metrics-nuttcp.sh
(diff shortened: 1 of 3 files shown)


grahamwhaley:
A quick note on the README update - yes, the two spaces added to the end of a line in the bullet list are deliberate - they force a line break, and hence a new paragraph, within that bullet.
Dropping the WIP now...

.ci/run_metrics_ci.sh
```diff
@@ -125,18 +125,8 @@ pushd "$CURRENTDIR/../metrics"
 	# ops/second
 	bash network/network-nginx-ab-benchmark.sh
 
-	# ping latency
-	bash network/network-latency.sh
```
jodh-intel: I see that we still run these tests in `metrics/run_all_metrics.sh`, but I'm now wondering if we shouldn't just have a single parameterised script to handle all metrics tests?

grahamwhaley: as in a single script where we can say 'script --all' or 'script --kpis' for instance, to remove the duplication between the two scripts?

We could do that. In my mind we'd re-factor the run_all_metrics (well, and change the name). We'd have to add the KSM handling and some other trickery from the run_metrics_ci script as well (like the results storing options as well).

How about I open an Issue to note we could do with a refactor of the scripts?

jodh-intel: Thanks - Sounds good to me.

devimc:
lgtm

.ci/run_metrics_ci.sh
jodh-intel: See #912.

jodh-intel:
lgtm

@klynnrif - could you take a look please?

metrics/network/README.md
```diff
@@ -35,7 +35,9 @@ and packet-per-second throughput using iperf3 on single threaded connections.
 The bandwidth test shows the speed of the data transfer. On the other hand,
 the jitter test measures the variation in the delay of received packets.
 The packet-per-second tests show the maximum number of (smallest sized) packets
-we can get through the transports.
+we can get through the transports.
```
klynnrif: allowed through the transports.

metrics/network/README.md
```diff
@@ -35,7 +35,9 @@ and packet-per-second throughput using iperf3 on single threaded connections.
 The bandwidth test shows the speed of the data transfer. On the other hand,
 the jitter test measures the variation in the delay of received packets.
 The packet-per-second tests show the maximum number of (smallest sized) packets
-we can get through the transports.
+we can get through the transports.  
+Command line options are used to choose which tests to run, for how long, and
```
klynnrif: Command-line options choose what tests to run, for how long, and...

grahamwhaley:
@klynnrif updated, thanks.

grahamwhaley:
hmm, looks like something went wonky with my rebase - let me fix and push that again..

grahamwhaley:
fixed

Issue #910, opened 2018-02-14
ci: metrics: restore 'no KSM' on proxy for now


grahamwhaley:
In https://github.com/clearcontainers/tests/pull/828 we removed the `--ksm initial` ksm disabling line that was used by the metric CI, on the grounds that we moved to the 'one proxy per VM' model - but - we still have the single global proxy installed to control KSM, as we do not have the 'ksm reaper' set up and running yet (https://github.com/clearcontainers/proxy/pull/177).

Until https://github.com/clearcontainers/proxy/pull/177 lands, we need to re-install the KSM disabling code.

grahamwhaley:
See #913

PR #911, opened 2018-02-14
ci: metrics: disable ksm on global proxy if under metrics

grahamwhaley:
If running under the metrics CI, then we want to disable the global
KSM control of the proxy, and control it ourselves.

This is effectively an unwind of #828, that was a touch premature.

Fixes: #910

### DIFF
--- a/.ci/install_proxy.sh
+++ b/.ci/install_proxy.sh
@@ -22,3 +22,18 @@ cidir=$(dirname "$0")
 source "${cidir}/lib.sh"
 
 clone_build_and_install "github.com/clearcontainers/proxy"
+
+systemd_extra_flags="-log debug"
+
+# If we are running under the metrics CI system then we do not want the proxy
+# to be dynmaically changing the KSM settings under us - we need control of them
+# ourselves
+if [[ $METRICS_CI ]]; then
+	systemd_extra_flags="${systemd_extra_flags} -ksm initial"
+fi
+
+proxy_systemd_file=$(sudo systemctl show cc-proxy | fgrep FragmentPath | awk 'BEGIN{FS="="}{print $2}')
+sudo sed -i "s/\(^ExecStart.*$\)/\1 ${systemd_extra_flags}/" "${proxy_systemd_file}"
+sudo systemctl daemon-reload
+echo "Start proxy service"
+sudo systemctl restart cc-proxy


grahamwhaley:
Please scrutinize - I've not gotten to run test this locally (but maybe we'll see what the metrics CI makes of it...)

jcvenegas:
lgtm lets compare new metrics to know if got expected values.

grahamwhaley:
The CI is looking good to me - I now only see one entry for `memory-footprint-ksm.csv`, and the file `memory-footprint.csv` (the non-KSM version) has turned up.
Once this is merged I'll tweak the CI checkmetrics toml config files to start tracking `memory-footprint.csv` again.

grahamwhaley:
Marking as DNM whilst I open an Issue to discuss the 'real' issue...

.ci/install_proxy.sh
```diff
@@ -22,3 +22,18 @@ cidir=$(dirname "$0")
...
+systemd_extra_flags="-log debug"
+
+# If we are running under the metrics CI system then we do not want the proxy
+# to be dynmaically changing the KSM settings under us - we need control of them
+# ourselves
+if [[ $METRICS_CI ]]; then
+	systemd_extra_flags="${systemd_extra_flags} -ksm initial"
+fi
+
+proxy_systemd_file=$(sudo systemctl show cc-proxy | fgrep FragmentPath | awk 'BEGIN{FS="="}{print $2}')
```
devimc: wait, the proxy is no more a systemd service.
also awk is not needed, `sudo systemctl -p FragmentPath show cc-proxy | cut -d= -f2` work

grahamwhaley: but, that is the problem - the proxy *still is a systemd service* under the CIs:
https://github.com/clearcontainers/tests/blob/master/.ci/install_proxy.sh
https://github.com/clearcontainers/proxy/blob/master/Makefile#L88

we have a mis-match between the packaging distro installs and the CI scripted installs - we opened a few new Issues on Thursday about this. Need input from @chavafg I think. Let me see if I can find those other relevant issues...

grahamwhaley: Here we go - have a look at: https://github.com/clearcontainers/tests/issues/913

chavafg: We are not installing cc-proxy service files any more. This was removed some time ago: https://github.com/clearcontainers/tests/commit/dd24a5b544b8f1cdab91f910792d994daeedae8a

chavafg: Sorry, I see that the `make install` rule of the `proxy` still installs the service files, but I think we do not make use of them.  https://github.com/clearcontainers/proxy/pull/177 should solve this, right?

grahamwhaley: but clearcontainers/proxy#177 is not merged ;-)
So, today we do install and run the proxy on the CI systems afaict - and that messes up my metrics CI memory measurements ;-)
Either we
- turn it off again with this patch (quick fix)
- or we fix the .ci install scrips and proxy Makefile and replace it with ksm-throttler installation (the better fix, but will take a little more work to write the scripts)

jodh-intel: We're going to need to resolve #913 (create `.ci/install_ksm_throttler.sh`) first because if I get https://github.com/clearcontainers/proxy/pull/177 landed, the CI and PnP will immediately break fwics as there will be no KSM functionality in the proxy.

chavafg: ohh ok, then, to me it is ok to merge this and have a proper fix once clearcontainers/proxy#177 gets merged.

grahamwhaley: metrics CI will be very happy if there is no KSM via either proxy or ksm-throtter ;-) - metrics CI controls KSM itself so it knows exactly what it is measuring...
but, yes, QA CI should have ksm-throttler running so we are testing the 'same' environment you'd get if you installed from the distro packages.

jodh-intel:
:boom: :boom: :boom:

PR #917, opened 2018-02-16
test: cri-containerd test

jcvenegas:
Add initial support for cri-containerd tests.

### DIFF
--- a/.ci/setup_env_fedora.sh
+++ b/.ci/setup_env_fedora.sh
@@ -56,3 +56,6 @@ chronic sudo -E dnf -y install bison
 if ! command -v docker > /dev/null; then
 	"${cidir}/../cmd/container-manager/manage_ctr_mgr.sh" docker install
 fi
+
+echo "Install cri-containerd dependencies"
+chronic sudo -E dnf -y install libseccomp-devel btrfs-progs-devel libseccomp-static
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -18,8 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-
-tmp_dir=$(mktemp -d -t tmp.XXXXXXXXXX)
+tmp_dir=$(mktemp -d -t cri-containerd-install.XXXXXXXXXX)
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 finish() {
@@ -39,8 +38,7 @@ info() {
 trap finish EXIT
 
 source "${script_dir}/lib.sh"
-get_cc_versions
-
+#get_cc_versions
 #TODO (jcvenega) remove after https://github.com/clearcontainers/runtime/pull/1091 is merged
 cri_containerd_version="v1.0.0-rc.0"
 
@@ -52,7 +50,8 @@ git fetch
 info "Checkout to ${cri_containerd_version}"
 git checkout "${cri_containerd_version}"
 info "Installing cri-containerd"
-make install.tools
 make install.deps
 make
 sudo -E PATH=$PATH make install
+#TODO (carlos) : Fedora does not look for /usr/local with sudo.
+sudo cp /usr/local/bin/containerd-shim /usr/bin/containerd-shim
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+# 
+# Copyright (c) 2018 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+
+tmp_dir=$(mktemp -d -t tmp.XXXXXXXXXX)
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+finish() {
+  rm -rf "$tmp_dir"
+}
+
+die() {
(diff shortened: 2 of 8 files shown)


jcvenegas:
@grahamwhaley  @chavafg I got an error from metrics ci, I suspect is the golang version. 

```
 vendor/github.com/stevvooe/ttrpc/unixcreds_linux.go:21: uc.SyscallConn undefined (type *net.UnixConn has no field or method SyscallConn)
```
Where is installed golang in metrics ci ?

grahamwhaley:
@jcvenegas that is (or should be) v1.8.3 of go on the metrics CI build slaves.

grahamwhaley:
it is installed in the standard go path `/usr/local/go`. We set that up in the Jenkins config file/dialog:

```
export PATH=/usr/local/go/bin:${GOPATH}/bin:/sbin:${PATH}
```

And you can find in the logs somewhere around the auto-code generation section:

```
11:40:00 • golang:
11:40:00 	go version go1.8.3 linux/amd64
```

.ci/install_cri-containerd.sh
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+cidir=$(dirname "$0")
+source "${cidir}/lib.sh"
+get_cc_versions
+# Fixme add a version to use in versions.txt file
```
jodh-intel: Is there an issue for this, or are you planning to update this PR to remove the "Fixme"?

.ci/setup_env_fedora.sh
```diff
@@ -50,6 +50,9 @@ chronic sudo -E dnf -y install btrfs-progs-devel device-mapper-devel 	  \
 	libgpg-error-devel libseccomp-devel libselinux-devel ostree-devel \
 	pkgconfig go-md2man
 
+echo "Install cri-containerd dependencies"
+chronic sudo -E dnf -y install libseccomp-devel btrfs-progs-devel libseccomp-static
```
jodh-intel: OOI, why do we need the `-static` version of this library? It doesn't appear to be needed on Ubuntu too.

integration/cri-containerd/test.sh
```diff
@@ -0,0 +1,57 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -x
+set -eE
+
+err_report() {
+	cat /tmp/test-cri/cri-containerd.log
```
jodh-intel: Looking at `hack/test-cri.sh`, this is the default, but we can change that by setting `REPORT_DIR`. I think we should do that as this directory may already exist for example. Alternatively, I'd delete `/tmp/test-cri/` before running the test.

integration/cri-containerd/test.sh
```diff
@@ -0,0 +1,57 @@
...
+	cat /tmp/test-cri/containerd.log
+}
+
+trap 'err_report' ERR
+
+
+repo="github.com/containerd/cri-containerd"
+# FIXME: cri-containerd fails using macvtap.
+sudo sed -r -i -e 's/^(internetworking_model=).*/\1"bridged"/g' /usr/share/defaults/clear-containers/configuration.toml
```
jodh-intel: Might be worth creating a variable for this path (atleast the `clear-containers/` element) to make it easier to migrate it to Kata?

integration/cri-containerd/test.sh
```diff
@@ -0,0 +1,57 @@
...
+
+CRITEST=${GOPATH}/bin/critest
+# make sure cri-containerd test install the proper installation for testing
+rm -f "${CRITEST}"
+pushd ${GOPATH}/src/${repo}
+
+FOCUS="runtime should support basic operations on container" CRI_CONTAINERD_FLAGS="--containerd-runtime-engine cc-runtime --stream-port 10011" make test-cri
```
jodh-intel: Same comment as above - might be worth creating a variable for `cc-runtime`.

integration/cri-containerd/test.sh
```diff
@@ -0,0 +1,57 @@
...
+                TestContainerListStatsWithIdSandboxIdFilter
+                TestDuplicateName
+                TestImageLoad
+                TestImageFSInfo
+                TestSandboxCleanRemove
+             )
+
+for t in "${passing_test[@]}"
+do
+CRI_CONTAINERD_FLAGS="--containerd-runtime-engine cc-runtime --stream-port 10011" FOCUS="${t}" make test-integration
```
jodh-intel: A runtime variable could be re-used here too.

integration/cri-containerd/test.sh
```diff
@@ -0,0 +1,57 @@
...
+sudo sed -r -i -e 's/^(internetworking_model=).*/\1"bridged"/g' /usr/share/defaults/clear-containers/configuration.toml
+
+CRITEST=${GOPATH}/bin/critest
+# make sure cri-containerd test install the proper installation for testing
+rm -f "${CRITEST}"
+pushd ${GOPATH}/src/${repo}
+
+FOCUS="runtime should support basic operations on container" CRI_CONTAINERD_FLAGS="--containerd-runtime-engine cc-runtime --stream-port 10011" make test-cri
+
+passing_test=(
```
jodh-intel: This is great. However, how about running all the tests and asserting that this list pass, and all the other tests fail?

Assuming we update `versions.txt` to pin the version of `cri-containerd` we're using, this should work as every time we update that version, and re-run the tests locally we *might* end up getting more tests passing, which can then be added to this list.

go1.9 provides a `go test -list '.*'` command which is supposed to list all tests in a package, but I cannot get it to work locally. As an alternative, this seems to be a more reliable method of listing all test files and test functions:

```
find . -type f|grep -v vendor/|grep "_test.go"|xargs grep "^func Test[^ ]"|sed 's/func //g'|cut -d\( -f1|cut -d: -f2|sort
```

integration/cri-containerd/test.sh
```diff
@@ -0,0 +1,57 @@
...
+err_report() {
+	cat /tmp/test-cri/cri-containerd.log
+	cat /tmp/test-cri/containerd.log
+}
+
+trap 'err_report' ERR
+
+
+repo="github.com/containerd/cri-containerd"
+# FIXME: cri-containerd fails using macvtap.
```
devimc: did you create an issue for this?

PR #919, opened 2018-02-16
integration: crio: Update bats to match latest version from CRI-O

### LINKED ISSUE #918: integration: crio: Rely on latest bats from CRI-O
We should rely on the latest version of CRI-O bats to update our `crio.bats` list of bats.

sboeuf:
This commit entirely relies on the latest version of CRIO in order
to update the bats properly. This is a simple copy/paste of the tests
we already support, but it updates the content to the latest version.

Fixes #918

Signed-off-by: Sebastien Boeuf <sebastien.boeuf@intel.com>

### DIFF
--- a/integration/cri-o/crio.bats
+++ b/integration/cri-o/crio.bats
@@ -16,24 +16,21 @@
 
 load helpers
 
-function setup() {
-	start_crio
-}
-
 function teardown() {
-	stop_crio
-	sudo umount "$TESTDIR/crio/overlay"
 	cleanup_test
 }
 
 @test "ctr not found correct error message" {
-	run crictl inspect randomid
+	start_crio
+	run crictl inspect "container_not_exist"
 	echo "$output"
 	[ "$status" -eq 1 ]
-	[[ "$output" =~ "container with ID starting with randomid not found" ]]
+
+	stop_crio
 }
 
 @test "ctr termination reason Completed" {
+	start_crio
 	run crictl runs "$TESTDATA"/sandbox_config.json
 	echo "$output"
 	[ "$status" -eq 0 ]
@@ -45,7 +42,7 @@ function teardown() {
 	run crictl start "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
-	sleep 5
+	run sleep 5
 	run crictl inspect --output yaml "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
@@ -53,9 +50,11 @@ function teardown() {
 
 	cleanup_ctrs
 	cleanup_pods
+	stop_crio
 }
 
 @test "ctr termination reason Error" {
+	start_crio
 	run crictl runs "$TESTDATA"/sandbox_config.json
 	echo "$output"
 	[ "$status" -eq 0 ]
@@ -69,6 +68,7 @@ function teardown() {
 	run crictl start "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
+	run sleep 5
 	run crictl inspect --output yaml "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
@@ -76,9 +76,11 @@ function teardown() {
 
 	cleanup_ctrs
 	cleanup_pods
+	stop_crio
 }
 
 @test "ctr remove" {
+	start_crio
 	run crictl runs "$TESTDATA"/sandbox_config.json
 	echo "$output"
 	[ "$status" -eq 0 ]
@@ -101,9 +103,11 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	cleanup_ctrs
 	cleanup_pods
+	stop_crio
 }
 
 @test "ctr lifecycle" {
+	start_crio
 	run crictl runs "$TESTDATA"/sandbox_config.json
 	echo "$output"
 	[ "$status" -eq 0 ]
@@ -111,6 +115,7 @@ function teardown() {
 	run crictl sandboxes --quiet
 	echo "$output"
 	[ "$status" -eq 0 ]
+	[[ "$output" == "$pod_id" ]]
 	run crictl create "$pod_id" "$TESTDATA"/container_redis.json "$TESTDATA"/sandbox_config.json
 	echo "$output"
 	[ "$status" -eq 0 ]
@@ -118,6 +123,7 @@ function teardown() {
 	run crictl ps --quiet
 	echo "$output"
 	[ "$status" -eq 0 ]
+	[[ "$output" == "$ctr_id" ]]
 	run crictl inspect "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
@@ -130,6 +136,7 @@ function teardown() {
 	run crictl ps --quiet
 	echo "$output"
 	[ "$status" -eq 0 ]
+	[[ "$output" == "$ctr_id" ]]
 	run crictl stop "$ctr_id"
 	echo "$output"
 	[ "$status" -eq 0 ]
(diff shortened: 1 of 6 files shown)


chavafg:
@sboeuf K8s is failing because crio service is down

```
[init] Using Kubernetes version: v1.9.3
[init] Using Authorization modes: [Node RBAC]
[preflight] Running pre-flight checks.
[preflight] Some fatal errors occurred:
        [ERROR CRI]: unable to check if the container runtime at "/var/run/crio/crio.sock" is running: exit status 1
[preflight] If you know what you are doing, you can make a check non-fatal with `--ignore-preflight-errors=...`
+ RC=2
```

can you also start the `crio` service in the `integration/kubernetes/init.sh` script?

integration/cri-o/crio.bats
```diff
@@ -16,24 +16,21 @@
...
-
 function teardown() {
-	stop_crio
-	sudo umount "$TESTDIR/crio/overlay"
 	cleanup_test
 }
 
 @test "ctr not found correct error message" {
-	run crictl inspect randomid
+	start_crio
```
jodh-intel: Why do all the tests now call what was happening in the per-test `setup()` and `teardown()` functions? This seems like a retrograde step and isn't explained in the commit.

integration/openshift/teardown.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 Intel Corporation
```
devimc: 2018

integration/cri-o/crio.bats
sboeuf: @jodh-intel those changes are following the bats from version 1.9 of CRIO. And those tests start and stop crio for every test. That's why this PR is also making sure that we start the crio service only when needed. Otherwise we end up with some failures when bats are trying to stop the service. Indeed, they kill the crio process, but it gets restarted automatically.

integration/openshift/teardown.sh
sboeuf: Will do

integration/cri-o/crio.bats
jodh-intel: I'm still confused. Previously, `crio.bats` had a `setup()` and `teardown()` and those started and stopped the crio service respectively.

But now, you've explicitly added a start and stop to every test - why can't we continue to use `setup()` and `teardown()` which do the same, but in a simpler way?

Oh... I think I've just worked out what you're doing now. "a simple copy/paste" has only just registered ;)

The `crio.bats` you're adding is a copy of https://github.com/kubernetes-incubator/cri-o/blob/master/test/ctr.bats right?

If so, why? Why can't we just `go get -d` / `git clone` (or even `curl`) the version we want to test with?

jodh-intel: Another observation: the `crio.bats` on this PR is very similar to https://github.com/kubernetes-incubator/cri-o/blob/master/test/ctr.bats, but we've added our own copyright header that isn't in the original...?

sboeuf: Yes that's exactly what you said, I'm doing a copy/paste here from the CRI-O repo. The reason I don't do a simple `go get ...` to retrieve the bats file, is that we don't run all the bats, because they don't pass all. That's why we have to pick only the one passing for us, and that's also the reason why we have our own copyright on this one.

sboeuf: I am fine with putting CRI-O copyright here, if that's what needs to be done. @grahamwhaley WDYT ?

sboeuf:
@chavafg please review it, the PR is ready now.

integration/cri-o/crio.bats
grahamwhaley: Oh what fun - the original does not have an explicit copyright in it, and I don't see an obvious `COPYRIGHT` file in the project/repo. I think we should attribute the file (say at the top basically 'this is a modified copy of the CRI-O file from https://github.com/kubernetes-incubator/cri-o/blob/master/test/ctr.bats). Don't make it an actual copyright notice (it's not really for us to say that - thats up to the CRI-O folks) - but make it clear where we cloned it from eh?

jodh-intel: @sboeuf - I'm still unclear why we need to have this stored in our repository? I understand we can't use `go get`, but surely we could have the test scripts do something like the following?:

```
commit="6de90e046a0009b69dcfbc53335d8742532004d8"
curl -O "https://raw.githubusercontent.com/kubernetes-incubator/cri-o/${commit}/test/ctr.bats"
```

That way, we don't need to store the file and we don't need to worry about copyrights surely?

grahamwhaley: I think it is because as @sboeuf said, we modify the script to remove some of the tests that we don't run? so, unless we could automate that with some sed/awk magic or similar, maybe having a locally modified copy is the easiest route at present?

integration/cri-o/init.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+echo "Stop crio service"
```
chavafg: This might be confusing for some people... Can you add a comment describing why we need to stop the `crio` service when running the tests?

integration/cri-o/crio.bats
sboeuf: Well we thought it would be easier to get our own test file, that we copy when running the CI. I am not modifying any behavior here, just following what we've been doing for months now. You can try to move to something different as you described here, but this needs to be done in a different PR, I just want to make sure here that we run the tests based on the latest scripts from CRI-O.

integration/cri-o/init.sh
sboeuf: Yes you're right, let me add this!

integration/cri-o/crio.bats
jodh-intel: @sboeuf - yes, I see that.

@chavafg - we might want to find a better way of dealing with this. If copying is the best process, we need to document it somewhere in a `README.md` imho (on a separate PR).

integration/cri-o/init.sh
sboeuf: PR updated !

integration/cri-o/crio.bats
chavafg: sure, I opened https://github.com/clearcontainers/tests/issues/921 to address this concern.

sboeuf:
@grahamwhaley could you check the CI metrics failure please ? I am not sure if this is significant, or if we can merge this PR anyway.

grahamwhaley:
Looking in the log, towards the bottom, we have:
```
09:27:15 version to install cc-20640-agent-4186b0
09:27:15 download https://github.com/clearcontainers/osbuilder/releases/download/cc-20640-agent-4186b0/image-cc-20640-agent-4186b0-binaries.tar.gz
...
 10 15.0M   10 1665k    0     0    990      0  4:26:06  0:28:41  3:57:25     0Build timed out (after 30 minutes). Marking the build as aborted.
09:55:57 Build was aborted
```

Nothing metrics specific - looks like we failed to download our own agent binary. @jcvenegas @chavafg 
I'll nudge a rebuild for you @sboeuf

sboeuf:
@grahamwhaley do you think we can merge this anyway ?

grahamwhaley:
Not an issue from metrics. You need more approvers though.
So, we are happy with the 'copy and maintain' approach, yes?

sboeuf:
@grahamwhaley yes I need @chavafg approval on this.
About copy and maintain, we might move to a different approach but this will be handle from a different PR.

chavafg:
lgtm

@sboeuf @grahamwhaley opened #921 to address the approach concern.

Issue #932, opened 2018-02-27
CI: remove sudo from image build step in install_asset.sh 


chavafg:
Now that https://github.com/clearcontainers/packaging/pull/277 is merged, we can remove sudo from https://github.com/clearcontainers/tests/blob/master/.ci/install_asset.sh#L123. 

By having sudo in that line, we end with $GOPATH/bin been owned by root, which is something undesired.

PR #933, opened 2018-02-27
CI: call release_image_github.sh script without sudo

chavafg:
We don't want to call release_image_github.sh using sudo,
as we end up having $GOPATH/bin been owned by root.
As https://github.com/clearcontainers/packaging/pull/277
is now merged, we can remove sudo.

Fixes: #932.


devimc:
LGTM

Issue #927, opened 2018-02-26
stability: rm soak: does not allow override of ITERATIONS


grahamwhaley:
The rm soak test allows override of many of its configs, but looks like I forgot to do that for the number of ITERATIONS:
https://github.com/clearcontainers/tests/blob/master/integration/stability/soak_parallel_rm.sh#L25

change to:
```
ITERATIONS="${ITERATIONS:-5}"
```

PR #935, opened 2018-02-28
stability: soak: allow iterations to be over-ridden

grahamwhaley:
Allow the iterations count to be over-ridden from the
environment.
Also, print out the iterations as 'n of x' as we run.

Fixes: #927


jodh-intel:
lgtm

Issue #946, opened 2018-03-13
run_kubernetes_tests.sh script does not fail


chavafg:
Add `set -e` to `run_kubernetes_tests.sh` to fail if some of the commands inside are not successful.

PR #947, opened 2018-03-13
CI: add set -e to run_kubernetes_tests.sh

chavafg:
Add `set -e` to fail if a command does not return
successful.

Fixes: #946.


jodh-intel:
lgtm

GabyCT:
lgtm

Issue #950, opened 2018-03-15
CI: do not hold kubernetes packages.


chavafg:
Now that the kubernetes installation is done using the CI scripts, kubernetes packages are being installed in the metrics CI server, and when we try to update to a new kubernetes version. The metrics CI fails because it cannot update them succsessfully:

```
13:07:23 The following held packages will be changed:
13:07:23   kubeadm kubectl kubelet
13:07:23 The following packages will be upgraded:
13:07:23   kubeadm kubectl kubelet
13:07:23 3 upgraded, 0 newly installed, 0 to remove and 78 not upgraded.
13:07:23 E: Held packages were changed and -y was used without --allow-change-held-packages.
13:07:23 Build step 'Execute shell' marked build as failure
```

PR #951, opened 2018-03-15
CI: do not hold kubernetes packages

chavafg:
There is no need to hold the kubernetes packages to
a specific version in the CI. Holding them was preventing
the metrics CI node to fail installing the correct packages.

Fixes: #950

### DIFF
--- a/.ci/install_kubernetes.sh
+++ b/.ci/install_kubernetes.sh
@@ -34,7 +34,6 @@ EOF"
 curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
 sudo -E apt update
 sudo -E apt install -y kubelet="$kubernetes_version" kubeadm="$kubernetes_version" kubectl="$kubernetes_version"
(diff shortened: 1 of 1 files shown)


jodh-intel:
This looks fine. But I wonder if we need to special-case the metrics CI so that only it removes the hold) rather than removing the hold entirely?

/cc @grahamwhaley.

grahamwhaley:
Right now the metrics CI doesn't do any k8s testing (yet...), so the change should not affect the present restults.
What I don't quite understand is why the metrics CI was failing to install packages, but presumably the QA CI was not - I'm guessing different versions of Ubuntu maybe? Do we have a few more details on this?

devimc:
lgtm

chavafg:
@grahamwhaley It started to fail when we moved k8s from v1.9.2 to v1.9.3. As `apt` was holding the version of the packages, they couldn't be upgraded and the CI failed on this step. And as on QA CI, every run is a new VM, the problem didnt appear.

Issue #952, opened 2018-03-16
ci: metrics: no longer need to restart cc-proxy


grahamwhaley:
During initialisation the ci metrics library restarts the systemd cc-proxy service:
https://github.com/clearcontainers/tests/blob/master/metrics/lib/common.bash#L111

We no longer have a global cc-proxy service to restart (in theory) -  so remove this.

Note: this line is very likely what has been inteferring with #923 and probably affected #940 verification as well.

/cc @chavafg

PR #953, opened 2018-03-16
metrics: do not restart proxy any more

grahamwhaley:
We no longer have a global cc proxy systemd service, so do not
try to restart it during metrics system initialisation.

Fixes: #952


jodh-intel:
lgtm

devimc:
lgtm

PR #955, opened 2018-03-19
metrics: add support for kata-runtime

grahamwhaley:
Whilst we transition metrics over to the kata repos, it will be very helpful to be able to run the metrics here on a kata-runtime.
Add kata-runtime recognition support to the metrics scripts.


jodh-intel:
lgtm

grahamwhaley:
Ooh, I see one commit is poorly formatted at least, let me correct that.

PR #957, opened 2018-03-20
metrics: Add FIO flags to produce stable results.

### LINKED ISSUE #956: metrics: Add FIO flags to produce stable results
Add FIO flags like time_based, drop host pages, invalidate, direct, etc to produce more stable results.

GabyCT:
Add FIO flags like time_based, drop host pages,
invalidate, direct, etc to produce more stable results.

Fixes #956

### DIFF
--- a/metrics/storage/fio_job.sh
+++ b/metrics/storage/fio_job.sh
@@ -37,7 +37,11 @@ NUM_JOBS="1"
 OPERATION="randread"
 TEST_NAME="storage fio test"
 EXTRA_ARGS=""
-TEST_TIME="30"
+TEST_TIME="300"
+RAMP_TIME="60"
+DIRECT_VALUE="1"
+INVALIDATE_VALUE="1"
+STATUS_TIME="30"
 
 # This docker image includes FIO tool.
 FIO_IMAGE="clusterhq/fio-tool"
@@ -69,6 +73,7 @@ Usage: $0 [-h] [options]
         -n,    Max number of jobs.
         -r,    Runtime for docker.
         -t,    Test name.
+        -u,    Ramp time.
         -x,    Extra fio arguments.
 EOF
 )
@@ -91,6 +96,11 @@ function create_fio_prep_job()
 		--filename="$FILE_NAME" \
 		--bs="$BLOCK_SIZE" \
 		--size="$FILE_SIZE" \
+		--time_based \
+		--direct="$DIRECT_VALUE" \
+		--invalidate="$INVALIDATE_VALUE" \
+		--status-interval="$STATUS_TIME" \
+		--ramp_time="$RAMP_TIME" \
 		--readwrite="$OPERATION" --max-jobs=$NUM_JOBS \
 		--runtime=1 \
 		${EXTRA_ARGS}
@@ -112,6 +122,11 @@ function create_fio_run_job()
 		--filename="$FILE_NAME" \
 		--bs="$BLOCK_SIZE" \
 		--size="$FILE_SIZE" \
+		--time_based \
+		--direct="$DIRECT_VALUE" \
+		--invalidate="$INVALIDATE_VALUE" \
+		--status-interval="$STATUS_TIME" \
+		--ramp_time="$RAMP_TIME" \
 		--readwrite="$OPERATION" --max-jobs=$NUM_JOBS \
 		--runtime="$TEST_TIME" \
 		${EXTRA_ARGS}
@@ -131,9 +146,9 @@ function parse_fio_results()
 		die "bandwidth result not found: check FIO job configuration"
 	fi
 
-	units=$(echo "$bw_data" grep "bw (" | grep -o "[A-Z]*")
-	avg=$(echo "$bw_data" | awk -F "," '{print $4}' | cut -d "=" -f2)
-	stdev=$(echo "$bw_data" | awk -F "," '{print $5}' | cut -d "=" -f2)
+	units=$(echo "$bw_data" | sed 's/^.*bw (/bw (/' | cut -f1 -d ':' | cut -f2 -d '(' | cut -f1 -d ')' | tail -1)
+	avg=$(echo "$bw_data" | sed 's/^.*bw (/bw (/' | awk -F "," '{print $4}' | cut -d "=" -f2 | tail -1)
+	stdev=$(echo "$bw_data" | sed 's/^.*bw (/bw (/' | awk -F "," '{print $5}' | cut -d "=" -f2 | tail -1)
 
 	echo "Units: $units"
 	case "$units" in
@@ -153,7 +168,7 @@ function main()
 {
 	cmds=("bc" "awk" "smem")
 	local OPTIND
-	while getopts "b:e:hs:o:n:r:t:T:x:" opt;do
+	while getopts "b:e:hs:o:n:r:t:T:u:x:" opt;do
 		case ${opt} in
 		b)
 		    BLOCK_SIZE="${OPTARG}"
@@ -182,6 +197,9 @@ function main()
 		    ;;
 		T)
 		    TEST_TIME="${OPTARG}"
+		    ;;
+		u)
+		    RAMP_TIME="${OPTARG}"
 		    ;;
 		x)
 		    EXTRA_ARGS="${OPTARG}"
@@ -230,6 +248,9 @@ function main()
 	(sleep 3; ps --no-headers -o %cpu -C $(basename ${PROCESS}) > ${cpu_temp})&
 	(sleep 4; sudo smem -H -P "^${PROCESS}" -c "pss" > ${pss_temp})&
 	(sleep 5; sudo smem -H -P "^${PROCESS}" -c "rss" > ${rss_temp})&
+
+	# Drop the host page cache
+	sudo bash -c "echo 3>/proc/sys/vm/drop_caches"
 
 	# Finally Launch container
 	output=$(docker exec ${CONTAINER_ID} bash -c "$FIO_RUN_JOB")
--- a/.ci/run_metrics_ci.sh
+++ b/.ci/run_metrics_ci.sh
@@ -131,10 +131,14 @@ pushd "$CURRENTDIR/../metrics"
 	#
 	# Run some IO tests
 	#
-	bash storage/fio_job.sh -b 16k -o randread -t "storage IO random read bs 16k"
-	bash storage/fio_job.sh -b 16k -o randwrite -t "storage IO random write bs 16k"
-	bash storage/fio_job.sh -b 16k -o read -t "storage IO linear read bs 16k"
-	bash storage/fio_job.sh -b 16k -o write -t "storage IO linear write bs 16k"
+	# Block Size and Ramp Time Settings for Fio Test
+	BLOCK_SIZE="16k"
+	RAMP_TIME="60"
+
+	bash storage/fio_job.sh -b ${BLOCK_SIZE} -u ${RAMP_TIME} -o randread -t "storage IO random read bs ${BLOCK_SIZE}"
+	bash storage/fio_job.sh -b ${BLOCK_SIZE} -u ${RAMP_TIME} -o randwrite -t "storage IO random write bs ${BLOCK_SIZE}"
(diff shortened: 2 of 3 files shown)


GabyCT:
@grahamwhaley and @amshinde, after running FIO tests , I think that if we enable these flags we can have more stable results

metrics/storage/fio_job.sh
```diff
@@ -29,15 +29,16 @@ TEST_ARGS="$@"
...
 ENGINE="sync"
 FILE_SIZE="4G"
 FIO_TEST_NAME="io_test"
 NUM_JOBS="1"
 OPERATION="randread"
 TEST_NAME="storage fio test"
 EXTRA_ARGS=""
-TEST_TIME="30"
+TEST_TIME="300"
+IO_DEPTH="1024"
```
amshinde: While it is nice to have this option, `iodepth` will be always 1 in case of `sync` engine. `iodepth` (number of concurrent requests) only makes a difference in case of `libaio` engine.

amshinde: Running test for a longer time does produce reproduceable results. But they do take some time now (around 7*4 min).
We may have to live with this initially, but we really should consider preparing an image with requirements installed and using that instead going forward to cut down on the test time.

@GabyCT Can you try doing a few runs with with FILE_SIZE=4G and taking out TEST_TIME and --timebased all together. This may perform faster on the xeon that we use for testing while still making sure we transfer 4G of data which is greater than the 2G of memory that we start the VM with.

grahamwhaley:
passing note (before I get to scrutinize)... can you add some notes here about:
 - how long a run takes on your machine (just for reference)
 - what is the variability between runs (say you do 5 runs) - and not just SD ( or better RSD (CoV)), but also the % range of the min-max as well please (as currently we check metrics as a 'one shot' measure and check it is within a min/max bound - so the actual min/max answer (the mean?) matters)

Thanks.

GabyCT:
@grahamwhaley so running the tests like the PR takes around ~50 minutes to complete all of them. Now if I decrease the runtime to 30 s and run the tests for example 5 times and then check SD, max, min, here are the results
Operation - storage IO linear read bs 16k  

30 s / 5 times                              
mean 46404.258                          
max 51273.7
min 43260.700
sd 3369.586 
cov 7.261 

vs 300 s / 1 time
mean 46679.775
max 48612.650
min 44746.900
sd 1932.875
cov 4.141

metrics/storage/fio_job.sh
```diff
@@ -29,15 +29,16 @@ TEST_ARGS="$@"
 # These values are a basic configuration for a FIO job
 # execution, however they could be overridden in order to
 # run a customized FIO job.
-BLOCK_SIZE="4k"
+BLOCK_SIZE="16k"
```
grahamwhaley: I think us setting sensible (meaningful, useful) defaults in this file is fine, but you also need to change the command line arguments that get passed in during the actual CI testing, which come from:
https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh#L134

metrics/storage/fio_job.sh
```diff
@@ -91,6 +92,10 @@ function create_fio_prep_job()
 		--filename="$FILE_NAME" \
 		--bs="$BLOCK_SIZE" \
 		--size="$FILE_SIZE" \
+		--time_based \
+		--direct=1 \
+		--invalidate=1 \
+		--ramp_time="$RAMP_TIME" \
```
grahamwhaley: Can you add a new command line option so we can optionally change RAMP_TIME please.

metrics/storage/fio_job.sh
```diff
@@ -206,6 +215,9 @@ function main()
 		die "Unknown runtime: $RUNTIME"
 	fi
 
+	# Drop the host page cache
+	sudo echo 3>/proc/sys/vm/drop_caches
```
grahamwhaley: I wonder if dropping the host page cache is better done after we have created the test image rather than before - when we create the image we will populate the host cache...
Can you try this below the image creation call and see if you can see a difference?

grahamwhaley: Metrics jenkins is failing on:
```
storage/fio_job.sh: line 219: /proc/sys/vm/drop_caches: Permission denied
```

You may need to do something like:
```
sudo bash -c "echo 0 > ${KSM_ENABLE_FILE}"
```
taken from https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh#L63

GabyCT:
@grahamwhaley currently this is failing in the CI of metrics as it says
Build timed out (after 30 minutes). Marking the build as aborted.

metrics/storage/fio_job.sh
GabyCT: @grahamwhaley done

metrics/storage/fio_job.sh
GabyCT: @grahamwhaley done

metrics/storage/fio_job.sh
GabyCT: @grahamwhaley one question why in the run_metrics_ci we have
$ bash storage/fio_job.sh -b 16k ....
when the block size is 4k?

GabyCT:
@amshinde taking out time_based and test_time is causing some weird results.

grahamwhaley:
Ah, the 30 minute timeout - we have Jenkins configured to kill a build after 30 minutes if it is 'stuck' - stuck being defined by 'generated no output'.
Before we consider bumping that timeout up, could you see if you can get fio to give some sort of rolling status update - maybe every minute or every 10% of progress or similar?

metrics/storage/fio_job.sh
grahamwhaley: iirc, we chose it as a middle ground between 'small transfers' and 'large transfers'.
Ideally we'd do a spectrum of block sizes, but:
- that would multiply the run time by `n`
- we could in theory do a mix of block sizes:
    - but if we got worse on only small or big blocks, that would be harder to tell from a mixed run
    - I have a `blogbench` test in flight, which does a mix of sizes and reads/writes - so that may cover the more 'real world' case

Yes, we could do with more storage tests, but they take a long time and are produce noisy results - it is a WIP

GabyCT: @grahamwhaley , I am a bit confused, so before my changes if we ran the run_metrics_ci that runs `$bash storage/fio_job.sh -b 16k ..` what is running is a `block size =16 k` but if you ran the `fio_job.sh` it is running with `block size  4k`. My question is you want to have in the `run_metrics_ci` a different block size than `fio_job.sh`?

GabyCT: @grahamwhaley what will confused me is that when you run `fio_job.sh` and `run_metrics_ci` the results file has a title with a `block size =16k`

grahamwhaley: The '4k' in the `fio_job.sh` file is just a default - the expectation is that when it is invoked via a higher level script (like `run_metrics_ci`), then that script will explicitly set the block size and other parameters it wants. `fio_job.sh` should be generic, and then the invocation scripts set it up specifically.
So, we can set sensible defaults in `fio_job.sh`, in case anybody does run it directly or does not set up all the parameters, but what really matters is how we set up `run_metrics_ci.sh` to invoke it under the CI system.

The name - ah, I see the `run_metrics_ci.sh` [sets the name explicitly](https://github.com/clearcontainers/tests/blob/master/.ci/run_metrics_ci.sh#L134) using the `-t` option - that may be where the name confusion is coming from.

GabyCT:
@grahamwhaley , I enabled the `--status-interval` flag in order to print the status more often of the `fio` test, however, even with this the CI is still failing `Build timed out (after 30 minutes). Marking the build as aborted.`. The main issue is that before running the `fio` test the CI takes ~26 minutes to do the setup of the runtime and to run the previous metrics tests.

GabyCT:
@grahamwhaley and @amshinde this is the final version of the test. Could you please check this? thanks

grahamwhaley:
Ah, I see the Jenkins timeout was configured as an absolute hard-limit of 30 minutes (for the whole run), indeed. I've re-configured the Jenkins timeout to be 'timeout after 300 seconds of no activity' - let's see how that flies.

metrics/storage/fio_job.sh
```diff
@@ -131,9 +146,9 @@ function parse_fio_results()
 		die "bandwidth result not found: check FIO job configuration"
 	fi
 
-	units=$(echo "$bw_data" grep "bw (" | grep -o "[A-Z]*")
-	avg=$(echo "$bw_data" | awk -F "," '{print $4}' | cut -d "=" -f2)
-	stdev=$(echo "$bw_data" | awk -F "," '{print $5}' | cut -d "=" -f2)
+	units=$(echo "$bw_data" | sed 's/^.*bw (/bw (/' | cut -f1 -d ':' | cut -f2 -d '(' | cut -f1 -d ')' | tail -1)
```
grahamwhaley: Just out of interest, why the change in the parsing? Ah, maybe this is because we added the periodic status update output? Curious, as it adds a bunch of extra command calls in, that's all.
Hmm, I wonder where the periodic status output comes out - if it is stdout and we are capturing it here, then maybe that would/will not fix the 'jenkins timeout due to no activity' (as we will consume it, and Jenkins will not see it...) - let's keep an eye on that...

metrics/storage/fio_job.sh
```diff
@@ -238,6 +256,9 @@ function main()
 	docker stop ${CONTAINER_ID}
 	docker rm ${CONTAINER_ID}
 
+	# Drop the host page cache
+	sudo bash -c "echo 3>/proc/sys/vm/drop_caches"
```
grahamwhaley: I thought we'd be dropping the host caches between the `FIO_PREP_JOB` and the `FIO_RUN_JOB`, not after we have completed the test? Give me your thoughts here?

metrics/storage/fio_job.sh
GabyCT: @grahamwhaley yes as we put the periodic status update output and it the test is taking the stdout of every single result, I needed to change the parsing because we were having multiple results displayed and saved in the .csv

GabyCT: @grahamwhaley even if the result is displayed in the stdout the jenkins metrics is still failing...should I take another approach?

grahamwhaley: @GabyCT I suspect the problem right now is that the output is *not* displayed on stdout when running - if you check the code where you added the status output, it ends up in `FIO_RUN_JOB`, which is invoked such as:

```
output=$(docker exec ${CONTAINER_ID} bash -c "$FIO_RUN_JOB")
```

So, the output gets captured, and not output. If you look in the logs you can see:
```
07:44:58 
07:44:58 Run status group 0 (all jobs):
07:44:58    READ: io=218976KB, aggrb=218757KB/s, minb=218757KB/s, maxb=218757KB/s, mint=1001msec, maxt=1001msec
07:44:58 
07:49:58 Build timed out (after 5 minutes). Marking the build as aborted.
07:49:58 Build was aborted
```

You can see the 5 minute gap there with no output before the timeout. The reason we do see output from the 'prep' job above there in the log is because that is executed as:
```
docker exec ${CONTAINER_ID} bash -c "$FIO_PREP_JOB"
```
aka - output not captured in the script.

Unless we can dream up some fancy way to both capture and output the log (hmm, maybe something with `tee` and stderr???), then I don't really see an alternative but for me to bump the 'no output timeout'. Remind me - is the run time per-test currently 300+60s (to take into account ramp time as well)? So, if I bump to 10minutes, we should be OK??

grahamwhaley:
@GabyCT I've bumpled the jenkins 'no activity' timeout to 600s, on the grounds that the test runs for 300s (the cycle that produces no visible output anyhow). I've nudged a re-run on Jenkins. I think we'll still fail due to #960 , but you should be able to check that the fio did not time out...

grahamwhaley:
Yes, that is the #960 failure - we need to wait for https://github.com/kata-containers/runtime/pull/191 to land... but, given the FIO tests now pass, and we know what that failure is, we don't have to block on that...

Issue #958, opened 2018-03-20
CRI-O: set manage_network_ns_lifecycle to true


chavafg:
We need to set `manage_network_ns_lifecycle = true ` for CRI-O v1.9.8 and above.

PR #959, opened 2018-03-20
crio: Set manage_network_ns_lifecycle flag to true

chavafg:
From CRI-O v1.9.8 and above, the manage_network_ns_lifecycle
flag is optional and it is set to 'false'. But for Clear
Containers, we need to setup this flag to true.

Fixes: #958.


sboeuf:
LGTM

Issue #961, opened 2018-04-03
intstall crio may fail run twice.


jcvenegas:
- The script does not fetch before check out.
- It fails if go-md2man symlink already exist.

PR #962, opened 2018-04-03
ci: prevent cri-o installation script fail.

jcvenegas:
Fix:
- The script does not fetch before checkout.
- It fails if go-md2man symlink already exist.

Fixes: #961


jodh-intel:
Ideally better as 2 commits but it's a small and easy to understand change, so...

lgtm

grahamwhaley:
Ignore metrics CI - network issue caused one storage failure, and there seems to be something (newly) flaky with the iperf3 test maybe...

Issue #964, opened 2018-04-03
install crio script should check for proxy variables


jcvenegas:
The installation script could check for  proxy variables be added  to crio service.

Issue #960, opened 2018-03-29
ci: metrics: metrics ci fails for networking tests


jcvenegas:
The logs show the following
```
12:39:53 ===== starting test [iperf3 tests] =====
12:39:53 command: docker: yes
12:39:53 docker pull'ing: gabyct/network
12:39:53 Using default tag: latest
12:39:55 latest: Pulling from gabyct/network
12:39:55 0c62fc2b46a9: Pulling fs layer
12:40:04 0c62fc2b46a9: Verifying Checksum
12:40:04 0c62fc2b46a9: Download complete
12:40:18 0c62fc2b46a9: Pull complete
12:40:18 Digest: sha256:c7abad113ea5f3829c3fcdb7b4886d02e28e3dbc392da3b442bb36ce0dedfc76
12:40:18 Status: Downloaded newer image for gabyct/network:latest
12:40:18 docker pull'd: gabyct/network
12:40:18 Iteration 1
12:40:24 ERROR: iperf server init fails
12:40:24 ERROR: result argument not supplied
12:40:24 
12:40:24 ===== starting test [storage IO random read bs 16k] =====
```

\cc  @grahamwhaley @sboeuf

grahamwhaley:
Let's start adding a little context (it would be great if we knew what the first PR that showed this was for instance). For reference then, we see this over on: https://github.com/clearcontainers/runtime/pull/1091

It does not happen with a very simple hand run at my desk, so I'm thinking:
- I'll update all components locally and see if it happens
- If I can't make that happen I'll run it directly on one of the CI machines
- once we've identified the issue, we'll see if we can make the test more verbose in its failure to help track down any future similar issues.

grahamwhaley:
OK, I updated to the latest runtime/proxy/shim - and now it fails locally for me. Looks like we may have broken something. If I had to guess, then it is most likely around the 9p/tmpfs workaround: https://github.com/clearcontainers/tests/blob/master/metrics/network/network-metrics-iperf3.sh#L63-L68

grahamwhaley:
Update: Looks like 3.0.22 works, but 3.0.23 fails. It looks like it fails with some sort of tmpfile access on the `/dev/shm` mount, but a quick look at the mount in the working and failing case shows them to superficially look the same.
Current suspect could be around https://github.com/kata-containers/runtime/pull/123
/cc @sboeuf

amshinde:
@grahamwhaley What is the workaround you use for /dev/shm for iperf?
[...]

PR #965, opened 2018-04-03
ci: Add proxy env variables to crio service.

jcvenegas:
Check proxy env variables are set and add them to
the crio service.

- Generate the crio service.

Fixes: #964


chavafg (approved):
lgtm

jodh-intel:
lgtm

jodh-intel:
The metrics CI is failing due to:

```
open results/network-bwd-host-contr-receiver.csv: no such file or directory
```

/cc @grahamwhaley.

grahamwhaley:
Indeed @jodh-intel. If you look further back up the logs you will find:
```
12:09:59 ERROR: iperf server init fails
```
which directly relates to #960 where I diagnosed and explained the problem.
Until/unless somebody explains and justifies the change in the mount of `/dev/shm` to a 9p mount, then I don't see this as a metrics test failure - I see it as a break in our shm mount. Nobody has explained yet. Until we have settled what is the right thing to do with `shm` (and possibly other tmpfs/ramfs mounts), then I'm afriad this test, and thus the metrics, is going to remain broken...
 /cc @sboeuf @amshinde .

(Vaguely related is https://github.com/kata-containers/kata-containers/issues/21 I would say as well)

PR #967, opened 2018-04-05
Install go script

jcvenegas:
This script allow to install the golang version
defined in Clear Containers versions file.

It also takes as argument the version to be installed.

E.g.

- Install go 1.10
./install_go.sh 1.10

- Install go using versions file from project.
./install_go.sh -p

### DIFF
--- a/.ci/install_go.sh
+++ b/.ci/install_go.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+tmp_dir=$(mktemp -d -t install-go-tmp.XXXXXXXXXX)
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+script_name="$(basename "${BASH_SOURCE[0]}")"
+USE_VERSIONS_FILE=""
+PROJECT="Clear Containers"
+
+finish() {
+	rm -rf "$tmp_dir"
+}
+
+die() {
+	echo >&2 -e "\e[1mERROR\e[0m: $*"
+	exit 1
+}
+
+info() {
+	echo -e "\e[1mINFO\e[0m: $*"
+}
+
+usage(){
+	exit_code="$1"
+	cat <<EOT
+Usage:
+
+${script_name} [options] <args>
+
+Args:
+<go-version> : Install a specific go version.
+
+Example:
+${script_name} 1.10
+
+Options
+-p : Install go defined in ${PROJECT} versions file.
+-h : Show this help
+
+EOT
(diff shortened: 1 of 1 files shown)


.ci/setup.sh
```diff
@@ -38,6 +38,8 @@ if [ -n "$PULL_REQUEST_NUMBER" ] || [ -n "$LOCALCI_PR_NUMBER" ]; then
 	bash -f "${cidir}/run_fetch_branches_tool.sh"
 fi
 
+echo "Install go"
+bash -c "${cidir}/install_go.sh -p"
```
jcvenegas: @grahamwhaley  @jodh-intel @chavafg @sboeuf  what do you think about this, do you think we can call this script in the Jenkins setup instead?

.ci/install_go.sh
```diff
@@ -0,0 +1,94 @@
...
+	rm -rf "$tmp_dir"
+}
+
+die() {
+	echo >&2 -e "\e[1mERROR\e[0m: $*"
+	exit 1
+}
+
+info() {
+	echo -e "\e[1mINFO\e[0m: $*"
```
grahamwhaley: urg, embedded control codes??

.ci/install_go.sh
```diff
@@ -0,0 +1,94 @@
...
+Options
+-p : Install go defined in ${PROJECT} versions file.
+-h : Show this help
+
+EOT
+exit "$exit_code"
+}
+
+
+trap finish EXIT
```
grahamwhaley: don't know if you want EXIT, QUIT and KILL here?

.ci/install_go.sh
```diff
@@ -0,0 +1,94 @@
...
+
+info "Download go version ${go_version}"
+curl -OL https://storage.googleapis.com/golang/go${go_version}.linux-amd64.tar.gz
+info "Remove old go installation"
+sudo rm -r /usr/local/go/
+info "Install go"
+sudo tar -C /usr/local -xzf go${go_version}.linux-amd64.tar.gz
+popd
+
+
```
grahamwhaley: trailing whitespace :-)

grahamwhaley (approved):
nits. overall:
lgtm

.ci/setup.sh
grahamwhaley: I think we can, as long as we define when and where well.
If we were to do that - I'd like to see a check that *only* installs if the version is different from what is already installed - no point installing the same go version over and over again (on a bare metal instance) iyswim

.ci/install_go.sh
jcvenegas: I can remove them, I like to see highlighted messages :P sorry

jodh-intel: You could always add `-c | --color` CLI option? :)

jcvenegas:
depends on https://github.com/clearcontainers/tests/pull/967 to update golang to 1.10 
depens on https://github.com/clearcontainers/runtime/pull/1091 to get cri-containerd version to install in ci.

.ci/install_go.sh
```diff
@@ -0,0 +1,88 @@
...
+Options
+-p : Install go defined in ${PROJECT} versions file.
+-h : Show this help
+
+EOT
+exit "$exit_code"
+}
+
+trap finish EXIT
```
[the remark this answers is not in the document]
icarus-sparry: I think just using EXIT is fine here.

In practice adding INT, KILL etc will not make much difference, you have `set -o errexit` and subprocesses being killed will exit with non zero codes anyhow so you will take the exit trap, and unless the parent of this script has set up signal handling to ignore for it, this script itself will exit on a keyboard signal. 

If something is going wrong then being able to type ctrl-\ and not have the evidence destroyed is useful. This argues for adding

        trap 'trap "" EXIT' QUIT

to unset the exit trap on a QUIT. As the files are in /tmp they will be cleaned eventually.

Thank you for the question, it makes me realize I need another slide in my bash talk.

PR #968, opened 2018-04-05
tests: tests cc-runtime as default runtime for cri-containerd

jcvenegas:
This PR adds tests to run `cc-runtime` as default runtime with `cri-containerd`

- ci: Add scripts to install cri-containerd
_Note: I added  the installation in `.ci/run.sh`, If install before run kueberentes test, k8s tests are going to fail. It is need to move our CI to k8s 1.10 first_

- Added new tests target to test only cri-containerd `make cri-containerd`
1. Added test-cri (critools) 
2. Added test-integration(cri-containerd integration tests.)

### DIFF
--- a/.ci/install_cri_containerd.sh
+++ b/.ci/install_cri_containerd.sh
@@ -18,8 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-
-tmp_dir=$(mktemp -d -t tmp.XXXXXXXXXX)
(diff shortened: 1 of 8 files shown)


grahamwhaley:
@sboeuf I think you are probably right - we should prefer adding to kata now, and should only add things here if they are a cc specific fix or addition. Make sense @jcvenegas ?

jcvenegas:
@sboeuf @grahamwhaley, sure I'll move it to Kata. Closing this PR.

